PR #4, opened 2019-10-03
Print examples

langston-barrett:
Looks like this:
```
Example: λx. x ↦ #<closure>
```


david-christiansen:
I think it makes more sense to only print example output for the module that we're presently loading/working on, rather than for all the transitive dependencies of the current module. 

Perhaps this should be in main?

PR #8, opened 2019-10-04
klister.el: init

langston-barrett:


langston-barrett:
@david-christiansen suggested adding `compile` support, which I will do before merging.

PR #13, opened 2019-10-04
n-ary-app: add 'compose' as another example

langston-barrett:


david-christiansen:
This is ready for merging, right?

PR #16, opened 2019-10-07
Quasiquotation macro

david-christiansen:
This does not yet handle variable-arity operators.

WIP. To merge, it needs:
 * [x] Recursive cons-list versions for variable-arity macros
 * [x] More copied cases for fixed arity up to five or so
 * [x] More tests that show it actually works
 * [x] Inclusion into the overall test suite for Klister

### DIFF
--- a/examples/n-ary-app.kl
+++ b/examples/n-ary-app.kl
@@ -66,9 +66,13 @@
 [export quote]
 [export meta]
 [export import]
+[export >>=]
 [export pure]
+[export if]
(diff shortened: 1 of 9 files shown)


david-christiansen:
CC @langston-barrett - this includes the catchall pattern we talked about for syntax matching.

david-christiansen:
OK, if Travis is happy then this is good to go.

david-christiansen:
This also includes the logging effect.

PR #18, opened 2019-10-08
Add special reader syntax for quotation

david-christiansen:
This is in the broad Lisp tradition. Basically, the parser just inserts calls to `quote`, `quasiquote`, or `unquote` when given the appropriate syntax afterwards.


david-christiansen:
Note that the syntax here corresponds not to the old-fashioned Lisp `quote`, but to Racket's `syntax-quote`, `quasisyntax`, and `unsyntax`. We don't have non-syntax quotation, so the simple names are fine here.

PR #19, opened 2019-10-08
Fix expander bugs and add a hygiene test that requires these fixes

david-christiansen:
@langston-barrett  I think this will likely fix the "variable out of scope" error you were getting before as well.

### DIFF
--- a/tests/Test.hs
+++ b/tests/Test.hs
@@ -328,7 +328,7 @@ moduleTests = testGroup "Module tests" [ shouldWork, shouldn'tWork ]
                   spec1 <- lam \x -> lam \_y -> x
                   spec2 <- lam \_x -> lam \y -> y
                   assertAlphaEq "First fun drops second argument" firstFun spec1
-                  assertAlphaEq "Second fun is drops first argument" secondFun spec2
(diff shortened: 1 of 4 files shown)


tests/Test.hs
```diff
@@ -319,6 +319,32 @@ moduleTests = testGroup "Module tests" [ shouldWork, shouldn'tWork ]
...
+        , ( "examples/hygiene.kl"
+          , \m ->
+              view moduleBody m & map (view completeDecl) &
+              \case
+                [Import _, Import _, Define _ fun1 firstFun, DefineMacros [_], Define _ fun2 secondFun,
+                 Example e1, Example e2, DefineMacros [_], Example e3] -> do
+                  spec1 <- lam \x -> lam \_y -> x
+                  spec2 <- lam \_x -> lam \y -> y
+                  assertAlphaEq "First fun drops second argument" firstFun spec1
+                  assertAlphaEq "Second fun is drops first argument" secondFun spec2
```
langston-barrett:
```suggestion
                  assertAlphaEq "Second fun drops first argument" secondFun spec2
```

PR #21, opened 2019-10-09
Add macros in declaration positions

david-christiansen:
This is illustrated by a function-definition macro that uses fix behind the scenes to implement recursive functions.

There are still no macros for module bodies yet.

### DIFF
--- a/examples/defun.kl
+++ b/examples/defun.kl
@@ -1,0 +1,30 @@
+#lang "n-ary-app.kl"
+[import [shift "n-ary-app.kl" 1]]
+[import [shift "quasiquote.kl" 1]]
+
+[define fix
+  [lambda (fun)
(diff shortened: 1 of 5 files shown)


david-christiansen:
It looks to me like Travis failed for network reasons rather than because the code is busted. @gelisam - can you give it a poke?

langston-barrett:
I clicked the restart button

david-christiansen:
Thanks @langston-barrett!

david-christiansen:
I think this is ready for merge whenever one of you two has a chance to look it over. It seems likely that some of the duplication between expanding a single expression or a single definition can be factored out, but I want to get module macros working first so I can see what the real overlaps and non-overlaps are before doing the refactor.

PR #22, opened 2019-10-09
Hedgehog generators for Core

langston-barrett:
Are there any other generic properties we should add here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ sudo: false
 matrix:
   include:
     - env: CABALVER=3.0 GHCVER=8.8.1
-      addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1],sources: [hvr-ghc]}}
+      addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1,happy-1.19.5],sources: [hvr-ghc]}}
     - env: CABALVER=2.4 GHCVER=8.6.5
-      addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.5],sources: [hvr-ghc]}}
+      addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.5,happy-1.19.5],sources: [hvr-ghc]}}
 
 before_install:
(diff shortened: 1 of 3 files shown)


david-christiansen:
It looks like the Travis script is failing because it needs `happy` to be installed.

david-christiansen:
As far as properties go, it might be interesting to split a core term, then delete one of its subtrees, then turn that into a `PartialCore` and check that the spot in question contains a `Nothing`.  But I don't think its pressing, and once Travis is fixed, I think we should just merge this.

langston-barrett:
Yeah, I wonder why it passes on GHC 8.8... Here's the relevant part of the log for my own future reference:
```
cabal: The program 'happy' is required but it could not be found.
cabal: Leaving directory '/tmp/cabal-tmp-4151/pretty-show-1.9.5'
Downloading  vector-0.12.0.3
Downloaded   vector-0.12.0.3
Starting     vector-0.12.0.3
Building     vector-0.12.0.3
Completed    unordered-containers-0.2.10.0
Completed    vector-0.12.0.3
cabal: Error: some packages failed to install:
hedgehog-1.0.1-48Mns07EeRZEzOx5N5HKhf depends on hedgehog-1.0.1 which failed
to install.
pretty-show-1.9.5-4s9IVBWSe01IF3L4DDRCg8 failed during the configure step. The
exception was:
ExitFailure 1
tasty-hedgehog-1.0.0.1-FsHMkdFEYVZGW7ySH5MekX depends on
tasty-hedgehog-1.0.0.1 which failed to install.
```

langston-barrett:
Huh, even after adding `happy` via `apt`, `cabal` still can't find it

david-christiansen:
I suspect that `cabal-install` being version 3.0 is more relevant than the fact that it's GHC 8.8, because version 3 is using `v2`-style builds by default. One easy fix here might be to write `v2-` in the Travis file.

langston-barrett:
Ugh... for some reason the GHC 8.6.5 build is erroring before even starting: https://travis-ci.org/gelisam/klister/jobs/595812912

Issue #23, opened 2019-10-09
Redundant pattern match in Expander


langston-barrett:
https://travis-ci.org/gelisam/klister/jobs/595776213#L770
```
src/Expander.hs:918:9: warning: [-Woverlapping-patterns]
    Pattern match is redundant
    In a case alternative: EUserMacro _otherVal -> ...
    |
918 |         EUserMacro _otherVal ->
    |         ^^^^^^^^^^^^^^^^^^^^^^^...
```

david-christiansen:
Fixed as a side effect of #24.

PR #24, opened 2019-10-13
Eliminate the square bracket/paren distinction

### LINKED ISSUE #15: Ditch paren-square bracket distinction in syntax
Now that I'm writing some more macros, the limitations of our approach to distinguishing fixed-arity vs variable-arity operators is becoming more clear.

In particular, I've been trying to write a quasiquoting macro. This would be used for writing other macros much more conveniently. However, the fixed-arity nature of our square bracket pattern matching operator means that I essentially need to copy-paste the pattern-match cases, once for each length of square-bracketed-list that I want to be able to support. I can't even generate all the cases with a metaprogram :-)

The way I see it, there's two ways forward: (1) eliminate the distinction, or (2) provide variable-arity metaprogramming operators for fixed-arity programs (e.g. a list-like API to the square-bracketed vectors).

Replicating the variable-arity API for fixed-arity things seems to be a failure, so I propose that we do number 1. Any objections?

david-christiansen:
This PR fixes #15 by doing the following:
1. The semantic distinction between parens and square brackets is removed. Now, as in Racket, they are synonymous, though square brackets and parens must still be paired with like delimiters (so `[)` is a parse error)
2. All instances of `vec` are renamed to `list` in operator names
3. The examples are rewritten to take into account the lack of fixed-width patterns. In particular, quasiquotation became much simpler.

While reworking the examples, bugs were encountered. This PR thus also fixes:
1. `free-identifier=?` returns false when one or both of its arguments have an ambiguous binding instead of throwing an ambiguous binding error. Otherwise, quasiquotation would crash on quotations that are ambiguous when comparing an identifier to `unquote`. Ambiguous quotations are useful when the ambiguous identifier occurs in a binding position, for instance.
2. Recursive loading of shifted modules did not shift the values of recursively-loaded modules. In other words, if module A loads B shifted by 1, and B loads C shifted by 1, then C should be shifted by 2. This was not happening, but now it is.
3. The Stack configuration did not allow loading `tasty-hedgehog` because Stackage didn't  have it in the chosen snapshot. Updating the snapshot fixed this so the tests could run with `stack` as well as `cabal-install`.

### DIFF
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -124,26 +124,27 @@ repl ctx startWorld = do
   forever $
     do putStr "> "
        l <- T.getLine
-       tryCommand theWorld l $ \l' -> case readExpr "<repl>" l' of
-         Left err -> T.putStrLn err
-         Right ok ->
-           do putStrLn "Parser output:"
-              T.putStrLn (syntaxText ok)
-              c <- execExpand (expandExpr ok) ctx
-              case c of
-                Left err -> putStr "Expander error: " *>
-                            prettyPrintLn err
-                Right (unsplit -> out) -> do
-                  putStrLn "Expander Output:"
-                  print out
-                  case runPartialCore out of
-                    Nothing -> putStrLn "Expression incomplete, can't evaluate"
-                    Just expr -> do
-                      putStrLn "Complete expression in core:"
-                      prettyPrint expr
-                      putStrLn ""
-                      currentWorld <- readIORef theWorld
-                      runExceptT (runReaderT (runEval (eval expr)) (phaseEnv runtime currentWorld)) >>=
-                        \case
-                          Left evalErr -> print evalErr
-                          Right val -> prettyPrint val >> putStrLn ""
+       tryCommand theWorld l $ \l' ->
+         case readExpr "<repl>" l' of
+           Left err -> T.putStrLn err
+           Right ok ->
+             do putStrLn "Parser output:"
+                T.putStrLn (syntaxText ok)
+                c <- execExpand (expandExpr ok) ctx
+                case c of
+                  Left err -> putStr "Expander error: " *>
+                              prettyPrintLn err
+                  Right (unsplit -> out) -> do
+                    putStrLn "Expander Output:"
+                    print out
+                    case runPartialCore out of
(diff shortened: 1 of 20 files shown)


david-christiansen:
This also fixes #23.

repl/Main.hs
```diff
@@ -124,26 +124,27 @@ repl ctx startWorld = do
   forever $
     do putStr "> "
        l <- T.getLine
-       tryCommand theWorld l $ \l' -> case readExpr "<repl>" l' of
```
langston-barrett: Hmm, the diff algorithm is bad here... It appears that the only change is that you put this `case` on a new line, is that right?

src/Expander.hs
```diff
@@ -164,41 +166,65 @@ getImports (PrefixImports spec pref) =
...
+    let m' = shift i m -- Shift the syntax literals in the module source code
+    evalResults <- inPhase p $ evalMod m'
     modifyState $
       set (expanderWorld . worldEvaluated . at modName)
           (Just evalResults)
-
-
   return (shift i es)
 
+-- | Evaluate an expanded module at the current expansion phase,
```
langston-barrett: What caused this move? Are there any changes in the code, or did it just migrate?

repl/Main.hs
david-christiansen: Thinking back, I remember being confused by having so much there and moving it for the sake of understanding it. So yes :)

src/Expander.hs
david-christiansen: It had to move because the handing of imports changed. Before, evaluating a module with imports in it just ignored the declared imports. But what actually needs to happen is to load those imports with respect to the phase at which we're evaluating the module. Loading the imports requires visiting the module, and visiting is in the expander monad rather than the evaluator monad. Hence the new location.

The resulting code is also simpler in the expander monad. A lot of the tracking of phases and environments with the `RWST` transformer on top of the expander was really simulating things that the expander does anyway.

david-christiansen: Here's the most important line of that change: https://github.com/gelisam/klister/pull/24/files/883e8954624a0c2c84cdbc129dcd4a2335ed516f#diff-a63cde2fb12e931100bbb12fc30e04eeR226

PR #25, opened 2019-11-06
Export specs

david-christiansen:
This PR does the following:
1. A more informative syntax for exports, allowing things like renaming and exporting at phases other than 0
2. Fixes a number of bugs in scope addition that were exposed by checking number 1
3. Fix long-standing issues in binding tables and shifted macros that were masked by the bugs from step 2.

In particular, expanded modules now contain a binding table that can be shifted along with their contents.


david-christiansen:
One more thing done in this PR is to add provenance indicators to scopes, which makes it much easier to debug scope-related issues. The provenance is just a string, and we can eliminate them later if they cause any unfortunate overhead.

david-christiansen:
Any objections to a merge?

PR #26, opened 2020-01-17
Initial draft of a scopechecker

langston-barrett:
The Big TODOs:
 - [ ] Tests!
 - [ ] The `Expander` monad should actually call this at some point! Where?
     - [ ] Should it pass an initial context consisting of imported names and definitions that have already been seen?

### DIFF
--- a/src/Env.hs
+++ b/src/Env.hs
@@ -2,7 +2,18 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE TypeFamilies #-}
-module Env (Env, empty, insert, singleton, lookup, lookupIdent, lookupVal, toList, named) where
+module Env
+  ( Env
+  , empty
+  , insert
+  , singleton
+  , lookup
+  , lookupIdent
+  , lookupVal
+  , toList
+  , named
+  , keys
+  ) where
 
 import Prelude hiding (lookup)
 
@@ -39,6 +50,9 @@ lookupIdent x env = fst <$> lookup x env
 
 named :: Text -> Env v a -> [(v, a)]
 named n (Env xs) = [(x, a) | (x, (Stx _ _ n', a)) <- Map.toList xs, n == n']
+
+keys :: Env v a -> [v]
+keys (Env e) = Map.keys e
 
 type instance Index (Env v a) = v
 type instance IxValue (Env v a) = (Ident, a)
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -26,7 +26,6 @@ module Expander (
   , currentPhase
   , expandEval
   , ExpansionErr(..)
-  , ExpanderContext
   , expanderBindingTable
   , expanderWorld
   , getState
@@ -34,7 +33,7 @@ module Expander (
   , addModuleScope
   ) where
 
-import Control.Lens hiding (List, children)
+import Control.Lens hiding (Context(..), List, children)
 import Control.Monad.Except
 import Control.Monad.Reader
 import Control.Monad.Writer
@@ -64,6 +63,7 @@ import Parser
 import Phase
 import Pretty
 import Scope
+import ScopeCheck
 import ScopeSet (ScopeSet)
 import Signals
 import ShortShow
@@ -878,6 +878,7 @@ runTask (tid, localData, task) = withLocal localData $ do
               \case
                 Nothing -> Just $ Env.singleton x n val
                 Just env -> Just $ env <> Env.singleton x n val
+    ScopeCheckTask splitCore -> scopeCheckExpand localData splitCore
   where
     laterMacro tid' b v x dest deps mdest stx = do
       localConfig <- view expanderLocal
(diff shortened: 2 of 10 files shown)


david-christiansen:
WRT getting this code called, I think that the right way to do it is to establish a newtype wrapper around expressions that have been checked, and then make it so that the evaluator and the module system require the checked version. It should then become clear where we need to call the checker! This will also generalize nicely to the type checker.

src/Env.hs
```diff
@@ -40,6 +51,9 @@ lookupIdent x env = fst <$> lookup x env
 named :: Text -> Env v a -> [(v, a)]
 named n (Env xs) = [(x, a) | (x, (Stx _ _ n', a)) <- Map.toList xs, n == n']
 
+keys :: Env v a -> [v]
```
david-christiansen: This doesn't seem to be used anywhere.

src/Expander.hs
```diff
@@ -1074,3 +1075,36 @@ interpretMacroAction (MacroActionIdentEq how v1 v2) = do
 interpretMacroAction (MacroActionLog stx) = do
   liftIO $ prettyPrint stx >> putStrLn ""
   pure $ Right (ValueBool False) -- TODO unit type
+
+type NewScopeCheckTask = (Context, SplitCore)
+
+scopeCheckExpand :: ExpanderLocal -> SplitCore -> Expand ()
+scopeCheckExpand localState splitCore = do
```
david-christiansen: I'd prefer a name other that `splitCore` for this, because that can be seen from the type signature anyway. This is the expression being checked, so I think that `expr` or `toCheck` would be more informative.

src/Expander.hs
```diff
@@ -1074,3 +1075,36 @@ interpretMacroAction (MacroActionIdentEq how v1 v2) = do
...
+  phase <- currentPhase
+  let ctx = view expanderContext localState
+  root <-
+    case getRoot splitCore of
+      Nothing -> fail "internal error"
```
david-christiansen: We have an internal error constructor on the expressions. Can this be `throwError`?

src/Expander.hs
```diff
@@ -1074,3 +1075,36 @@ interpretMacroAction (MacroActionIdentEq how v1 v2) = do
...
+      env <- ask -- possibly-expanded
+      case subtree splitCore splitCorePtr of
+        Nothing -> fail "Malformed splitcore!"
+        Just splitCoreSubtree -> lift $ tell [(env, splitCoreSubtree)]
+  let errorOr =
+        runWriter $
+          runScopeCheckT ctx $
+            scopeCheckCoreF recursiveCase phase root
+  case errorOr of
+    (Left _, _) -> fail "Scope check error!"
```
david-christiansen: This should presumably be a real error constructor, rather than a panic.

src/Expander/Monad.hs
```diff
@@ -1,8 +1,10 @@
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
```
david-christiansen: Why do we need this?

david-christiansen:
I added some concrete comments. I think that this approach is highly plausible.
Here's a few ideas:

1. I would think that we want something like this for the output of the checker:
```
data Checked e a = Ok a | Failure e
newtype CheckedCore = CheckedCore (Checked ScopeErr (CoreF CheckedCore))
checkOK :: CheckedCore -> Either (NonEmpty ScopeErr) Core
```
This allows us to nicely report on multiple errors at once, rather than just throwing an exception, and will eventually allow us to do things like make tools that are still useful on programs that contain type errors. Really, this should probably be used for errors during macro expansion as well. Thoughts? Should we save this for a future iteration? Should the errors instead be accumulated as a `NonEmpty ScopeErr` at the top level of the tree structure, with subtrees being well-formed?

2. The output datatype should probably include the context in which the expression was checked, and not just be a newtype wrapper that says "we did it". That can be used to ensure that the environment provided to the evaluator is suitable for that context, for instance, and as a basis for the `local-expand` implementation that I sketched in our meeting.

I really like the approach of providing a monad that can be interpreted into the expander monad or directly over core expressions. This helps with our argument that it's just the usual kind of type checker that's being augmented with macros.

I don't think that the current limited recursive structure is a big deal (that is, the inability to write arbitrary recursive stuff that we could get by using a higher-order representation of checker actions in the Expand monad, instead of just recording its arguments) because a given type checker will only have a finite number of judgment forms, each of which can get a task constructor / means of invocation.

david-christiansen:
Here's some further reflections on how to integrate it. I hope to have some time soon to do this (assuming you don't object), but please feel free to grab it first if you can!

I think that the way to integrate the scope checker (and eventually the type checker) is to first rearchitect it so that the first-order representation of a checking task lists the `SplitCorePtr` that it is intended to check, rather than have a `SplitCore` itself. Then, it's easy to check when a step is ready to run: it's ready when the overall mapping in the expander state connects it to a `CoreF SplitCorePtr`. This makes it more closely follow the other tasks, and it also gives a fairly simple way to create the top-level tasks. Once a top-level definition is created, the system can see which others are in scope, and use that to create the context for scope checking the RHS, setting up the task appropriately.

Thoughts?

david-christiansen:
The first part of that comment can be seen on branch https://github.com/gelisam/klister/tree/dtc/scopecheck-additions .

langston-barrett:
Having the task contain a pointer rather than the syntax sounds like a great idea, it could be more performant as well.

david-christiansen:
I suspect that there will be no performance difference because this just
changes when the lookup is done. It does make the API more flexible, though!

Mind if I just push that change into the PR branch when I'm at a computer
again?

langston-barrett:
Not at all, sounds good!

david-christiansen:
OK, this is pushed.

gelisam:
Is this branch still relevant, or has it been superseded by the typechecker?

langston-barrett:
I think it's been superseded.

PR #27, opened 2020-01-29
WIP: Begin implementing type checker

david-christiansen:
This is still in an early state.

To do before merging:
 * [ ] Finish type checker infrastructure (add `let`, generalization, etc)
 * [ ] Add typing rules for all `CoreF` constructors
 * [ ] Track source locations on `Core` terms to report type error locations
 * [ ] Tests, tests, more tests
 * [ ] Invoke type checker where needed, not just on annotations

### DIFF
--- a/examples/tiny-types.kl
+++ b/examples/tiny-types.kl
@@ -1,0 +1,11 @@
+#lang kernel
+
+(example (the Bool #true))
+(example (the Bool #false))
+(example (the Syntax 'foo))
+(example (the (Macro Syntax) (pure 'foo)))
+
+
+
+
+
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE RankNTypes #-}
@@ -61,6 +62,7 @@ import Evaluator
 import Expander.DeclScope
 import Expander.Syntax
 import Expander.Monad
+import Expander.TC
 import Module
 import ModuleName
 import Parser
@@ -71,8 +73,10 @@ import ScopeSet (ScopeSet)
 import Signals
 import ShortShow
 import SplitCore
+import SplitType
 import Syntax
 import Syntax.SrcLoc
+import Type
 import Value
 import World
 import qualified ScopeSet
@@ -386,8 +390,57 @@ initializeKernel = do
   traverse_ (uncurry addExprPrimitive) exprPrims
   traverse_ (uncurry addModulePrimitive) modPrims
   traverse_ (uncurry addDeclPrimitive) declPrims
+  traverse_ (uncurry addTypePrimitive) typePrims
 
   where
+    typePrims :: [(Text, SplitTypePtr -> Syntax -> Expand ())]
+    typePrims =
+      [ ( "Bool"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TBool
+        )
+      , ( "Unit"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TUnit
+        )
+      , ( "Syntax"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TSyntax
+        )
+      , ( "Ident"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TIdent
+        )
+      , ( "Signal"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TSignal
+        )
+      , ( "->"
+        , \ dest stx -> do
+            Stx _ _ (_ :: Syntax, arg, ret) <- mustHaveEntries stx
+            argDest <- scheduleType arg
+            retDest <- scheduleType ret
+            linkType dest (TFun argDest retDest)
+        )
+      , ( "Macro"
+        , \ dest stx -> do
+            Stx _ _ (_ :: Syntax, t) <- mustHaveEntries stx
+            tDest <- scheduleType t
+            linkType dest (TMacro tDest)
+        )
+      , ( "List"
+        , \ dest stx -> do
+            Stx _ _ (_ :: Syntax, e) <- mustHaveEntries stx
+            entryTypeDest <- scheduleType e
+            linkType dest (TList entryTypeDest)
+        )
+      ]
+
     modPrims :: [(Text, Syntax -> Expand ())]
     modPrims =
       [ ( "#%module"
@@ -561,6 +614,14 @@ initializeKernel = do
     exprPrims =
       [ ( "oops"
         , \ _ stx -> throwError (InternalError $ "oops" ++ show stx)
+        )
+      , ( "the"
+        , \ dest stx -> do
+            Stx _ _ (_, ty, expr) <- mustHaveEntries stx
+            tyDest <- scheduleType ty
+            -- TODO add type to elaborated program? Or not?
+            forkExpandSyntax (ExprDest dest) expr
+            forkTypeCheck dest tyDest
         )
       , ( "lambda"
         , \ dest stx -> do
@@ -756,6 +817,11 @@ initializeKernel = do
       bind b (EVarMacro var)
       return (sc, x', var)
 
+    scheduleType :: Syntax -> Expand SplitTypePtr
+    scheduleType stx = do
+      dest <- liftIO newSplitTypePtr
+      forkExpandType dest stx
+      return dest
 
     schedule :: Syntax -> Expand SplitCorePtr
     schedule stx = do
@@ -781,6 +847,12 @@ initializeKernel = do
       bind b val
       addToKernel name runtime b
 
+    addTypePrimitive :: Text -> (SplitTypePtr -> Syntax -> Expand ()) -> Expand ()
+    addTypePrimitive name impl = do
+      let val = EPrimTypeMacro impl
+      b <- freshBinding
+      bind b val
+      addToKernel name runtime b
 
     addExprPrimitive :: Text -> (SplitCorePtr -> Syntax -> Expand ()) -> Expand ()
     addExprPrimitive name impl = do
@@ -866,6 +938,8 @@ runTask (tid, localData, task) = withLocal localData $ do
           expandOneExpression d stx
         DeclDest d sc ph ->
           expandOneDeclaration sc d stx ph
+    ExpandType dest stx ->
+      expandOneType dest stx
     AwaitingSignal dest signal kont -> do
       signalWasSent <- viewIORef expanderState (expanderReceivedSignals . at signal)
       case signalWasSent of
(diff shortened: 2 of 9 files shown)


gelisam:
I came here to ask what type `defun.kl`'s `fix` will have, but I see you're defining recursive `let` and `flet` as builtins now, so that answers my question, thank you!

david-christiansen:
Yeah, HM really needs a dedicated definition form as the site of type
generalization.

The flet is the only one with recursion here because it's a strict language
for now.

>
>
>

david-christiansen:
I just updated `defun` to the new regime.

david-christiansen:
I think that this is basically ready for a review. I need to find more programs that it should reject, and test that it rejects them, and it would be nice to have the occurs check blame a source location. But I don't expect major changes from here on out.

examples/defun.kl
```diff
@@ -37,4 +28,12 @@
 
 [example (last-stx '(a b c d e f g))]
 
+(defun map-syntax (f stx)
```
gelisam: this definition of `map-syntax` is dead code; should there be an `example`, or perhaps it should be in the `export` list?

david-christiansen: I should just delete it. I had an incorrect `flet*`, and that led to type errors in this, so I had it here so I could test more easily.

david-christiansen:
Anything else blocking a merge, besides a few more tests?

gelisam:
I've barely started reviewing it, but if you want to merge it now I won't stop you :)

david-christiansen:
Ah, sorry, I couldn't figure out if that was your whole review :-) I was
confused by the GitHub interface.

examples/n-ary-app.kl
```diff
@@ -59,6 +59,30 @@
...
+  ([flet*
+    (lambda (stx)
+      (syntax-case stx
+        [(list (_ fun body))
+         (syntax-case fun
+           [(list (name args def))
+            (syntax-case args
+              [(cons arg more-args)
+               (pure (list-syntax
+                      ('flet
```
gelisam: Woah! You [explained](https://github.com/gelisam/klister/pull/28#discussion_r374862904) that `define-macros` enables quotations in its body to refer to the macro's name. But here it looks like `define-macros` also enables `flet*` to refer to the names of macros which aren't even defined yet? How is that possible?

My first hypothesis was that unbound identifiers produced by a macro will simply refer to whatever identifier is in scope at the call site, whereas bound identifier won't change their meaning thanks to hygiene. However, this simple example proves that theory wrong:

    (define-macros
      ((m (lambda (stx)
            (pure 'unbound)))))
    
    (define unbound 2)
    
    (example m)  -- error: unknown <unbound>

The only difference between `unbound` and `flet` seems to be that `flet` is a macro, so my second hypothesis was that unbound identifiers produced by a macro may refer to a _macro_ which is in scope at the call site, but that hypothesis was also easy to disprove:

    (define-macros
      ((m (lambda (stx)
            (pure 'unbound)))))
    
    (define-macros
      ((unbound (lambda (stx)
                  (pure ''bound)))))
    
    (example m)  -- error: unknown <unbound>

So what's the difference between `flet*` and `m`?

gelisam: Ah, I get it! `flet*` is not referencing the `flet` below, it's referencing the `flet` builtin, and the `flet` below is shadowing the builtin, just like `#%app` and `lambda` do.

examples/n-ary-app.kl
```diff
@@ -59,6 +59,30 @@
...
+                       (list-syntax (name
+                                     (list-syntax (arg) stx)
+                                     (list-syntax ('lambda more-args def) stx))
+                                    stx)
```
gelisam: We're pretty much always using `stx` for the last argument of `list-syntax` and `cons-list-syntax`; it seems like this argument complicates the API without bringing much value. Now that we have `replace-loc`, maybe we can simplify that API in the common case where we don't want to think much about this third argument, and use `replace-loc` in the few places where the location matters?

langston-barrett: I think that's a reasonable idea, but we might want to tackle it in a separate PR

examples/lets.kl
```diff
@@ -0,0 +1,14 @@
...
+(example (flet (f (x) (if x 'a 'b)) (f #true)))
+
+(example (the Syntax (flet (f (x) 'a) (list-syntax ((f #true) (f 4)) 'here))))
+
+(example (the Syntax (let (f (lambda (x) 'a)) (list-syntax ((f #true) (f 4)) 'here))))
+
+(define id (lambda (x) x))
+
+(example (the Syntax (id 'foo)))
+(example (the Signal (id 45)))
```
langston-barrett: It would be nice to have an example `the T` for every primitive type `T` in a separate file (perhaps `these.kl`)

examples/tiny-types.kl
```diff
@@ -0,0 +1,20 @@
...
+(example (the (Macro Syntax) (pure 'foo)))
+(example (the (-> Bool Bool) (lambda (x) x)))
+(example (the (-> Bool Syntax) (lambda (x) 'hello)))
+(example (the (Macro (-> Bool Bool))
+              (>>= (free-identifier=? 'a 'b)
+                   (lambda (eq?)
+                     (if eq?
+                         (pure (lambda (x) x))
+                         (pure (lambda (x) (if x #false #true))))))))
+
```
langston-barrett: Lots of blank lines here

examples/lets.kl
langston-barrett: Ah, looks like that's in `tiny-types.kl`. These can probably be moved there? What's the difference between the content of that file and this one?

src/Expander.hs
```diff
@@ -119,6 +124,7 @@ expandModule thisMod src = do
                            }
     bs <- view expanderCurrentBindingTable <$> getState
     modifyState $ set expanderCurrentBindingTable startBindings
+    modifyState $ set expanderDefTypes startDefTypes
```
langston-barrett: Is the purpose here to remove `defTypes` that were added during the other expansion tasks?

src/Expander.hs
```diff
@@ -386,8 +398,57 @@ initializeKernel = do
...
   traverse_ (uncurry addDeclPrimitive) declPrims
+  traverse_ (uncurry addTypePrimitive) typePrims
 
   where
+    typePrims :: [(Text, SplitTypePtr -> Syntax -> Expand ())]
+    typePrims =
+      [ ( "Bool"
+        , \ dest stx -> do
+            _actualName <- mustBeIdent stx
+            linkType dest TBool
```
langston-barrett: Looks like the body of this lambda is just copy/pasted with different `T*` constructors, maybe add a `let` binding for it before this list?

src/Expander.hs
```diff
@@ -562,6 +628,39 @@ initializeKernel = do
       [ ( "oops"
         , \ _ stx -> throwError (InternalError $ "oops" ++ show stx)
         )
+      , ( "the"
+        , \ dest stx -> do
+            Stx _ _ (_, ty, expr) <- mustHaveEntries stx
+            tyDest <- scheduleType ty
+            -- TODO add type to elaborated program? Or not?
```
langston-barrett: It would make sense to me to have the output of expansion+typechecking be something like
```haskell
newtype TypedCore = TypedCore { getTypedCore :: (Type, CoreF TypedCore) }
```
for easy analysis (e.g. thinking of an optimization pass), and then have the input to evaluation be `Core`. Then after expansion but before evaluation we could
```haskell
eraseTypes :: TypedCore -> Core
```

src/Expander.hs
```diff
@@ -882,7 +1001,17 @@ runTask (tid, localData, task) = withLocal localData $ do
...
+        [] -> do
+          inEarlierPhase $
+            forkCompleteTypeCheck mdest $
+              Ty (TFun (Ty TSyntax)
+                       (Ty (TMacro (Ty TSyntax))))
+          forkAwaitingMacroType b v x mdest dest stx
+    AwaitingMacroType dest (TaskAwaitMacroType b v x mdest stx) ->
+      isExprChecked mdest >>=
+      \case
+        False -> stillStuck tid task
```
langston-barrett: Just thinking out loud, but I feel like the optimal API for tasks would be two monadic functions: one in a `Reader` to check readiness/preconditions, and another in `State` to actually do the thing. That way, the expander could check what tasks are ready without potentially mutating state.

src/Expander.hs
```diff
@@ -952,13 +1081,101 @@ runTask (tid, localData, task) = withLocal localData $ do
...
           ((tid', localConfig, AwaitingMacro dest (TaskAwaitMacro b v x deps mdest stx)) :)
 
+expandOneType :: SplitTypePtr -> Syntax -> Expand ()
+expandOneType dest stx
+  | Just ident <- identifierHeaded stx = do
+      b <- resolve =<< addRootScope ident
+      v <- getEValue b
+      case v of
+        EPrimTypeMacro impl -> impl dest stx
+        EPrimMacro _ -> throwError $ InternalError "Context expects a type"
```
langston-barrett: Could these errors arise from user macros in macro positions returning non-type expressions?

src/Expander.hs
```diff
@@ -1153,3 +1373,215 @@ interpretMacroAction (MacroActionIdentEq how v1 v2) = do
 interpretMacroAction (MacroActionLog stx) = do
   liftIO $ prettyPrint stx >> putStrLn ""
   pure $ Right (ValueBool False) -- TODO unit type
+
+-- | Invariant: the SplitCorePtr points at the CoreF in question
+typeCheckLayer :: SplitCorePtr -> CoreF SplitCorePtr -> Ty -> Expand ()
+typeCheckLayer dest (CoreVar _) t =
+  forkCheckVar dest t
+typeCheckLayer dest (CoreLet x ident def body) t = do
+  xTy <- inTypeBinder do
```
langston-barrett:
```suggestion
  xTy <- inTypeBinder $ do
```

src/Expander/Error.hs
```diff
@@ -113,3 +118,19 @@ instance Pretty VarInfo ExpansionErr where
     text "Internal error during expansion! This is a bug in the implementation." <> line <> string str
   pp _env (ReaderError txt) =
     vsep (map text (T.lines txt))
+  pp env (NotValidType stx) =
+    hang 2 $ group $ vsep [text "Not a type:", pp env stx]
+  pp env (TypeMismatch loc expected got) =
+    group $ vsep [ group $ hang 2 $ vsep [ text "Type mismatch at"
+                                         , maybe (text "unkown location") (pp env) loc <> text "."
```
langston-barrett:
```suggestion
                                         , maybe (text "unknown location") (pp env) loc <> text "."
```

examples/n-ary-app.kl
david-christiansen: That's right!

examples/n-ary-app.kl
david-christiansen: The last `stx` argument is mostly for the scopes, and only incidentally for the source location. It's a bit repetitive, but the point is to say "the identifiers in the resulting syntax object should be understood as originating in *this* context, unless otherwise specified on the identifier itself", not to say "it comes from this file and line".

examples/lets.kl
david-christiansen: Only a chronological difference WRT when I wrote it. It's not a thought-through placement.

src/Expander.hs
david-christiansen: The idea is that each module has its own Γ, and recursively loading a dependency shouldn't put the things I've just defined in my own module into scope in the dependency.

src/Expander.hs
david-christiansen: Good idea!

src/Expander.hs
david-christiansen: One downside to that approach is that sometimes the check for "can I run" makes a little progress. For instance, `AwaitingMacro` has a list of roots that need to be fully filled in, but when it checks whether it can run, it recalculates the fringe of the trees to avoid retraversals. I don't know that we can do this with just a `Reader`-style effect.

src/Expander.hs
david-christiansen: Exactly.

src/Expander.hs
david-christiansen: Luckily it's 2020 and we have `-XBlockArguments`, so there's no need for an infix application operator to say "here comes a `do`!".

src/Expander.hs
langston-barrett: Then they aren't "internal errors" but rather "user errors" :-)

david-christiansen: I agree, and that's why it's now fixed :-)

src/Expander.hs
david-christiansen: Done!

PR #28, opened 2020-02-03
adding some comments

gelisam:
I was reading the examples which use `{bound,free}-identifier=?` and I took the opportunity to add a few comments here and there. And to simplify some occurrences of `[quote a]` with `'a`. Probably more to come as I continue to read the examples.

### DIFF
--- a/examples/defun.kl
+++ b/examples/defun.kl
@@ -32,7 +32,8 @@
     [[cons a d]
      (syntax-case d
        [() a]
-       [_ (last-stx d)])]
+       [_ (last-stx d)])]  -- note that this is a recursive call to last-stx, whereas earlier (e.g. in n-ary-app.kl)
+                           -- we'd have to return the syntax of a macro call to last-stx.
     [_ stx])]
 
 [example (last-stx '(a b c d e f g))]
--- a/examples/defun.kl
+++ b/examples/defun.kl
@@ -26,7 +26,6 @@
 
 [example (const 'a 'b)]
 
--- (last-stx foo bar baz) => baz
 [defun last-stx (stx)
   (syntax-case stx
     [[cons a d]
--- a/examples/defun.kl
+++ b/examples/defun.kl
@@ -8,6 +8,13 @@
      [lambda (rec) (fun [lambda (x) (rec rec x)])])]]
 
 [define-macros
+  -- (defun f (x y z)
+  --   (... (f foo bar baz) ...))
+  -- =>
+  -- (define f
+  --   (fix (lambda (f)
+  --          (lambda (x y z)
+  --            ... (f foo bar baz) ...))))
   ([defun
     [lambda (stx)
       (syntax-case stx
@@ -19,6 +26,7 @@
 
 [example (const 'a 'b)]
 
+-- (last-stx foo bar baz) => baz
 [defun last-stx (stx)
   (syntax-case stx
     [[cons a d]
--- a/examples/n-ary-app.kl
+++ b/examples/n-ary-app.kl
@@ -5,6 +5,7 @@
 
 
 [define-macros
+  -- (my-app f foo bar baz) => (((f foo) bar) baz)
   ([my-app
     [lambda [stx]
       (syntax-case stx
@@ -22,6 +23,12 @@
                     more-args
                     stx]
                   stx]]])])])]]
+   -- (my-lam (x y z) ...)
+   -- =>
+   -- (lambda (x)
+   --   (lambda (y)
+   --     (lambda (z)
+   --       ...)))
    [my-lam
     [lambda [stx]
      (syntax-case stx
@@ -37,6 +44,7 @@
            [pure [list-syntax [[quote lambda] args body] stx]]])])]])]
 
 [define-macros
+  -- (f foo bar baz) => (my-app foo bar baz)
   ([#%app
     [lambda [stx]
      (syntax-case stx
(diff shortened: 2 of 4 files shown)


examples/defun.kl
```diff
@@ -19,12 +26,14 @@
...
 
+-- (last-stx foo bar baz) => baz
 [defun last-stx (stx)
   (syntax-case stx
     [[cons a d]
      (syntax-case d
        [() a]
-       [_ (last-stx d)])]
+       [_ (last-stx d)])]  -- note that this is a recursive call to last-stx, whereas earlier (e.g. in n-ary-app.kl)
+                           -- we'd have to return the syntax of a macro call to last-stx.
```
gelisam: Speaking of which, how come that trick works in `n-ary-app.kl`? In the same manner we can write

    (let ((x 2))
      (let ((x (+ x x)))
        x))

And the `x` in `(+ x x)` refers to the outer `x`, I would expect to be able to write something like

    (define-macro (my-macro) #'2)
    (define-macro (my-macro) #'(+ (my-macro) (my-macro)))

and to have the `my-macro` in `#'(+ (my-macro) (my-macro))` refer to the first definition of `my-macro` (unless `define-macro` is like a `letrec`, but I don't think that's the case otherwise we wouldn't need `defun`). In the case of `my-app`, since there is no previous definition of `my-app`, that should be an undefined identifier, I don't think the body of `my-app` should be able to refer to the not-yet-in-scope `my-app`.

Am I understanding this all wrong? Or did we perhaps accidentally add a bug in our scope-set implementation somewhere in order to make the `n-ary-app.kl` test pass when it shouldn't? If so, I think we should just make `define-macro` behave like `letrec`, fix the bug, and skip `defun`.

david-christiansen: The key here is the way that macro scoping interacts with program scoping.

Before I added the recursive let form on the type checker branch, there were no ordinary recursive definitions in Klister. This was for the sake of minimality, because a Y-combinator is easy to write and stick behind a macro. But part of the point of typed lambda calculi is to rule out Y, so the type system required that addition (also to make the H-M let-generalization do the right thing).

Recursive macro definitions are different from recursive value definitions. A recursive value is defined in terms of itself - its definition refers to its own name. This reference takes the form of a use, and its meaning is given by substitution. A recursive macro is merely able to *mention* itself, and it can emit quoted code that refers to the same macro. But it's not recursive in the sense that a recursive value is.

david-christiansen:
Are you ready for a merge, or do you want to do more based on our conversation above?

examples/defun.kl
```diff
@@ -19,12 +26,14 @@
 
 [example (const 'a 'b)]
 
+-- (last-stx foo bar baz) => baz
```
david-christiansen: This comment isn't really true. The issue is that `last-stx` is not a macro, it's a function. Its type would be `Syntax -> Syntax`, and its behavior is to return the last element in its argument. Here's a Haskell equivalent:

```
lastStx :: Syntax -> Syntax
lastStx (Syntax ctx (List xs)) = last xs
  where
    last (a:[]) = a
    last (a:d) = lastStx (Stx ctx d)
```

examples/n-ary-app.kl
```diff
@@ -16,45 +17,53 @@
...
                 [cons-list-syntax
-                  [quote my-app]
+                  'my-app
                   [cons-list-syntax
-                    [list-syntax [[quote #%app] fun arg0] stx]
+                    [list-syntax ['#%app fun arg0] stx]
                     more-args
                     stx]
                   stx]]])])])]]
+   -- (my-lam (x y z) ...)
```
david-christiansen: I would write this
```
(my-lam (x y z) body) => (lambda (x) (lambda (y) (lambda (z) body))) 
```
because I worry that the `...` would mean to readers that the argument list could be repeated 0 or more times.

src/Core.hs
```diff
@@ -84,7 +84,7 @@ data CoreF core
   | CoreSyntaxError (SyntaxError core)  -- :: Macro a
   | CoreSendSignal core                 -- :: Signal -> Macro ()
   | CoreWaitSignal core                 -- :: Signal -> Macro ()
-  | CoreIdentEq HowEq core core
+  | CoreIdentEq HowEq core core         -- :: HowEq -> Syntax -> Syntax -> Macro Bool
```
david-christiansen: The HowEq is insterted by the parser, so the types at runtime are `Syntax -> Syntax -> Macro Bool` instead.

examples/defun.kl
gelisam: Would it be correct to say that `define-macros` is similar to `letrec` in that it makes the identifier available within the scope of the definition, except that `letrec` makes the identifier available in the current phase whereas `define-macros` only makes it available in the next phase?

david-christiansen: I don't think so, because the meaning of "making it available" is quite different.  A `letrec` introduces a recursive binding (recursion at core language level), while `define-macros` enables quotations in its body to refer to the macro's name (recursion at user syntax level). So there's a use-mention distinction going on, but also a conceptual difference.

examples/defun.kl
gelisam: removed; there is already an `[example (last-stx '(a b c d e f g))]` immediate afterwards anyway.

examples/n-ary-app.kl
gelisam: fixed

PR #29, opened 2020-02-07
`bound-identifier=?` bug

gelisam:


examples/bound-identifier.kl
```diff
@@ -0,0 +1,14 @@
...
+(define-macros
+  ((m (lambda (stx)
+        (syntax-case stx
+          ((list (_ x y))
+           (>>= (bound-identifier=? x y)
+                (lambda (bool)
+                  (pure (if bool ''t ''f))))))))))
+
+(example (m x x))
+(example (m x y))
```
gelisam: Should produce `'t` and `'f`, but produces `'t` and `'t`

david-christiansen:
Nice catch! Bug is fixed. I'll merge after Travis is happy.

PR #30, opened 2020-02-07
make bound-identifier=? pure

gelisam:
`free-idenfitier=?` looks at the internal interpreter state (all the matching bindings in the current phase), and so it makes sense for it to be a side-effect in the `Macro` monad. However, `bound-identifier=?` only looks at the value of the identifiers being compared, so it doesn't need to be in the `Macro` monad.

Making it pure has two advantages. First, it is now possible to call `bound-identifier=?` outside of a macro. Second, it simplifies our paper, because we only need to worry about non-determinism coming from the `Macro` actions, so the fewer `Macro` actions the better.

### DIFF
--- a/commentary.md
+++ b/commentary.md
@@ -124,9 +124,9 @@ The macro monad includes operations to compare the binding information of identi
 
 #### Macro monad effects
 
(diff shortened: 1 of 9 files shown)


david-christiansen:
I don't think this is a good idea.

It happens to be the case that the internal implementation of bound-identifier=? does not use any surrounding macro state. But this is an implementation concern based on how the set-of-scopes expander works. The semantics are defined with respect to what would bind what in the current context, and this means that the semantics are monadic.

PR #31, opened 2020-02-09
Examples

gelisam:
I was writing the part of the paper which explains why we think our `bound-identifier=?` and `free-identifier=?` implementations are deterministic, and I wanted to give an example. I then wrote a lot of helper modules so that this example could be written more succinctly.


examples/bool.kl
```diff
@@ -0,0 +1,64 @@
...
+              (()
+               (pure '#false))
+              ((cons x xs)
+               (pure `(binary-or ,x
+                                 ,(cons-list-syntax 'or xs stx)))))))))))
+
+(example (binary-and #false #false))
+(example (binary-and #false #true))
+(example (binary-and #true #false))
+(example (binary-and #true #true))
```
gelisam: I checked the results using `klister-repl run`, but I didn't bother adding examples to `test/Test.hs`, it's just so time-consuming. Maybe we're due for an `assert` macro?

examples/bool.kl
```diff
@@ -0,0 +1,64 @@
...
+
+(define not
+  (lambda (b)
+    (if b #false #true)))
+
+(define binary-and
+  (lambda (x y)
+    (if x y #false)))
+
+(define binary-or
```
gelisam: I initially wanted to name those `&&` and `||`, but it looks like `|` is not allowed in an identifier? Is that special syntax for something?

examples/bound-vs-free.kl
```diff
@@ -0,0 +1,27 @@
+#lang "n-ary-app.kl"
+
```
gelisam: If I add `(import "quasiquote.kl")` here I get the following failure on this file and I really don't understand why.

```
Ambiguous identifier in phase p1 #[/Users/gelisam/working/haskell/klister/examples/quasiquote.kl:49.63-49.66]
                                  <#%app>
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"},Scope {scopeNum = 80, scopePurpose = "For declaration at siquote.kl:4.1-7.1"}], _phaseScopes = fromList [(Phase {phaseNum = 0},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"}], _phaseScopes = fromList [(Phase {phaseNum = 0},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"},Scope {scopeNum = 80, scopePurpose = "For declaration at siquote.kl:4.1-7.1"}], _phaseScopes = fromList [(Phase {phaseNum = 2},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"}], _phaseScopes = fromList [(Phase {phaseNum = 2},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"},Scope {scopeNum = 80, scopePurpose = "For declaration at siquote.kl:4.1-7.1"}], _phaseScopes = fromList [(Phase {phaseNum = 2},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
    ScopeSet {_universalScopes = fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 78, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/quasiquote.kl"},Scope {scopeNum = 79, scopePurpose = "For declaration at siquote.kl:3.1-4.1"}], _phaseScopes = fromList [(Phase {phaseNum = 2},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
user error
```

gelisam:
I am pleased to report that all the examples work fine if I merge the `feature/types` branch!

examples/meta-macro.kl
```diff
@@ -0,0 +1,30 @@
+#lang "n-ary-app.kl"
```
gelisam: I wrote these while trying to hunt down a bug which said

```
Internal error during expansion! This is a bug in the implementation.
Current context won't accept expressions
```

It turned out the bug had nothing to do with the complexities of macro-defining macros... you simply can't have a raw `(lambda (x) x)` at the top-level, you must wrap it in a `define`, `define-macros`, or `example` 😅

examples/meta-macro.kl
```diff
@@ -0,0 +1,30 @@
...
+                    ((list (_ macro-name))
+                     (pure `(define-macros
+                              ((,macro-name (lambda (stx)
+                                              (syntax-case stx
+                                                ((list (_ var-name value))
+                                                 (pure (list-syntax ('define var-name value) stx)))))))))))))))
+
+(meta-macro1 my-macro1 x 'bar)
+(my-macro1)
+(example x)  -- 'foo
```
gelisam: The fact that we get `'foo` here instead of `'bar` means that the `x` in the `meta-macro1` call is _not_ considered the same as the `x` in the `example` call. Is that the intended result? If find that counter-intuitive.

examples/bool.kl
david-christiansen: What about a "crash" primitive that terminates the program? `assert` can be nicely built on top of that.

examples/bool.kl
david-christiansen: Right now, Klister uses R6RS Scheme rules for identifiers, because it seemed to be a reasonable default. In Scheme, `|` is a delimiter for identifiers that contain non-identifier characters, like back-tick in F#. So I could name a function `|My awesome function!!!|`. I think we can just turn `|` into an identifier char with no prob if you want.

examples/bound-vs-free.kl
david-christiansen: That is confusing! At the very least, we need to improve this message for our own sake.

The way to debug this for now is to hack up `resolve` in `Expander.hs` to show which bindings each refers to, and then to follow that back through the code to see if it's really ambiguous. As a start, we should at least show the source locations of each of the bindings that we're referring to.

examples/meta-macro.kl
david-christiansen: Yeah, that needs a proper error constructor rather than just using `InternalError`.

examples/meta-macro.kl
david-christiansen: This is the intended result. The idea is that references to names don't get to know about the internal implementation details of macros, including what they bind. In other words, the only thing that gets to refer to a macro-introduced binding is a macro-introduced use site. What you're intending here is to actually break the hygiene mechanism and do a bit of variable capture.

Here are two Racket programs that might be instructive. The first has the behavior that you're seeing in Klister:
```
#lang racket/base
(require (for-syntax racket/base racket/match))

(define x 'foo)

(define-syntax meta-macro1
  (lambda (stx)
    (match (syntax-e stx)
      [(list _ macro-name var-name value)
       #`(define-syntax #,macro-name
           (lambda (stx)
             (datum->syntax stx (list #'define #'#,var-name #'#,value))))])))

(meta-macro1 my-macro1 x 'bar)
(my-macro1)
x
```

The second has the behavior that you're wanting:
```
#lang racket/base
(require (for-syntax racket/base racket/match))

#;
(define x 'foo)

(define-syntax meta-macro1
  (lambda (stx)
    (match (syntax-e stx)
      [(list _ macro-name var-name value)
       #`(define-syntax #,macro-name
           (lambda (stx)
             (syntax-local-introduce (datum->syntax stx (list #'define #'#,var-name #'#,value)))))])))

(meta-macro1 my-macro1 x 'bar)
(my-macro1)
x
```

There are two differences:
 1. The second version comments out the `(define x 'foo)` because Racket doesn't allow top-level names to be duplicated.
2. The second invokes `syntax-local-introduce`, which breaks hygiene at one step of macro expansion.

`syntax-local-introduce` basically means "ignore hygiene right here". The way to implement it is to have it flip the current expansion scope. Then, the expander will flip it back when done. If we want it, and I think we do, then its type is `Syntax -> Macro Syntax`.

examples/define-syntax-rule.kl
```diff
@@ -0,0 +1,33 @@
+#lang "n-ary-app.kl"
```
gelisam: If I'm going to be writing meta-macros, I might as well write useful ones ;)

examples/bool.kl
gelisam: We already have `log` and `syntax-error`, I think that's good enough for writing `assert` in Klister

examples/bool.kl
gelisam: I'm fine either way, thanks for the explanation.

examples/bool.kl
david-christiansen: I think `crash` needs to be `forall a . Syntax -> a` rather than `forall a . Syntax -> Macro a`

examples/meta-macro.kl
david-christiansen: This got fixed in the type checker branch, and is now merged.

gelisam:
Now with golden tests! Next: update this branch to work with the new typechecker we just merged.

src/Pretty.hs
```diff
@@ -275,12 +275,8 @@ instance Pretty VarInfo SrcPos where
...
 
 instance Pretty VarInfo a => Pretty VarInfo (Stx a) where
-  pp env (Stx _ loc v) =
-    text "#" <>
-    (align . group)
-      (text "[" <> pp env loc <> text "]" <> line' <> text "<" <>
-       align (pp env v) <>
-       text ">")
+  pp env (Stx _ _ v) =
+    text "'" <> pp env v
```
gelisam: This may be a controversial change. Variables used to be pretty-printed as `#[/.../file.kl:11.36-11.37]<foo>`, now I'm pretty-printing them as `'foo`. I imagine that some find the location information helpful when debugging, but I personally find it extremely unreadable. CI didn't like it either, because the golden files had my home folder baked in that `/.../file.kl` path. I could fix that by only including the basename of the file... but I find `'foo` a lot more readable. Let me know what you think!

gelisam:
The examples all type-check, and the golden tests now include the types as well, so that should help us catch regressions to the type-checker in the future.

tests/Golden.hs
```diff
@@ -0,0 +1,77 @@
...
+  return $ testGroup "Golden tests"
+    [ goldenVsString testName goldenFile $ do
+        outputLines <- execWriterT $ runExamples klisterFile
+        pure . Text.encodeUtf8 $ outputLines
+    | klisterFile <- klisterFiles
+    , let testName = takeBaseName klisterFile
+    , let goldenFile = replaceExtension klisterFile ".golden"
+    ]
+
+runExamples :: FilePath -> WriterT Text IO ()
```
gelisam: `runExamples` and `expandFile` are both adapted from code in `repl/Main.hs`. I didn't bother moving it to a library function because I couldn't reuse it _exactly_ as is; I'm accumulating the results to a file rather than stdout, and I didn't bother printing the resulting core. Maybe I should?

src/Pretty.hs
david-christiansen: There was a good reason for sticking this information on them: it makes it possible to trace where they're from. And this change doesn't just strip that information from identifiers, it does it from _all_ syntax objects!

What about doing the following instead?
 * Show the paths relative to that of the current module being expanded (this should make them shorter and also fix the CI issue
 * Add a `Symbol` and/or `SyntaxContents` type, and then have a way to extract them from identifiers/arbitrary syntax objects, and use this in tests
 * Add an extra argument to `pp` that consists of pretty-printer options, and then have the display of source locations be an option that can be controlled by a command-line flag

tests/Golden.hs
david-christiansen: It's unfortunate that Haskell doesn't have an abstraction like Scheme ports, which would let us combine these better. I think it's fine to have this little bit of duplication, personally.

src/Pretty.hs
gelisam: That current module isn't available in `pp`, so for now, I print the basename but not the folder. Is that good enough to merge this?

david-christiansen:
Would you mind augmenting the README with instructions for regenerating the golden files?

david-christiansen:
Would you mind augmenting the README with instructions for regenerating the golden files?

gelisam:
> Would you mind augmenting the README with instructions for regenerating the golden files?

Even better: I added a `examples/regenerate-golden-files.sh` script which regenerates them.

david-christiansen:
I'm not sure why the golden tests are failing. Could it be some Unicode encoding issue with respect to your machine vs Travis's machines, perhaps? I really can't figure out why we get half a green check.

gelisam:
I do think there is a difference between my terminal's unicode settings and TravisCI's, but I don't think that's the problem. The TravisCI logs print the type of `(const 'a 'b)` as `âÎ±0. Î±0`, whereas on my machine it is printed as `\226\136\128\206\177\&0. \206\177\&0`. Neither are particularly readable, but by regenerating the golden files, I can see that the actual type is `∀α0. α0`. (Note that the `\&` expands to the empty string, it's there to make sure the result is `"\177" ++ "0"`, not the single-character string `"\1770"`.)

We're getting a half checkmark because the tests pass on this branch, but not when we merge this branch into `master`. I am guessing this is because the type checker has changed since I generated the golden files. Here are the types which were inferred when I generated them:

    [defun const (x y) x]
    (const 'a 'b) : Syntax

    [define const [lambda (x y) x]]
    (const 'a 'b) : Syntax

    [define compose [lambda (f g x) (f (g x))]]
    (compose id) : ∀α β. ((α → β) → (α → β))
    (compose id id) : ∀α. (α → α)
    (compose id id id) : ∀α. (α → α)

    (define compose* (flet (comp (f g x) (f (g x))) comp))
    (compose* id) : ∀α β. ((α → β) → (α → β))
    (compose* id id) : ∀α. (α → α)
    (compose* id id id) : ∀α. (α → α)

And here are the types which are inferred now:

    [defun const (x y) x]
    (const 'a 'b) : ∀α. α  -- wrong!

    [define const [lambda (x y) x]]
    (const 'a 'b) : Syntax

    [define compose [lambda (f g x) (f (g x))]]
    (compose id) : ∀α β. ((α → β) → (α → β))
    (compose id id) : ∀α. (α → α)
    (compose id id id) : ∀α. (α → α)

    (define compose* (flet (comp (f g x) (f (g x))) comp))
    (compose* id) : ∀α. α  -- wrong!
    (compose* id id) : ∀α. α  -- wrong!
    (compose* id id id) : ∀α. α  -- wrong!

So I guess we accidentally merged a regression in the type checker, and this half checkmark is catching it after the fact. Apparently, now that we're interleaving macro-expansion and type-checking, all the `flet`s are incorrectly inferred to have type `∀α. α`! (`defun` is defined in terms of `flet`)

david-christiansen:
Should be fixed with these commits. Yay tests!

david-christiansen:
Thanks so much for helping out the testing infrastructure like you did! That will be extremely valuable going forward!

PR #34, opened 2020-02-29
Datatypes

david-christiansen:
I want to make sure that our metaprogramming-time splitting on types works ergonomically for an open universe of (nominal) datatypes. So I figured the best way to do that is to add simple ML-style datatypes to Klister. See `examples/datatypes.kl`  for how it works. Another thing that' s necessary is a bit of macro writing to simplify some of the syntax, so that we can write
```
(datatype Nat
  zero
  (add1 Nat))
``` 
instead of
```
(datatype (Nat)
  (zero)
  (add1 Nat))
```

TODO:
 * [x] Add case analysis for datatypes to macro expander and type checker
 * [x] Better examples/tests
 * [x] Clean up some cruft that is no longer needed
 * [ ] Simpler datatype definitions/case analysis as macro
   * [x] Macros in pattern positions
   * [ ] Multi-declaration decl macros (e.g. expand to a datatype and some macros about it)


src/Core.hs
```diff
@@ -35,14 +37,20 @@ newtype MacroVar = MacroVar Unique
...
 
-data Pattern
-  = PatternIdentifier Ident Var
-  | PatternEmpty
-  | PatternCons Ident Var Ident Var
-  | PatternList [(Ident, Var)]
-  | PatternAny
+data ConstructorPattern
+  = ConstructorPattern !Constructor [(Ident, Var)]
+  | AnyConstructor Ident Var
```
langston-barrett: Checking my understanding: Is this the `a` in something like
```haskell
   case x of
     a -> a -- you've learned nothing about x
```

src/Evaluator.hs
```diff
@@ -263,28 +268,47 @@ withScopeOf scope expr = do
     Syntax (Stx scopeSet loc _) ->
       pure $ ValueSyntax $ Syntax $ Stx scopeSet loc expr
 
-doCase :: Value -> [(Pattern, Core)] -> Eval Value
+doDataCase :: Value -> [(ConstructorPattern, Core)] -> Eval Value
+doDataCase v0 [] = throwError (EvalErrorCase v0)
```
langston-barrett: Having an empty list of cases seems more like a compile-time error than a run-time error, we could use a `NonEmpty` in `CoreDataCase` to ensure that we catch it earlier.

examples/datatypes.kl
```diff
@@ -0,0 +1,54 @@
...
+  (zero)
+  (add1 (Nat)))
+
+(datatype (List A)
+  (nil)
+  (:: A (List A)))
+
+(datatype (Tree A)
+  (leaf)
+  (branch A (Tree A) A))
```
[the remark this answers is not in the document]
david-christiansen: Yeah, this was an example fail :-) Thanks for catching it!

src/Core.hs
david-christiansen: Yes! It generalizes catchall patterns.

src/Evaluator.hs
david-christiansen: This should be dynamic, for two reasons:
 1. It's the base case for recursion when no pattern matches (we have no exhaustiveness checking ATM)
 2. Empty cases are the natural eliminator for constructorless datatypes, witnessing unreachability.

david-christiansen:
My next task is making patterns in `case` into a macro expansion position. That'll make it reasonable to do things like deep pattern matching as a macro, or just eliminating mandatory parens around nullary constructors.

david-christiansen:
Merging now, because the extra feature isn't worth diverging branches.

PR #37, opened 2020-03-28
declaration groups

gelisam:
We want to allow macros running in a declaration context to be able to produce more than one declaration.

TODO:

- [x] Make ModuleBody a binary tree instead of a list
- [x] Adapt the task queue to support filling-in branches of that tree
- [x] Make sure all the names in the left branch are bound before expanding the right branch
- [x] Fix the failing tests
- [x] ~Add `no-decls` and `decls` primitives~ Added `group` directly
- [x] ~Implement the `group` macro in terms of `no-decls` and `decls`~
- [x] As a test, implement a macro which uses `group` to generate multiple declarations
- [x] Fix the `PrettyBinder VarInfo [CompleteDecl]` instance
- [x] Avoid using the word "environment" incorrectly

### DIFF
--- a/examples/double-define.kl
+++ b/examples/double-define.kl
@@ -1,0 +1,12 @@
+#lang kernel
+
+(import "define-syntax-rule.kl")
+
+(define-syntax-rule (double-define x y value)
+  (group
+    (define x value)
+    (define y value)))
+
+(double-define a b 1)
+(example a)
+(example b)
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -95,7 +95,7 @@ mainWithOptions opts =
           mn <- moduleNameFromPath file
           ctx <- mkInitContext mn
           void $ execExpand initializeKernel ctx
-          st <- flip execExpand ctx $ completely $ do
+          st <- flip execExpand ctx $ do
             visit mn >> getState
           case st of
             Left err -> prettyPrintLn err *> fail ""
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -95,7 +95,7 @@ mainWithOptions opts =
           mn <- moduleNameFromPath file
           ctx <- mkInitContext mn
           void $ execExpand initializeKernel ctx
-          st <- execExpand (visit mn >> getState) ctx
+          st <- execExpand (fullyVisit mn >> getState) ctx
           case st of
             Left err -> prettyPrintLn err *> fail ""
             Right result ->
@@ -127,7 +127,7 @@ repl ctx startWorld = do
            Right ok ->
              do putStrLn "Parser output:"
                 T.putStrLn (syntaxText ok)
-                c <- execExpand (expandExpr ok) ctx
+                c <- execExpand (fullyExpandExpr ok) ctx
                 case c of
                   Left err -> putStr "Expander error: " *>
                               prettyPrintLn err
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -95,7 +95,8 @@ mainWithOptions opts =
           mn <- moduleNameFromPath file
           ctx <- mkInitContext mn
           void $ execExpand initializeKernel ctx
-          st <- execExpand (fullyVisit mn >> getState) ctx
+          st <- flip execExpand ctx $ completely $ do
+            visit mn >> getState
           case st of
             Left err -> prettyPrintLn err *> fail ""
             Right result ->
@@ -127,7 +128,7 @@ repl ctx startWorld = do
            Right ok ->
              do putStrLn "Parser output:"
                 T.putStrLn (syntaxText ok)
-                c <- execExpand (fullyExpandExpr ok) ctx
+                c <- flip execExpand ctx $ completely $ expandExpr ok
                 case c of
                   Left err -> putStr "Expander error: " *>
                               prettyPrintLn err
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -609,6 +609,11 @@ initializeKernel = do
             inEarlierPhase $
               expandDeclForms subDest mempty vp =<< addRootScope subDecls
         )
+      , ( "group"
+        , \dest vp stx -> do
+            (_ :: Syntax, decls) <- mustHaveShape stx
+            expandDeclForms dest mempty vp decls
+        )
       ]
       where
         importSpec :: Syntax -> Expand ImportSpec
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -116,9 +116,10 @@ expandModule thisMod src = do
     lang <- mustBeModName (view moduleLanguage src)
     initializeLanguage lang
     declTreeDest <- liftIO $ newDeclTreePtr
+    vp <- newDeclValidityPtr
     modifyState $ set expanderModuleTop $ Just declTreeDest
     decls <- addModuleScope (view moduleContents src)
-    forkExpandDecls declTreeDest decls
+    expandDeclForms declTreeDest mempty vp decls
     expandTasks
     body <- getDeclGroup declTreeDest
     let modName = view moduleSource src
@@ -465,19 +466,21 @@ initializeKernel = do
 
                 resolveImports imports
 
-                forkExpandDecls bodyPtr body
+                vp <- newDeclValidityPtr
+                expandDeclForms bodyPtr mempty vp body
 
                 buildExports exports
                 pure ()
         )
       ]
 
-    declPrims :: [(Text, Scope -> DeclTreePtr -> DeclValidityPtr -> Syntax -> Expand ())]
+    declPrims :: [(Text, DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ())]
     declPrims =
       [ ( "define"
-        , \ sc dest vp stx -> do
+        , \ dest scs vp stx -> do
             p <- currentPhase
             Stx _ _ (_, varStx, expr) <- mustHaveEntries stx
+            sc <- freshScope $ T.pack $ "For definition at " ++ shortShow (stxLoc stx)
             x <- flip addScope' sc <$> mustBeIdent varStx
             b <- freshBinding
             addDefinedBinding x b
@@ -494,15 +497,16 @@ initializeKernel = do
             modifyState $ over (expanderDefTypes . at ph . non Env.empty) $
               Env.insert var x schPtr
             forkGeneralizeType exprDest t schPtr
-            linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
+            linkDeclValidity vp (ScopeSet.insertAtPhase p sc scs)
         )
       , ( "datatype"
-        , \ sc dest vp stx -> do
-            Stx scs loc (_ :: Syntax, more) <- mustBeCons stx
-            Stx _ _ (nameAndArgs, ctorSpecs) <- mustBeCons (Syntax (Stx scs loc (List more)))
+        , \ dest scs vp stx -> do
+            Stx stxScs loc (_ :: Syntax, more) <- mustBeCons stx
+            Stx _ _ (nameAndArgs, ctorSpecs) <- mustBeCons (Syntax (Stx stxScs loc (List more)))
             Stx _ _ (name, args) <- mustBeCons nameAndArgs
             typeArgs <- for (zip [0..] args) $ \(i, a) ->
               prepareTypeVar i a
+            sc <- freshScope $ T.pack $ "For datatype at " ++ shortShow (stxLoc stx)
             let typeScopes = map fst typeArgs ++ [sc]
             realName <- mustBeIdent (addScope' name sc)
             p <- currentPhase
@@ -533,15 +537,16 @@ initializeKernel = do
               Just info
 
 
-            forkEstablishConstructors sc vp d ctors
+            forkEstablishConstructors (ScopeSet.insertAtPhase p sc scs) vp d ctors
 
             linkOneDecl dest (Data realName (view datatypeName d) (fromIntegral arity) ctors)
         )
       , ( "define-macros"
-        , \ sc dest vp stx -> do
+        , \ dest scs vp stx -> do
             Stx _ _ (_ :: Syntax, macroList) <- mustHaveEntries stx
             Stx _ _ macroDefs <- mustBeList macroList
             p <- currentPhase
+            sc <- freshScope $ T.pack $ "For macros at " ++ shortShow (stxLoc stx)
             macros <- for macroDefs $ \def -> do
               Stx _ _ (mname, mdef) <- mustHaveEntries def
               theName <- flip addScope' sc <$> mustBeIdent mname
@@ -554,36 +559,38 @@ initializeKernel = do
               bind b $ EIncompleteMacro v theName macroDest
               return (theName, v, macroDest)
             linkOneDecl dest $ DefineMacros macros
-            linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
+            linkDeclValidity vp (ScopeSet.insertAtPhase p sc scs)
         )
       , ( "example"
-        , \ sc dest vp stx -> do
+        , \ dest scs vp stx -> do
             p <- currentPhase
             Stx _ _ (_ :: Syntax, expr) <- mustHaveEntries stx
             exprDest <- liftIO $ newSplitCorePtr
             sch <- liftIO newSchemePtr
             linkOneDecl dest (Example (view (unSyntax . stxSrcLoc) stx) sch exprDest)
+            sc <- freshScope $ T.pack $ "For example at " ++ shortShow (stxLoc stx)
             t <- inTypeBinder do
               t <- Ty . TMetaVar <$> freshMeta
               forkExpandSyntax (ExprDest t exprDest) (addScope p expr sc)
               return t
             forkGeneralizeType exprDest t sch
-            linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
+            linkDeclValidity vp (ScopeSet.insertAtPhase p sc scs)
         )
       , ( "import"
          -- TODO Make import spec language extensible and use bindings rather than literals
-        , \sc dest vp stx -> do
-            Stx scs loc (_ :: Syntax, toImport) <- mustHaveEntries stx
+        , \dest scs vp stx -> do
+            Stx stxScs loc (_ :: Syntax, toImport) <- mustHaveEntries stx
             spec <- importSpec toImport
             modExports <- getImports spec
+            sc <- freshScope $ T.pack $ "For import at " ++ shortShow (stxLoc stx)
             flip forExports_ modExports $ \p x b -> inPhase p do
-              imported <- addRootScope' $ addScope' (Stx scs loc x) sc
+              imported <- addRootScope' $ addScope' (Stx stxScs loc x) sc
               addImportBinding imported b
             linkOneDecl dest (Import spec)
-            linkDeclValidity vp (ScopeSet.singleUniversalScope sc)
+            linkDeclValidity vp (ScopeSet.insertUniversally sc scs)
         )
       , ( "export"
-        , \_sc dest vp stx -> do
+        , \dest _scs vp stx -> do
             Stx _ _ (_, protoSpec) <- mustBeCons stx
             exported <- exportSpec stx protoSpec
             es <- getExports exported
@@ -592,12 +599,12 @@ initializeKernel = do
             linkDeclValidity vp mempty
         )
       , ( "meta"
-        , \sc dest vp stx -> do
+        , \dest scs vp stx -> do
             Stx _ _ (_ :: Syntax, subDecl) <- mustHaveEntries stx
             subDest <- liftIO newDeclTreePtr
             linkOneDecl dest (Meta subDest)
             inEarlierPhase $
-              forkExpandOneDecl subDest sc vp =<< addRootScope subDecl
+              expandDeclForm subDest scs vp =<< addRootScope subDecl
         )
       ]
       where
@@ -668,6 +675,9 @@ initializeKernel = do
                 Stx _ _ y' <- mustBeIdent y
                 pure (x', y')
             getRenames _ = throwError $ NotExportSpec blame
+
+        stxLoc :: Syntax -> SrcLoc
+        stxLoc (Syntax (Stx _ srcloc _)) = srcloc
 
     exprPrims :: [(Text, Ty -> SplitCorePtr -> Syntax -> Expand ())]
     exprPrims =
@@ -985,7 +995,7 @@ initializeKernel = do
       bind b val
       addToKernel name runtime b
 
-    addDeclPrimitive :: Text -> (Scope -> DeclTreePtr -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
+    addDeclPrimitive :: Text -> (DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
     addDeclPrimitive name impl = do
       let val = EPrimDeclMacro impl
       b <- freshBinding
@@ -1073,29 +1083,9 @@ addRootScope' stx = do
   return (addScope' stx rsc)
 
 
-forkExpandOneDecl :: DeclTreePtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
-forkExpandOneDecl dest sc declValidityPtr stx = do
-  forkExpanderTask $ ExpandDeclTree dest sc stx declValidityPtr
-
-forkExpandDecls :: DeclTreePtr -> Syntax -> Expand ()
-forkExpandDecls dest (Syntax (Stx _ _ (List []))) =
-  linkDeclTree dest DeclTreeLeaf
-forkExpandDecls dest (Syntax (Stx scs loc (List (d:ds)))) = do
-  -- Create a scope for this new declaration
-  sc <- freshScope $ T.pack $ "For declaration at " ++ shortShow (stxLoc d)
-  carDest <- liftIO $ newDeclTreePtr
-  cdrDest <- liftIO $ newDeclTreePtr
-  declValidityPtr <- newDeclValidityPtr
-  linkDeclTree dest (DeclTreeBranch carDest cdrDest)
-  forkExpanderTask $
-    ExpandDependentDeclTree cdrDest
-      (Syntax (Stx scs loc (List ds)))
-      declValidityPtr
-  forkExpandOneDecl carDest sc declValidityPtr =<< addRootScope d
-  where
-    stxLoc (Syntax (Stx _ srcloc _)) = srcloc
-forkExpandDecls _dest _other =
-  error "TODO real error message - malformed module body"
+forkExpandDeclForms :: DeclTreePtr -> DeclValidityPtr -> DeclValidityPtr -> Syntax -> Expand ()
+forkExpandDeclForms dest ivp ovp stx = do
+  forkExpanderTask $ ExpandDeclForms dest ivp ovp stx
 
 
 identifierHeaded :: Syntax -> Maybe Ident
@@ -1127,8 +1117,8 @@ runTask (tid, localData, task) = withLocal localData $ do
       case dest of
         ExprDest t d ->
           expandOneExpression t d stx
-        DeclTreeDest d sc vp ->
-          expandOneDeclTreeNode sc d stx vp
+        DeclTreeDest d scs vp ->
+          expandDeclForm d scs vp stx
         TypeDest d ->
           expandOneType d stx
         PatternDest exprT scrutT d ->
@@ -1191,15 +1181,20 @@ runTask (tid, localData, task) = withLocal localData $ do
         Just _v -> do
           bind b $ EVarMacro x
           forkExpandSyntax (ExprDest t dest) stx
-    ExpandDeclTree dest sc stx vp ->
-      expandOneDeclTreeNode sc dest stx vp
-    ExpandDependentDeclTree dest stx waitingOn -> do
-      readyYet <- view (expanderDeclValidities . at waitingOn) <$> getState
+    ExpandDeclForm dest ivp ovp stx -> do
+      readyYet <- view (expanderDeclValidities . at ivp) <$> getState
       case readyYet of
         Nothing ->
           stillStuck tid task
-        Just scopeSet ->
-          forkExpandDecls dest (addScopes stx scopeSet)
+        Just scs ->
+          expandDeclForm dest scs ovp (addScopes stx scs)
+    ExpandDeclForms dest ivp ovp stx -> do
+      readyYet <- view (expanderDeclValidities . at ivp) <$> getState
+      case readyYet of
+        Nothing ->
+          stillStuck tid task
+        Just scs ->
+          expandDeclForms dest scs ovp (addScopes stx scs)
     InterpretMacroAction dest act outerKont -> do
       interpretMacroAction act >>= \case
         Left (signal, innerKont) -> do
@@ -1247,7 +1242,7 @@ runTask (tid, localData, task) = withLocal localData $ do
           specialize sch >>= unify eDest ty
           saveExprType eDest ty
           linkExpr eDest (CoreVar x)
-    EstablishConstructors sc vp dt ctors -> do
+    EstablishConstructors scs vp dt ctors -> do
       ctors' <- sequenceA <$> for ctors \(cn, ctor, argTys) -> do
                   perhapsArgs <- sequenceA <$> traverse linkedType argTys
                   pure (fmap (\ts -> (cn, ctor, ts)) perhapsArgs)
@@ -1256,8 +1251,7 @@ runTask (tid, localData, task) = withLocal localData $ do
         Just ready -> do
           for_ ready \(cn, ctor, ts) ->
             addConstructor cn dt ctor ts
-          p <- currentPhase
-          linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
+          linkDeclValidity vp scs
     AwaitingPattern patPtr ty dest stx -> do
       view (expanderCompletedPatterns . at patPtr) <$> getState >>=
         \case
@@ -1331,8 +1325,8 @@ problemContext (TypeDest _) = TypeCtx
 problemContext (ExprDest _ _) = ExpressionCtx
 problemContext (PatternDest _ _ _) = PatternCaseCtx
 
-requireDeclarationCtx :: Syntax -> MacroDest -> Expand (Scope, DeclTreePtr, DeclValidityPtr)
-requireDeclarationCtx _ (DeclTreeDest dest sc vp) = return (sc, dest, vp)
+requireDeclarationCtx :: Syntax -> MacroDest -> Expand (ScopeSet, DeclTreePtr, DeclValidityPtr)
+requireDeclarationCtx _ (DeclTreeDest dest scs vp) = return (scs, dest, vp)
 requireDeclarationCtx stx other =
   throwError $ WrongMacroContext stx DeclarationCtx (problemContext other)
 
@@ -1399,8 +1393,8 @@ expandOneForm prob stx
         EPrimModuleMacro _ ->
           throwError $ WrongMacroContext stx (problemContext prob) ModuleCtx
         EPrimDeclMacro impl -> do
-          (sc, dest, vp) <- requireDeclarationCtx stx prob
-          impl sc dest vp stx
+          (scs, dest, vp) <- requireDeclarationCtx stx prob
+          impl dest scs vp stx
         EPrimTypeMacro impl -> do
           dest <- requireTypeCtx stx prob
           impl dest stx
@@ -1471,9 +1465,24 @@ expandOneForm prob stx
           Id _ -> error "Impossible happened - identifiers are identifier-headed!"
 
 
-expandOneDeclTreeNode :: Scope -> DeclTreePtr -> Syntax -> DeclValidityPtr -> Expand ()
-expandOneDeclTreeNode sc dest stx vp =
-  expandOneForm (DeclTreeDest dest sc vp) stx
+expandDeclForm :: DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ()
+expandDeclForm dest scs vp stx =
+  expandOneForm (DeclTreeDest dest scs vp) stx
+
+expandDeclForms :: DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ()
+expandDeclForms dest scs vp (Syntax (Stx _ _ (List []))) = do
+  linkDeclTree dest DeclTreeLeaf
+  linkDeclValidity vp scs
+expandDeclForms dest scs vp (Syntax (Stx stxScs loc (List (d:ds)))) = do
+  carDest <- liftIO $ newDeclTreePtr
+  cdrDest <- liftIO $ newDeclTreePtr
+  carValidityPtr <- newDeclValidityPtr
+  linkDeclTree dest (DeclTreeBranch carDest cdrDest)
+  forkExpandDeclForms cdrDest carValidityPtr vp
+    (Syntax (Stx stxScs loc (List ds)))
+  expandDeclForm carDest scs carValidityPtr =<< addRootScope d
+expandDeclForms _dest _scs _vp _stx =
+  error "TODO real error message - malformed module body"
 
 
 -- | Link the destination to a literal signal object
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -1478,6 +1478,11 @@ expandDeclForms dest scs vp (Syntax (Stx stxScs loc (List (d:ds)))) = do
   cdrDest <- liftIO $ newDeclTreePtr
   carValidityPtr <- newDeclValidityPtr
   linkDeclTree dest (DeclTreeBranch carDest cdrDest)
+
+  -- This is where the ScopeSet is threaded through: from the input scs, to the
+  -- car's expandDeclForm task, who adds some Scopes to it and writes the result
+  -- to carValidityPtr, which is read (and blocked on) by the forkExpandDeclForms
+  -- task, who adds even more Scopes and writes the result to vp.
   forkExpandDeclForms cdrDest carValidityPtr vp
     (Syntax (Stx stxScs loc (List ds)))
   expandDeclForm carDest scs carValidityPtr =<< addRootScope d
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -115,12 +115,12 @@ expandModule thisMod src = do
   local (set (expanderLocal . expanderModuleName) thisMod) do
     lang <- mustBeModName (view moduleLanguage src)
     initializeLanguage lang
-    modBodyDest <- liftIO $ newModBodyPtr
-    modifyState $ set expanderModuleTop $ Just modBodyDest
+    declTreeDest <- liftIO $ newDeclTreePtr
+    modifyState $ set expanderModuleTop $ Just declTreeDest
     decls <- addModuleScope (view moduleContents src)
-    forkExpandDecls modBodyDest decls
+    forkExpandDecls declTreeDest decls
     expandTasks
-    body <- getBody modBodyDest
+    body <- getDeclGroup declTreeDest
     let modName = view moduleSource src
     let theModule = Module { _moduleName = modName
                            , _moduleImports = noImports
@@ -458,7 +458,7 @@ initializeKernel = do
               Just _ ->
                 error "TODO throw real error - already expanding a module"
               Nothing -> do
-                bodyPtr <- liftIO newModBodyPtr
+                bodyPtr <- liftIO newDeclTreePtr
                 modifyState $ set expanderModuleTop (Just bodyPtr)
                 Stx _ _ (_ :: Syntax, name, imports, body, exports) <- mustHaveEntries stx
                 _actualName <- mustBeIdent name
@@ -472,7 +472,7 @@ initializeKernel = do
         )
       ]
 
-    declPrims :: [(Text, Scope -> ModBodyPtr -> DeclValidityPtr -> Syntax -> Expand ())]
+    declPrims :: [(Text, Scope -> DeclTreePtr -> DeclValidityPtr -> Syntax -> Expand ())]
     declPrims =
       [ ( "define"
         , \ sc dest pdest stx -> do
@@ -595,7 +595,7 @@ initializeKernel = do
       , ( "meta"
         , \sc dest pdest stx -> do
             Stx _ _ (_ :: Syntax, subDecl) <- mustHaveEntries stx
-            subDest <- liftIO newModBodyPtr
+            subDest <- liftIO newDeclTreePtr
             linkOneDecl dest (Meta subDest)
             inEarlierPhase $
               forkExpandOneDecl subDest sc pdest =<< addRootScope subDecl
@@ -986,7 +986,7 @@ initializeKernel = do
       bind b val
       addToKernel name runtime b
 
-    addDeclPrimitive :: Text -> (Scope -> ModBodyPtr -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
+    addDeclPrimitive :: Text -> (Scope -> DeclTreePtr -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
     addDeclPrimitive name impl = do
       let val = EPrimDeclMacro impl
       b <- freshBinding
@@ -1074,22 +1074,22 @@ addRootScope' stx = do
   return (addScope' stx rsc)
 
 
-forkExpandOneDecl :: ModBodyPtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
+forkExpandOneDecl :: DeclTreePtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
 forkExpandOneDecl dest sc declValidityPtr stx = do
-  forkExpanderTask $ ExpandModBody dest sc stx declValidityPtr
+  forkExpanderTask $ ExpandDeclTree dest sc stx declValidityPtr
 
-forkExpandDecls :: ModBodyPtr -> Syntax -> Expand ()
+forkExpandDecls :: DeclTreePtr -> Syntax -> Expand ()
 forkExpandDecls dest (Syntax (Stx _ _ (List []))) =
-  linkModBody dest NoDecls
+  linkDeclTree dest DeclTreeLeaf
 forkExpandDecls dest (Syntax (Stx scs loc (List (d:ds)))) = do
   -- Create a scope for this new declaration
   sc <- freshScope $ T.pack $ "For declaration at " ++ shortShow (stxLoc d)
-  carDest <- liftIO $ newModBodyPtr
-  cdrDest <- liftIO $ newModBodyPtr
+  carDest <- liftIO $ newDeclTreePtr
+  cdrDest <- liftIO $ newDeclTreePtr
   declValidityPtr <- newDeclValidityPtr
-  linkModBody dest (Decls carDest cdrDest)
+  linkDeclTree dest (DeclTreeBranch carDest cdrDest)
   forkExpanderTask $
-    ExpandDependentModBody cdrDest sc
+    ExpandDependentDeclTree cdrDest sc
       (Syntax (Stx scs loc (List ds)))
       declValidityPtr
   forkExpandOneDecl carDest sc declValidityPtr =<< addRootScope d
@@ -1128,8 +1128,8 @@ runTask (tid, localData, task) = withLocal localData $ do
       case dest of
         ExprDest t d ->
           expandOneExpression t d stx
-        ModBodyDest d sc ph ->
-          expandOneModuleBodyNode sc d stx ph
+        DeclTreeDest d sc ph ->
+          expandOneDeclTreeNode sc d stx ph
         TypeDest d ->
           expandOneType d stx
         PatternDest exprT scrutT d ->
@@ -1192,9 +1192,9 @@ runTask (tid, localData, task) = withLocal localData $ do
         Just _v -> do
           bind b $ EVarMacro x
           forkExpandSyntax (ExprDest t dest) stx
-    ExpandModBody dest sc stx ph ->
-      expandOneModuleBodyNode sc dest stx ph
-    ExpandDependentModBody dest sc stx waitingOn -> do
+    ExpandDeclTree dest sc stx ph ->
+      expandOneDeclTreeNode sc dest stx ph
+    ExpandDependentDeclTree dest sc stx waitingOn -> do
       readyYet <- view (expanderDeclPhases . at waitingOn) <$> getState
       case readyYet of
         Nothing ->
@@ -1329,13 +1329,13 @@ addApp ctor (Syntax (Stx scs loc _)) args =
     app = Syntax (Stx scs loc (Id "#%app"))
 
 problemContext :: MacroDest -> MacroContext
-problemContext (ModBodyDest _ _ _) = DeclarationCtx
+problemContext (DeclTreeDest _ _ _) = DeclarationCtx
 problemContext (TypeDest _) = TypeCtx
 problemContext (ExprDest _ _) = ExpressionCtx
 problemContext (PatternDest _ _ _) = PatternCaseCtx
 
-requireDeclarationCtx :: Syntax -> MacroDest -> Expand (Scope, ModBodyPtr, DeclValidityPtr)
-requireDeclarationCtx _ (ModBodyDest dest sc ph) = return (sc, dest, ph)
+requireDeclarationCtx :: Syntax -> MacroDest -> Expand (Scope, DeclTreePtr, DeclValidityPtr)
+requireDeclarationCtx _ (DeclTreeDest dest sc ph) = return (sc, dest, ph)
 requireDeclarationCtx stx other =
   throwError $ WrongMacroContext stx DeclarationCtx (problemContext other)
 
@@ -1453,7 +1453,7 @@ expandOneForm prob stx
             Just other -> throwError $ ValueNotMacro other
   | otherwise =
     case prob of
-      ModBodyDest _ _ _ ->
+      DeclTreeDest _ _ _ ->
         throwError $ InternalError "All declarations should be identifier-headed"
       TypeDest _dest ->
         throwError $ NotValidType stx
@@ -1474,9 +1474,9 @@ expandOneForm prob stx
           Id _ -> error "Impossible happened - identifiers are identifier-headed!"
 
 
-expandOneModuleBodyNode :: Scope -> ModBodyPtr -> Syntax -> DeclValidityPtr -> Expand ()
-expandOneModuleBodyNode sc dest stx ph =
-  expandOneForm (ModBodyDest dest sc ph) stx
+expandOneDeclTreeNode :: Scope -> DeclTreePtr -> Syntax -> DeclValidityPtr -> Expand ()
+expandOneDeclTreeNode sc dest stx ph =
+  expandOneForm (DeclTreeDest dest sc ph) stx
 
 
 -- | Link the destination to a literal signal object
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -563,18 +563,16 @@ initializeKernel = do
         )
       , ( "example"
         , \ dest scs vp stx -> do
-            p <- currentPhase
             Stx _ _ (_ :: Syntax, expr) <- mustHaveEntries stx
             exprDest <- liftIO $ newSplitCorePtr
             sch <- liftIO newSchemePtr
             linkOneDecl dest (Example (view (unSyntax . stxSrcLoc) stx) sch exprDest)
-            sc <- freshScope $ T.pack $ "For example at " ++ shortShow (stxLoc stx)
             t <- inTypeBinder do
               t <- Ty . TMetaVar <$> freshMeta
-              forkExpandSyntax (ExprDest t exprDest) (addScope p expr sc)
+              forkExpandSyntax (ExprDest t exprDest) expr
               return t
             forkGeneralizeType exprDest t sch
-            linkDeclValidity vp (ScopeSet.insertAtPhase p sc scs)
+            linkDeclValidity vp scs
         )
       , ( "import"
          -- TODO Make import spec language extensible and use bindings rather than literals
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -603,11 +603,11 @@ initializeKernel = do
         )
       , ( "meta"
         , \dest vp stx -> do
-            Stx _ _ (_ :: Syntax, subDecl) <- mustHaveEntries stx
+            (_ :: Syntax, subDecls) <- mustHaveShape stx
             subDest <- newDeclTreePtr
             linkOneDecl dest (Meta subDest)
             inEarlierPhase $
-              expandDeclForm subDest vp =<< addRootScope subDecl
+              expandDeclForms subDest mempty vp =<< addRootScope subDecls
         )
       ]
       where
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -1496,9 +1496,9 @@ expandDeclForms dest scs vp (Syntax (Stx stxScs loc (List (d:ds)))) = do
   -- car's expandDeclForm task, who adds some Scopes to it and writes the result
   -- to carValidityPtr, which is read (and blocked on) by the forkExpandDeclForms
   -- task, who adds even more Scopes and writes the result to vp.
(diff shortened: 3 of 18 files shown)


src/Expander.hs
```diff
@@ -283,8 +285,8 @@ evalMod (Expanded em _) = snd <$> runWriterT (traverse_ evalDecl (view moduleBod
             modifyState $
               over (expanderWorld . worldTransformerEnvironments . at p . non Env.empty) $
                 Env.insert n x v
-        Meta decl -> do
-          ((), out) <- lift $ inEarlierPhase (runWriterT $ evalDecl decl)
+        Meta decls -> do
+          ((), out) <- lift $ inEarlierPhase (runWriterT $ traverse_ evalDecl decls)
```
gelisam: One design decision is whether the `meta` builtin should wrap a single declaration (which could be a `group`) or allow a list of declarations. I implemented the list approach.

src/Expander.hs
```diff
@@ -483,7 +485,7 @@ initializeKernel = do
             exprDest <- liftIO $ newSplitCorePtr
             bind b (EIncompleteDefn var x exprDest)
             schPtr <- liftIO $ newSchemePtr
-            linkDecl dest (Define x var schPtr exprDest)
+            linkOneDecl dest (Define x var schPtr exprDest)
```
gelisam: `linkDecl` still exists and still links a `DeclPtr` to a `Decl`. `linkOneDecl` is a version which binds a `ModBodyPtr` to a `Decl`. It is defined in terms of `linkDecl` and `linkModBody`.

src/Expander.hs
```diff
@@ -587,14 +589,14 @@ initializeKernel = do
...
-            linkDecl dest (Export exported)
+            linkOneDecl dest (Export exported)
             nowValidAt pdest (SpecificPhase p)
         )
       , ( "meta"
         , \sc dest pdest stx -> do
             Stx _ _ (_ :: Syntax, subDecl) <- mustHaveEntries stx
-            subDest <- liftIO newDeclPtr
-            linkDecl dest (Meta subDest)
+            subDest <- liftIO newModBodyPtr
```
gelisam: Since the inside of a `meta` block isn't really a module body, but rather a location in which a group of declarations can be defined, I'm thinking of renaming `ModBody` to `DeclGroup`.

src/Expander.hs
```diff
@@ -984,7 +986,7 @@ initializeKernel = do
       bind b val
       addToKernel name runtime b
 
-    addDeclPrimitive :: Text -> (Scope -> DeclPtr -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
+    addDeclPrimitive :: Text -> (Scope -> ModBodyPtr -> DeclValidityPtr -> Syntax -> Expand ()) -> Expand ()
```
gelisam: Currently all the primitives bind a single `Decl`, so this `DeclPtr` doesn't really need to change to a `ModBodyPtr` yet. But soon we'll have `no-decls` and `decls` primitives which do require a `ModBodyPtr`.

src/Expander.hs
```diff
@@ -1072,26 +1074,25 @@ addRootScope' stx = do
   return (addScope' stx rsc)
 
 
-forkExpandOneDecl :: DeclPtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
-forkExpandOneDecl dest sc phaseDest stx = do
-  forkExpanderTask $ ExpandDecl dest sc stx phaseDest
+forkExpandOneDecl :: ModBodyPtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
+forkExpandOneDecl dest sc declValidityPtr stx = do
+  forkExpanderTask $ ExpandModBody dest sc stx declValidityPtr
```
gelisam: I want to change the name of `forkExpandOneDecl` and `forkExpandDecls`, but I'm not sure to what.

src/Expander.hs
```diff
@@ -1072,26 +1074,25 @@ addRootScope' stx = do
   return (addScope' stx rsc)
 
 
-forkExpandOneDecl :: DeclPtr -> Scope -> DeclValidityPtr -> Syntax -> Expand ()
-forkExpandOneDecl dest sc phaseDest stx = do
-  forkExpanderTask $ ExpandDecl dest sc stx phaseDest
```
gelisam: Why are `DeclValidityPtr`s typically named `phaseDest`? My understanding is that a `DeclValidityPtr` gets filled after the name gets bound (but before the thing it gets bound to gets expanded). The expansion of later declarations blocks until all the earlier `DeclValidityPtr`s get filled-in, so we can make sure that all the relevant names are known. Names are related to phases, sure, because the same name can have different meanings in different phases, but that seems like a very minor aspect of a `DeclValidityPtr`. Perhaps that type used to be called `PhasePtr`, and that type later grew to encompass more responsibilities?

src/Expander.hs
```diff
@@ -1328,24 +1329,24 @@ addApp ctor (Syntax (Stx scs loc _)) args =
...
 problemContext (TypeDest _) = TypeCtx
 problemContext (ExprDest _ _) = ExpressionCtx
 problemContext (PatternDest _ _ _) = PatternCaseCtx
 
-requireDecl :: Syntax -> MacroDest -> Expand (Scope, DeclPtr, DeclValidityPtr)
-requireDecl _ (DeclDest dest sc ph) = return (sc, dest, ph)
-requireDecl stx other =
+requireDeclarationCtx :: Syntax -> MacroDest -> Expand (Scope, ModBodyPtr, DeclValidityPtr)
+requireDeclarationCtx _ (ModBodyDest dest sc ph) = return (sc, dest, ph)
+requireDeclarationCtx stx other =
```
gelisam: `requireModBody` sounded wrong, so I renamed this family of function to match the `MacroContext` constructor names.

src/Expander/Monad.hs
```diff
@@ -365,10 +367,20 @@ linkPattern :: PatternPtr -> ConstructorPattern -> Expand ()
...
-linkDecl :: DeclPtr -> Decl SplitTypePtr SchemePtr DeclPtr SplitCorePtr -> Expand ()
+linkModBody :: ModBodyPtr -> ModuleBodyF DeclPtr ModBodyPtr -> Expand ()
+linkModBody dest moduleBodyF =
+  modifyState $ over expanderCompletedModBody $ (<> Map.singleton dest moduleBodyF)
+
+linkDecl :: DeclPtr -> Decl SplitTypePtr SchemePtr ModBodyPtr SplitCorePtr -> Expand ()
 linkDecl dest decl =
   modifyState $ over expanderCompletedDecls $ (<> Map.singleton dest decl)
 
+linkOneDecl :: ModBodyPtr -> Decl SplitTypePtr SchemePtr ModBodyPtr SplitCorePtr -> Expand ()
```
gelisam: The combination `Decl SplitTypePtr SchemePtr ModBodyPtr SplitCorePtr` occurs often enough that it should probably be given a name. Maybe `ShallowDecl`?

src/Expander/Monad.hs
```diff
@@ -507,7 +521,7 @@ getDecl ptr =
     Just decl -> flattenDecl decl
   where
     flattenDecl ::
```
gelisam: Arguably, it is `getBody` above which is flattening a declaration tree into a flat list of declarations. Perhaps `completeDecl` would be a better name than `flattenDecl`?

src/Expander/Task.hs
```diff
@@ -36,10 +36,11 @@ data ExpanderTask
...
   | AwaitingType SplitTypePtr [AfterTypeTask]
-  | ExpandDecl DeclPtr Scope Syntax DeclValidityPtr
-    -- ^ Where to put it, the scope, the decl, where to put the phase
-  | ExpandMoreDecls ModBodyPtr Scope Syntax DeclValidityPtr
-    -- Depends on the binding of the name(s) from the decl and the phase
+  | ExpandModBody ModBodyPtr Scope Syntax DeclValidityPtr
+    -- ^ The tree node to fill, the scope, the decls, and how to notify that all
+    -- the names have been bound.
+  | ExpandDependentModBody ModBodyPtr Scope Syntax DeclValidityPtr
+    -- ^ Depends on the binding of the names in DeclValidityPtr
```
gelisam: `ExpandDecl` and `ExpandMoreDecls` were expanding the head and the tail of a list of declarations. The tail is blocked until the head's name is bound. Now, `ExpandModBody` is expending a node in the tree, while `ExpandDependentModBody` is expanding the right child of a node in the tree, by blocking until some name is bound. The intent is to wait until all the names in the left child of the node are bound, but I haven't tried to make sure that's the case yet.

The comments for `ExpandDecl` and `ExpandMoreDecls` didn't make 100% sense to me, so I hope I didn't lose anything important when I rewrote the comments to describe `ExpandModBody` and `ExpandDependentModBody` instead.

src/Pretty.hs
```diff
@@ -241,6 +241,16 @@ instance Pretty VarInfo core => Pretty VarInfo (ScopedList core) where
...
   ppBind env (CompleteDecl d) = ppBind env d
 
+instance PrettyBinder VarInfo [CompleteDecl] where
+  ppBind env [] = (text "no-decls", env)
+  ppBind env [decl] = ppBind env decl
+  ppBind env decls =
+    let (docs, envs) = unzip $ fmap (ppBind env) decls
+        doc = align $ group $ vsep docs
+        env' = mconcat envs
+    in (text "group" <> doc, env')
```
gelisam: At first I incorrectly assumed that the `env` was supposed to be threaded like a `State`, but I later realized that the input `Env` is like a `Reader` and the output `Env` is like a `Writer`. I hope I got it right this time; what would be a good way to test this implementation?

src/Expander.hs
gelisam: I renamed it to `DeclTree` instead:

* `newModBodyPtr` and `linkModBody` are now `newDeclTreePtr` and `linkDeclTree`
* `expandOneModuleBodyNode` is now `expandOneDeclTreeNode`
* `_expanderCompletedModBody` is now `_expanderCompletedDeclTrees`
* `DeclDest` became `ModBodyDest` and is now `DeclTreeDest`
* `data ModuleBodyF ... = Done | Decl ...` became `data ModuleBodyF ... = NoDecls | Decl ... | Decls ...` and is now `data DeclTreeF ...  = DeclTreeLeaf | DeclTreeAtom ... | DeclTreeBranch ...`
* `SplitModuleBody` is now `SplitDeclTree`
* `ExpandDecl` and `ExpandMoreDecls` became `ExpandModBody` and `ExpandDependentModBody` and are now `ExpandDeclTree` and `ExpandDependentDeclTree`

but

* `forkExpandOneDecl` and `forkExpandDecls` are still named the same (I still want to rename them but I don't know to what)
* `getBody` is now `getDeclGroup`, not `getDeclTree`, because it returns a list and not a tree

src/Expander.hs
gelisam: It would now be `requireDeclTree` instead of `requireModBody`, which doesn't sound as wrong, so I could revert this if you want. I like the new names a bit better though.

src/Expander/Task.hs
gelisam: (now named `ExpandDeclTree` and `ExpandDependentDeclTree`)

src/Expander.hs
david-christiansen: Seems fine to me - each can macro-express the other.

src/Expander.hs
david-christiansen: Sounds reasonable to me!

src/Expander.hs
david-christiansen: This all sounds reasonable to me.

src/Expander.hs
david-christiansen: Yep :-) This is exactly how I expected the code to turn out.

src/Expander.hs
david-christiansen: How about "forkExpandDecl" and "forkExpandDeclGroup"?

src/Expander.hs
david-christiansen: The reason I called them that when I was writing the code is that a validity pointer represents the place to which I will write information about which phase the name I've just bound is valid in. In other words, the rest of the decls block until they know which phase the previous ones are bound at.

If this was mysterious, please feel free to alpha-convert to something that makes sense to you.

src/Expander.hs
david-christiansen: This sounds like an improvement to me!

src/Expander/Monad.hs
david-christiansen: What about `SplitDecl`, by analogy to the other spit things it wraps? Should we `newtype` it to make type error messages less complicated?

src/Expander/Monad.hs
david-christiansen: My only concern about `complete` is that it can also be an adjective. What about "finalizeDecl" instead?

src/Expander/Task.hs
david-christiansen: Sounds good!

src/Pretty.hs
david-christiansen: I don't suspect that this is correct, because each decl should be in scope for the ones just after it. In other words, this would make it a foldr-like computation, rather than an fmap-like.

A way to test is to have a list with two definitions, where the second refers to the first, and then see if the var in the pretty-printed second one shows up as an internal thing with !s, or as the actual name.

At a higher level, why would the core language need pretty-printing like this? As of now, it's not attempting to mimic the surface language, but instead looks a bit more like ML, just to keep them conceptually separate.

If this is called for the body of a completed module and the body of a Meta, then I would think that a quick foldr surrounded by a vsep would be sufficient, rather than this instance.

src/Expander/Monad.hs
gelisam: Hmm, it does wrap `SplitTypePtr` and `SplitCorePtr`, but it wraps `DeclTreePtr` (formerly `ModBodyPtr`), not `SplitDeclTree`, and it wraps `SchemePtr` not `SplitScheme` (for the good reason that `SplitScheme` doesn't exist). Is this perhaps an indication that it _should_ wrap `SplitDeclTree` and `SplitScheme` instead?

david-christiansen:
No - the scheme is always created by generalizing free type variables from
an existing type, rather than being created by a macro. So I don't think
splitting it makes sense.

src/Expander/Monad.hs
gelisam: Wait a minute, we already have a specialized verb for what `flattenDecl` is doing! `flattenDecl` is replacing a bunch of `Ptr`s with their target. So `zonkDecl`?

david-christiansen:
WFM

src/Expander/Monad.hs
gelisam: renamed to zonkDecl

src/Expander.hs
gelisam: I renamed them to things like `declValidityPtr`, `valdityPtr` or `vp`, so the names no longer reference phases; but I added a comment in src/Expander/DeclScope.hs explaining the relationship between DeclValidityPtr and phases.

src/Expander.hs
gelisam: `forkExpandOneDecl` is no longer expanding a single declaration, so I don't think either `forkExpandOneDecl` nor `forkExpandDecl` is appropriate. Both functions are expanding a DeclTree... but one of the two takes an extra `DeclValidityPtr` argument. I am hoping that the difference between those two functions will become clearer to me once I make sure all the names in the left branch are bound before expanding the right branch.

gelisam: As I was hoping, that work helped me figure it out: in addition to declarations and declaration trees, there is a third concept called a declaration _form_. `(define ...)`, `(no-decls)`, and `(decls ...)` are all declaration forms, not declaration trees, and each of them expands to a tree declaration, which contains zero or more declarations. Unlike declaration trees, which are can be grouped into a tree structure, declaration forms can only be grouped into _lists_. Some positions expect a single declaration form, while other positions expect a list of zero or more declaration forms.

In addition to that, there are also _dependent_ declaration forms. A dependent declaration form is just like a regular declaration form, except its corresponding task must wait until some other `DeclValidityPtr` gets filled before the task can begin.

Before this PR, there was a coincidence between the two distinctions: the position at the head of a list always held a single declaration and was always non-dependent, while the position at the tail of a list always held a list of declarations and was always dependent on the head. With declaration trees, however, the coincidence no longer holds: the two positions inside a `decls`  both hold a single declaration, while only the second depends on the first. The tail of a list of declarations is still an example of position which holds a list of declarations and is dependent on the head, while the position inside a `meta` holds a list a declaration but is not dependent.

So I think `forkExpandOneDecl` and `forkExpandDecls` need to become _four_ different functions:

1. `forkExpandDeclForm`
2. `forkExpandDeclForms`
3. `forkExpandDependentDeclForm`
4. `forkExpandDependentDeclForms`

Furthermore, `(decls (no-decls) (define ...))` demonstrates that it is now possible for a dependent declaration form to depend on a declaration form which ends up binding zero identifiers, I think there was another coincidence between the part of `DeclValidityPtr` which indicates that a dependent task can proceed, and the part which indicates which phase each identifier is valid in. I'm not sure what to call the two parts yet.

gelisam: Hmm, another difference between `forkExpandOneDecl` and `forkExpandDecls` is that the former takes a `Scope` argument. I don't understand why the places which only accept a single declaration form inherit a `Scope` from elsewhere, while the places which accept multiple declaration forms create on `Scope` for each entry in the list. I would have expected each binding site to generate a new `Scope`, but instead it looks like all the primitives in `declPrims` receive that new `Scope` as an argument. I think that's because of yet another coincidence: each entry in the list of declaration form binds exactly one name, and so there is no need to wait to see if the head declaration form is a `define` or something else, `forkExpandDecls` already knows that it will bind a name and so it pre-emptively generates a `Scope`. This way, it can hand that `Scope` to the tail.

I guess that answers the question of how to name the part of `DeclValidityPtr` which indicates that a dependent task can proceed: a `ScopePtr`!

david-christiansen: What I like less about this is that it's not a `Scope` created as a result of user macro expansion, and this name doesn't communicate that it exists mainly for synchronization purposes.

gelisam: > this name doesn't communicate that it exists mainly for synchronization purposes.

Hmm, don't all our `Ptr`s exist for synchronization purposes? Each `Ptr` is a place where we'd normally do a call, but instead we schedule a task, and instead of doing something with the result, we schedule another task which is blocked until that result is available.

In this case, I previously thought that these DeclValidityPtrs were a special case, because the code which waits for that Ptr doesn't necessarily need to know the value it points to, it simply waits on it in order to make sure that some global map of names has been updated with all the names it might use.

But now that I have realized that we need to thread a ScopeSet (I said Scope before, but there could be multiple declarations each introducing their own Scope, so it's a ScopeSet), the picture is much more uniform: the code which waits for that ScopeSetPtr is waiting for it because it needs the ScopeSet it points to. That ScopeSet is conceptually an environment mapping all the names which are in scope to (a pointer to) their values, except we've implemented that via a more complicated setup whereby you can get access to that environment from the global environment by tagging your name lookups with a ScopeSet. So obviously we shouldn't fill a ScopeSetPtr until after we've updated that global environment.

In conclusion, I think `ScopeSetPtr` is a pretty descriptive name to indicate that we're waiting until all the names in the corresponding environment are bound!

gelisam: I have now split DeclValidityPtr into PhaseValidityPtr and ScopeSetPtr, but I don't think I'll push my changes, as I no longer think they need to be split.

My understanding is that before this PR, as we traverse a list of declarations, we allocate one Scope per declaration and then we add that Scope to the Syntax of the remaining declarations at the right phase (or possibly all phases in the case of `import`). These two pieces of information, the Scope and the phase, are transmitted via two very different paths. The Scope is created first, in `forkExpandDecls`, and is passed as a parameter of both the `ExpandDecl` and the `ExpandMoreDecls` tasks. This allows the `ExpandDecl` task to bind its identifier to the same Scope which `ExpandMoreDecls` will add to the Syntax of the remaining declarations. For the phase, `forkExpandDecls` instead creates a new `DeclValidityPtr` which the `ExpandDecl` task fills and which the `ExpandMoreDecls` task waits upon.

A different design would be to transfer both the Scope and the phase via the same path: `forkExpandDecls` would only create a new `DeclValidityPtr`, not a new Scope, and then the `ExpandDecl` task would fill that `DeclValidityPtr` with _both_ a Scope and a phase.

That design generalizes better to the declaration-tree case, because a declaration form can now bind an arbitrary number of identifiers and thus needs to be able to allocate an arbitrary number of Scopes, so the `ExpandDecl` cannot simply receive a single Scope as input. Instead, it must now be able to fill the `DeclValidityPtr` with zero or more Scopes, each for each of them, it must specify whether the scope is valid in a particular phase or in all phases. Put another way, the thing which `DeclValidityPtr` points to is map from AllPhases to a set of Scopes, and a map from each Phase to a set of Scopes. That is, it points to a `ScopeSet`!

Now that I no longer want to split `DeclValidityPtr` into two parts, I don't really need to rename it, but I still think `ScopeSetPtr` would be a better name than `DeclValidityPtr` if it points to a ScopeSet.

gelisam: I now realize that I don't really understand the "it's not a Scope created as a result of user macro expansion" part of your objection to the name, can you explain why the name ScopePtr implies a relation with user macros?

src/Expander/Monad.hs
gelisam: I think I'll just leave it alone.

src/Expander/DeclScope.hs
```diff
@@ -2,6 +2,13 @@ module Expander.DeclScope where
 
 import Data.Unique
 
+-- | A 'DeclValidityPtr' gets filled with a 'ScopeSet' representing an
+-- environment. Each 'Scope' in this 'ScopeSet' corresponds to an identifier
+-- which is bound in that environment, and the 'ScopeSet' indicates the phase
+-- or phases in which each 'Scope' is valid.
+--
+-- Note that 'DeclValidityPtr' gets filled once we know which _names_ get
+-- bound, the values to which they are bound may not be fully-expanded yet.
```
gelisam: I changed this comment again, to reflect the fact that `DeclValidityPtr`s now point to `ScopeSet`s instead of `PhaseSpec`s.

src/Expander.hs
```diff
@@ -492,10 +494,10 @@ initializeKernel = do
             modifyState $ over (expanderDefTypes . at ph . non Env.empty) $
               Env.insert var x schPtr
             forkGeneralizeType exprDest t schPtr
-            nowValidAt pdest (SpecificPhase p)
+            linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
```
gelisam: Instead of filling the `DeclValididtyPtr` with a `PhaseSpec` (which no longer exists), we now fill it with an equivalent `ScopeSet`. For `SpecificPhase p`, the equivalent `ScopeSet` is one which assigns the scope to phase `p`.

src/Expander.hs
```diff
@@ -531,12 +533,12 @@ initializeKernel = do
               Just info
 
 
-            forkEstablishConstructors pdest d ctors
+            forkEstablishConstructors sc vp d ctors
```
gelisam: I now have to forward the `Scope` to `forkEstablishConstructors` because that's where the `DeclValidityPtr` gets filled.

src/Expander.hs
```diff
@@ -551,52 +553,52 @@ initializeKernel = do
...
+        , \sc dest vp stx -> do
             Stx _ _ (_, protoSpec) <- mustBeCons stx
             exported <- exportSpec stx protoSpec
             p <- currentPhase
             es <- getExports exported
             modifyState $ over expanderModuleExports $ (<> es)
-            linkDecl dest (Export exported)
-            nowValidAt pdest (SpecificPhase p)
+            linkOneDecl dest (Export exported)
+            linkDeclValidity vp (ScopeSet.singleScopeAtPhase sc p)
```
gelisam: The `Scope` was previously unused, presumably because `export` doesn't bind anything. So I think that the `nowValidAt pdest (SpecificPhase p)` was filling its `DeclValidityPtr` with a dummy phase, but that I could now fill it with an empty `ScopeSet` instead?

src/Expander.hs
```diff
@@ -1072,26 +1074,25 @@ addRootScope' stx = do
...
+  forkExpanderTask $ ExpandDeclTree dest sc stx declValidityPtr
 
-forkExpandDecls :: ModBodyPtr -> Syntax -> Expand ()
+forkExpandDecls :: DeclTreePtr -> Syntax -> Expand ()
 forkExpandDecls dest (Syntax (Stx _ _ (List []))) =
-  modifyState $ over expanderCompletedModBody (<> Map.singleton dest Done)
+  linkDeclTree dest DeclTreeLeaf
 forkExpandDecls dest (Syntax (Stx scs loc (List (d:ds)))) = do
   -- Create a scope for this new declaration
   sc <- freshScope $ T.pack $ "For declaration at " ++ shortShow (stxLoc d)
```
gelisam: I'm doing this one step at a time, and for now I'm still generating the scope inside `forkExpandDecls`. Next, I want to move this line to each `declPrim`.

src/Expander.hs
```diff
@@ -1191,17 +1192,15 @@ runTask (tid, localData, task) = withLocal localData $ do
...
+      readyYet <- view (expanderDeclValidities . at waitingOn) <$> getState
       case readyYet of
         Nothing ->
           stillStuck tid task
-        Just (SpecificPhase p) ->
-          forkExpandDecls dest (addScope p stx sc)
-        Just AllPhases ->
-          forkExpandDecls dest (addScope' stx sc)
+        Just scopeSet ->
+          forkExpandDecls dest (addScopes stx scopeSet)
```
gelisam: We're now adding all the scopes in the `ScopeSet` to `stx`, not just `sc`. This task is no longer receiving that `sc`, because we now receive the scope from the `DeclValidityPtr`.

src/Expander/Monad.hs
```diff
@@ -668,8 +688,3 @@ datatypeInfo datatype = do
     Nothing ->
       throwError $ InternalError $ "Unknown datatype " ++ show datatype
     Just info -> pure info
-
-
-nowValidAt :: DeclValidityPtr -> PhaseSpec -> Expand ()
-nowValidAt ptr p =
-  modifyState $ over expanderDeclPhases $ Map.insert ptr p
```
gelisam: replaced with `linkDeclValidity`, next to the other `link***` functions

src/Expander/Task.hs
```diff
@@ -36,18 +36,18 @@ data ExpanderTask
...
     -- ^ Waiting on var, binding, and definiens, destination, syntax to expand
   | AwaitingType SplitTypePtr [AfterTypeTask]
-  | ExpandDecl DeclPtr Scope Syntax DeclValidityPtr
-    -- ^ Where to put it, the scope, the decl, where to put the phase
-  | ExpandMoreDecls ModBodyPtr Scope Syntax DeclValidityPtr
-    -- Depends on the binding of the name(s) from the decl and the phase
+  | ExpandDeclTree DeclTreePtr Scope Syntax DeclValidityPtr
+    -- ^ The tree node to fill, the scope, the decls, and the output environment.
+  | ExpandDependentDeclTree DeclTreePtr Syntax DeclValidityPtr
+    -- ^ The tree node to fill, the decls, and the input environment.
```
gelisam: I changed the comments again to describe the `DeclTreePtr`s as pointing to environments. Maybe `EnvPtr` would be a more acceptable name than `ScopeSetPtr`?

src/Phase.hs
```diff
@@ -14,9 +14,6 @@ makePrisms ''Phase
 instance ShortShow Phase where
   shortShow (Phase i) = "p" ++ show i
 
-data PhaseSpec = AllPhases | SpecificPhase !Phase
-makePrisms ''PhaseSpec
-
```
gelisam: `PhaseSpec` no longer exists; a `DeclValidityPtr` now needs to point to a map from `PhaseSpec` to `Scope`, and that's already what a `ScopeSet` is.

src/Expander.hs
gelisam: I changed it to return `mempty`. Later, I changed the `declPrim` contract from "fill the DeclValidityPtr with a ScopeSet listing the newly-bound Scopes" to "receive a ScopeSet, add the newly-bound Scopes to it, then fill the DeclValidityPtr with the result", so now this instead returns the input ScopeSet unmodified.

src/Expander.hs
gelisam: I have now removed this `freshScope` call and replaced it with multiple `freshScope` calls elsewhere, one per `primDecl`.

src/Expander.hs
```diff
@@ -115,12 +115,13 @@ expandModule thisMod src = do
   local (set (expanderLocal . expanderModuleName) thisMod) do
     lang <- mustBeModName (view moduleLanguage src)
     initializeLanguage lang
-    modBodyDest <- liftIO $ newModBodyPtr
-    modifyState $ set expanderModuleTop $ Just modBodyDest
+    declTreeDest <- liftIO $ newDeclTreePtr
+    vp <- newDeclValidityPtr
```
gelisam: This DeclValidityPtr gets filled, but isn't read anywhere. An alternative would be to use `Maybe DeclValidityPtr` everywhere, and to only fill it when it's a `Just`.

src/Expander.hs
```diff
@@ -115,12 +115,13 @@ expandModule thisMod src = do
...
+    modifyState $ set expanderModuleTop $ Just declTreeDest
     decls <- addModuleScope (view moduleContents src)
-    forkExpandDecls modBodyDest decls
+    expandDeclForms declTreeDest mempty vp decls
```
gelisam: Here and in a few other places, I now call `expand***` instead of `forkExpand***`. An alternative would be to duplicate the `ExpandDeclForm` and `ExpandDeclForms` into two more tasks, which take a `ScopeSet` instead of an input `DeclValidityPtr`. Yet another alternative would be to create a new `DeclValidityPtr` and to immediately fill it with a `ScopeSet`, in this case `mempty`.

src/Expander.hs
```diff
@@ -115,12 +115,13 @@ expandModule thisMod src = do
...
     expandTasks
```
gelisam: Whenever an `import` is expanded, the tests fail here, at `expandTasks`, with `No progress was possible`. The error message lists a single stuck task: the `ExpandDeclForms` which is waiting for the `import` to complete. I do expect the task list to still contain that task while the `import` is expanded, so I don't understand why `expandTasks` complains. Is that function trying to expand _all_ the tasks? At least one task? As many as possible?

src/Expander.hs
```diff
@@ -456,34 +459,36 @@ initializeKernel = do
...
+                vp <- newDeclValidityPtr
+                expandDeclForms bodyPtr mempty vp body
 
                 buildExports exports
                 pure ()
         )
       ]
 
-    declPrims :: [(Text, Scope -> DeclPtr -> DeclValidityPtr -> Syntax -> Expand ())]
+    declPrims :: [(Text, DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ())]
```
gelisam: I took the opportunity to reorder a few type signatures to be more consistent: the destination is always first, and the syntax is always last.

src/Expander.hs
```diff
@@ -456,34 +459,36 @@ initializeKernel = do
...
     declPrims =
       [ ( "define"
-        , \ sc dest pdest stx -> do
+        , \ dest scs vp stx -> do
             p <- currentPhase
             Stx _ _ (_, varStx, expr) <- mustHaveEntries stx
+            sc <- freshScope $ T.pack $ "For definition at " ++ shortShow (stxLoc stx)
```
gelisam: Instead of receiving a `Scope` which can be used to bind a single new identifier, each `declPrim` now creates as many `Scope`s as they need, and the fills the `DeclValidityPtr` with a `ScopeSet` containing all those `Scope`s (with the proper phase for each). This will make it possible for `no-decls` to create 0 and for `decls` to create many.

Since each `Scope` created in that tree of declarations needs to be visible from all the later scopes, we need to accumulate more an more `Scope`s as we traverse that tree. To do that, each `declPrim` is now receiving an input `ScopeSet` as well, consisting of all the `Scope`s which have been created so far, and then it adds its own `Scope`s to this input `ScopeSet` and puts the resulting output `ScopeSet` in the `DeclValidityPtr`. I think I might be able to make this simpler.

src/Expander.hs
```diff
@@ -551,52 +558,53 @@ initializeKernel = do
...
       , ( "example"
-        , \ sc dest pdest stx -> do
+        , \ dest scs vp stx -> do
             p <- currentPhase
             Stx _ _ (_ :: Syntax, expr) <- mustHaveEntries stx
             exprDest <- liftIO $ newSplitCorePtr
             sch <- liftIO newSchemePtr
-            linkDecl dest (Example (view (unSyntax . stxSrcLoc) stx) sch exprDest)
+            linkOneDecl dest (Example (view (unSyntax . stxSrcLoc) stx) sch exprDest)
+            sc <- freshScope $ T.pack $ "For example at " ++ shortShow (stxLoc stx)
```
gelisam: I'm only creating a `Scope` because the code was previously using its input `Scope`. But why does an `example` need a `Scope`? It doesn't bind any new identifier.

src/Expander.hs
```diff
@@ -1072,30 +1083,9 @@ addRootScope' stx = do
...
-    (<> Map.singleton dest (Decl declDest restDest))
-  forkExpanderTask $
-    ExpandMoreDecls restDest sc
-      (Syntax (Stx scs loc (List ds)))
-      declPhase
-  forkExpandOneDecl declDest sc declPhase =<< addRootScope d
-  where
-    stxLoc (Syntax (Stx _ srcloc _)) = srcloc
-forkExpandDecls _dest _other =
-  error "TODO real error message - malformed module body"
```
gelisam: `forkExpandDecls` is now `expandDeclForm`, and has been moved next to `expandDeclForm` (formerly `expandOneDeclaration`)

src/Expander.hs
```diff
@@ -1258,8 +1251,7 @@ runTask (tid, localData, task) = withLocal localData $ do
         Just ready -> do
           for_ ready \(cn, ctor, ts) ->
             addConstructor cn dt ctor ts
-          p <- currentPhase
-          nowValidAt pdest (SpecificPhase p)
+          linkDeclValidity vp scs
```
gelisam: the current phase is now added to the input `ScopeSet`, when `forkEstablishConstructors` is called.

src/Expander.hs
```diff
@@ -1473,8 +1465,24 @@ expandOneForm prob stx
...
+  linkDeclTree dest DeclTreeLeaf
+  linkDeclValidity vp scs
+expandDeclForms dest scs vp (Syntax (Stx stxScs loc (List (d:ds)))) = do
+  carDest <- liftIO $ newDeclTreePtr
+  cdrDest <- liftIO $ newDeclTreePtr
+  carValidityPtr <- newDeclValidityPtr
+  linkDeclTree dest (DeclTreeBranch carDest cdrDest)
+  forkExpandDeclForms cdrDest carValidityPtr vp
+    (Syntax (Stx stxScs loc (List ds)))
+  expandDeclForm carDest scs carValidityPtr =<< addRootScope d
```
gelisam: This is where the `ScopeSet` is threaded through: from the input `scs`, to the `car`'s `expandDeclForm` task, who adds some `Scope`s to it and writes the result to `carValidityPtr`, which is read (and blocked on) by the `forkExpandDeclForms` task, who adds even more `Scope`s and writes the result to `vp`.

src/Expander/Monad.hs
```diff
@@ -184,7 +187,7 @@ data EValue
   = EPrimMacro (Ty -> SplitCorePtr -> Syntax -> Expand ()) -- ^ For special forms
   | EPrimTypeMacro (SplitTypePtr -> Syntax -> Expand ()) -- ^ For type-level special forms
   | EPrimModuleMacro (Syntax -> Expand ())
-  | EPrimDeclMacro (Scope -> DeclPtr -> DeclValidityPtr -> Syntax -> Expand ())
+  | EPrimDeclMacro (DeclTreePtr -> ScopeSet -> DeclValidityPtr -> Syntax -> Expand ())
```
gelisam: Now takes a `ScopeSet` instead of a `Scope`.

src/Expander/Monad.hs
```diff
@@ -232,16 +235,16 @@ data ExpanderState = ExpanderState
...
+  , _expanderCompletedDecls :: !(Map.Map DeclPtr (Decl SplitTypePtr SchemePtr DeclTreePtr SplitCorePtr))
+  , _expanderModuleTop :: !(Maybe DeclTreePtr)
   , _expanderModuleImports :: !Imports
   , _expanderModuleExports :: !Exports
   , _expanderPhaseRoots :: !(Map Phase Scope)
   , _expanderModuleRoots :: !(Map ModuleName Scope)
   , _expanderKernelBindings :: !BindingTable
   , _expanderKernelExports :: !Exports
-  , _expanderDeclPhases :: !(Map DeclValidityPtr PhaseSpec)
+  , _expanderDeclValidities :: !(Map DeclValidityPtr ScopeSet)
```
gelisam: Now maps to a `ScopeSet`.

src/Expander/Task.hs
```diff
@@ -23,7 +23,8 @@ import Value
 data MacroDest
   = ExprDest Ty SplitCorePtr
   | TypeDest SplitTypePtr
-  | DeclDest DeclPtr Scope DeclValidityPtr
+  | DeclTreeDest DeclTreePtr ScopeSet DeclValidityPtr
+    -- ^ tree node to fill, input environment, output environment
```
gelisam: Added a comment to further drive home that the `ScopeSet` / first `DeclValidityPtr` is always the input environment while the second `DeclValidityPtr` is the output environment.

src/Expander/Task.hs
```diff
@@ -36,18 +37,18 @@ data ExpanderTask
...
     -- ^ Waiting on var, binding, and definiens, destination, syntax to expand
   | AwaitingType SplitTypePtr [AfterTypeTask]
-  | ExpandDecl DeclPtr Scope Syntax DeclValidityPtr
-    -- ^ Where to put it, the scope, the decl, where to put the phase
-  | ExpandMoreDecls ModBodyPtr Scope Syntax DeclValidityPtr
-    -- Depends on the binding of the name(s) from the decl and the phase
+  | ExpandDeclForm DeclTreePtr DeclValidityPtr DeclValidityPtr Syntax
+    -- ^ The tree node to fill, the input environment, the output environment, and the declaration form.
+  | ExpandDeclForms DeclTreePtr DeclValidityPtr DeclValidityPtr Syntax
+    -- ^ The tree node to fill, the input environment, the output environment, and the declaration forms.
```
gelisam: There used to be two types of tasks, there would now need to be 2*2=4, depending on whether the input syntax is a single declaration form or a list of declaration forms, and on whether the input environment is a `DeclValidityPtr` we're waiting for or a `ScopeSet` we already have. For simplicity, I now call `expandDeclForm[s]` directly instead of forking a task when I already have a `ScopeSet`.

src/Expander/Task.hs
david-christiansen: Good! By "environment" you mean "scopes to add to these decls" vs "scopes to add to subsequent decls", right?

src/Expander/Monad.hs
david-christiansen: This is a nice improvement.

src/Expander/Monad.hs
david-christiansen: I think using ScopeSets here was a very good decision, and it seems to be making the whole thing work out nicely.

src/Expander.hs
david-christiansen: What about adding this text as a comment in the source file?

src/Expander.hs
david-christiansen: It doesn't. That was an infelicity in the old approach that you've resolved with this generalization.

src/Expander.hs
david-christiansen: Excellent!

src/Expander.hs
david-christiansen: That sounds like the import is not filling its validity pointer with something.

That function should be running as many tasks as possible.

src/Expander.hs
david-christiansen: If the whole thing works, then I see no reason to round-trip through the task system.

src/Expander/Task.hs
david-christiansen: We already have something called an `Env`, and it's not this.

More generally, I think it's important to keep the concept of environment (mapping core names to values), typing contexts (mapping core names to type schemes), and scope sets (which aren't really even conceptually mappings - the internal `Map` is used to make them have different natures at different times, but they're intended to be a bit more set-like).

What about `DeclOutputScopesPtr`?

src/Expander/Task.hs
gelisam: To subsequent decls. The input `ScopeSet` is the set of `Scopes` which the previous decls have created (they have already been added to the `Syntax` in `forkExpandDecls`), and the output `DeclValidityPtr` must be filled with a `ScopeSet` which includes both the input `ScopeSet` and the `Scope`s created by the code which fills this `DeclTreePtr`, so that subsequent decls can see the identifiers bound by that code.

src/Expander.hs
gelisam: good idea, added!

src/Expander.hs
gelisam: excellent, removed the `freshScope` call.

src/Expander.hs
gelisam: > That sounds like the import is not filling its validity pointer with something.

`import` doesn't get the chance to fill its validity pointer, because `expandTasks` fails [here](https://github.com/gelisam/klister/pull/37/files#diff-a63cde2fb12e931100bbb12fc30e04eeR582), while we're expanding `import`.

> That function should be running as many tasks as possible.

Then should [this](https://github.com/gelisam/klister/blob/master/src/Expander.hs#L1118) be `return ()` instead of `throwError`? It looks to me like `expandTask` is trying to run as many tasks as possible until there are none left, and if it ever reaches a point where all the remaining tasks are stuck, it fails.

gelisam: changing it to `return ()` allows the tests which use `import` to pass, but now `wait for a signal which is never coming` and `send and receive different signals` are failing, because they are expecting a `NoProgress` failure which is no longer thrown.

david-christiansen: > It looks to me like expandTask is trying to run as many tasks as possible until there are none left, and if it ever reaches a point where all the remaining tasks are stuck, it fails.

This is exactly what it should do.

Your link labeled "here" doesn't work for me - it just takes me to the overall diff for this PR.

What might be going wrong is that you might not be creating a fresh task ID when necessary. Tasks in the queue don't just get to be structurally equal - they need a notion of identity. This is so that the check for progress can tell the difference between a task that couldn't make progress and a task that made progress (updating the global state somewhere) but left behind something that looks identical to itself.

The `return ()` is definitely not right - we want to detect deadlock, not just skip things arbitrarily in case of deadlock!

src/Expander/DeclScope.hs
```diff
@@ -2,6 +2,13 @@ module Expander.DeclScope where
 
 import Data.Unique
 
+-- | A 'DeclValidityPtr' gets filled with a 'ScopeSet' representing an
+-- environment. Each 'Scope' in this 'ScopeSet' corresponds to an identifier
```
david-christiansen: I don't think `ScopeSets` represent environments *per se*. They instead represent scopes in the abstract sense, rather than in the sense of the specific datatype we have here - remember, they are generalizations of de Bruijn indices that account for multiple dimensions of binding induced by macro expansion. The difference is that environments are fundamentally mappings from names to values, while these are just about which names can be referred to by a given identifier.

src/Expander.hs
gelisam: > This is exactly what it should do.

Hmm, I'm confused. If it should run as many tasks as possible, then I would expect it to run all the tasks it can, continuing if new tasks are forked, and stopping if all the tasks are stuck. Failing if all the available tasks are stuck only seems to make sense if the goal of `expandTasks` is to complete all the tasks which have ever been forked, but I don't understand why we would want that.

> Your link labeled "here" doesn't work for me - it just takes me to the overall diff for this PR.

GitHub's UX is quite poor here; it "helpfully" hides large diffs such as `src/Expander.hs`'s, and so you have to click `load diff` on that file, and then you have to scroll up because the selected line is hidden by the top bar. [Here is another link](https://github.com/gelisam/klister/blob/75362ae938116f06a9ae21c065ba223f736834f4/src/Expander.hs#L576) which doesn't suffer from those problems.

> What might be going wrong is that you might not be creating a fresh task ID when necessary.

I am not creating a fresh task ID, I am using `stillStuck tid task` to re-enqueue the same task when it is stuck, same as the `ExpandMoreDecls` task was doing.

> The `return ()` is definitely not right - we want to detect deadlock, not just skip things arbitrarily in case of deadlock!

I don't think we're skipping _arbitrary_ tasks: we are skipping the tasks which cannot currently make progress, because they are blocked on something else. If this was the top level call into the task system, I agree that we should make sure that all the tasks are expanded and that no leftover tasks remain, as that would indicate that the tasks got permanently stuck and that some of the file is not fully type-checked nor fully expanded. In this case, however, we are importing another file. It might make sense to make sure that all the tasks spawned while expanding the imported file are complete, but I don't think we are equipped to distinguish those tasks from the tasks which were spawned in the file which calls `import`. And I don't think it makes sense to require all the tasks in both files to complete before the `import` task can complete, as there are some already-spawned tasks which are blocked waiting for that `import` to complete, and that's what's causing the failures.

gelisam: I managed to get the tests to pass by distinguishing the places where I do expect all tasks to be completed (after every repl command, when running a file from the command-line, and after each test) from the places where I expect some of the tasks to be left over (when importing a file or initializing the #lang). The former now call `fullyExpandExpr` and `fullyVisit` (there is also an unused `fullyExpandModule`) while the latter call `expandModule` and `visit` (there is also an `expandExpr` which is only used in `fullyExpandExpr`).

Do you still think this is the wrong approach?

david-christiansen: Unless I've misunderstood you, I do still think this is the wrong approach, for one principled reason and for a heuristic reason.

The principled reason is that modules should be *modular* - they should make sense on their own. At a minimum, a module system should support things like independent type checking/compilation and a separation of interface and implementation. But allowing tasks to remain unsolved after expanding a module means that a task could be left like "Await signal 5, then go do something". Whether or not that thing is ever done could then depend on whether some other module sent the signal. Similar issues could arise from other forms of dependencies. And the `#lang` is just specifying a module from which the default bindings are drawn. So really, when expanding a module, the task queue should start out containing only the top level task, and it should end by requiring the queue to be empty, achieving modularity. Don't recursive calls to the module expander save their task queue and restore it once the recursive expansion is complete?

The heuristic reason for this is that it seems to have worked before, and I don't know why the changes you're making to the system should introduce this behavior, and that makes me suspicious that something else has gone wrong.

src/Expander/Task.hs
david-christiansen: ScopeSets don't associate identifiers with definitions, they associate identifiers with their bindings, which might be definitions, but that also might be e.g. a local variable in a lambda.

src/Expander.hs
david-christiansen: Only that when we have used `Ptr`s elsewhere, it's been to make an incrementally-built structure that comes out of the expander. But that's not really inherent to pointerness. I retract my objection :-)

src/Expander.hs
gelisam: > Don't recursive calls to the module expander save their task queue and restore it once the recursive expansion is complete?

Ah, that's the piece I was missing! Yes, if the tasks which I claim should still be stuck after the `import` task completes are removed from the task queue before we call `expandTasks`, then I agree that we should expect the task queue to be empty afterwards, we shouldn't expect it to still contain those tasks we removed.

david-christiansen: What about an operator like this sketch, then?
```
completely :: Expand a -> Expand a
completely act =
  do oldTasks <- getTasks
     setTasks []
     act
     remaining <- getTasks
     if remaining != [] then throwError ... else pure ()
     setTasks oldTasks
```
Then module expansion, recursive or otherwise, can be ``completely (expandModule ...)`` etc. Or `expandModule` can have the `completely` in its definition, because it kinda doesn't make sense to expand a module incompletely here.

gelisam: That's a much better API, thanks.

gelisam: I got rid of the `fully***` functions and am now using `completely` instead. I did not yet change `expandModule` to always use `completely`, but now that the semantics is to save and restore the task queue rather than to fail if the queue is not empty at the end, I think it makes sense to use `completely` a lot more often.

It does not look like we were previously saving and restoring the task queue, I think that's the root cause of the failures. I don't yet understand how the tests were previously passing without this.

gelisam: Okay, so the reason the tests are passing on `master` even though we don't save and restore the task queue is by sheer luck. The order of those two `fork***` calls in `forkExpandDecls` shouldn't matter, and yet the tests fail if I swap them:

```haskell
  forkExpanderTask $
    ExpandMoreDecls restDest sc
      (Syntax (Stx scs loc (List ds)))
      declPhase
  forkExpandOneDecl declDest sc declPhase =<< addRootScope d
```

Ironically, I think those two lines are in that unintuitive order specifically to demonstrate that we don't need to expand the head of the decls before the tail of the decls, we can put those two lines in any order and the task-dependency logic will nevertheless ensure that the tail doesn't start too early. However, since `forkExpanderTask` uses `(task:)` rather than `(++ [task])`, forking the tail task before the head task is precisely the order which leads the expander to run the head task before the tail task.

The next bit of luck is `expandTasks` doesn't pop the tasks from the queue one by one, it pops _all_ the tasks and then runs each of those tasks one at a time. This means that as the very first task gets executed, the global task queue is empty, even though `expandTasks` still has a few other tasks to expand. Since the head `import` task is ran first, the queue starts empty, so its recursive call to `expandTasks` successfully ends with an empty queue as well. When I switch the order, the tail task runs first, it re-enqueues itself, the queue starts full, and so the recursive call to `expandTasks` ends with a full queue and errors out.

The final bit of luck is that a sequence of `import` declarations happens to interleave in such a way that the `import` is always the first task in the queue: first the queue contains the head and the tail decl tasks, we successfully run the head import decl task, and then we successfully run the tail decl task, which pushes the next head and tail decl tasks onto the queue. If there are several `import`s in a row, that next head decl task is another import.

As a conclusion, I have now reached the "How did that ever work?" [stage of debugging](http://plasmasturm.org/log/6debug/) for this particular bug ;)

src/Expander/DeclScope.hs
gelisam: I changed this comment yet again, because I had used the word "environment" incorrectly.

src/Expander.hs
gelisam: I had switched from `SpecificPhase p` to `ScopeSet.singleScopeAtPhase sc p` and then to `addPhase p scs sc` (or equivalently `scs <> ScopeSet.singleScopeAtPhase sc p`), and now I'm back to `ScopeSet.singleScopeAtPhase sc p` because I found a way to accumulate multiple scopes without forcing each declaration form to thread through a `ScopeSet`. Now each declaration form only needs to fill their `DeclValidityPtr` with the scopes they introduce.

src/Expander.hs
gelisam: Now returning `mempty` again.

src/Expander/Task.hs
gelisam: I have now updated the comments to talk about scopes, not environments.

src/Expander.hs
gelisam: I did manage to make this simpler: instead of receiving a `ScopeSet` and adding their own scopes to it, most declaration forms do not receive a `ScopeSet`, and so they only need to fill their `DeclValidityPtr` with the scopes they introduce themselves. The exception to that is the `ExpandFormDecls` task, which now receives both a `ScopeSet` _and_ an input `DeclValidityPtr` _and_ an output `DeclValidityPtr`. The comments I wrote there and on `expandDeclForm[s]` are hopefully clear enough about how this works.

src/Expander/Task.hs
gelisam: I have changed the comment to avoid misusing the word "environment". The comment is much longer now; maybe too long?

src/Expander/DeclScope.hs
gelisam: I have now changed the comment to avoid misusing the word "environment".

src/Expander.hs
```diff
@@ -1191,17 +1189,19 @@ runTask (tid, localData, task) = withLocal localData $ do
...
-      expandOneDeclaration sc dest stx ph
-    ExpandMoreDecls dest sc stx waitingOn -> do
-      readyYet <- view (expanderDeclPhases . at waitingOn) <$> getState
+    ExpandDeclForms dest earlierScopeSet waitingOn vp stx -> do
+      -- The scopes in waitingOn (from the preceding declaration) haven't yet
+      -- been added to stx, while those in earlierScopeSet (from the
+      -- declarations before that) have. The scopes from both, plus the scopes
+      -- introduced by the declaration forms in stx, must all eventually be
+      -- placed in vp so that the code which isn't part of this declaration
+      -- group can have access to those declarations.
```
gelisam, line 1198: Here is one of the comments which explains how I'm accumulating the scopes now.

src/Expander.hs
```diff
@@ -1473,8 +1472,37 @@ expandOneForm prob stx
...
+  linkDeclTree dest (DeclTreeBranch headDest tailDest)
+
+  -- The head declaration will fill the headValidityPtr with the scopes it
+  -- introduces, and then the ExpandDeclForms task will fill vp with the
+  -- combination of scs, headValidityPtr's scopes, and the scopes indroduced by
+  -- the tail's declarations.
+  expandDeclForm headDest headValidityPtr =<< addRootScope d
+  forkExpanderTask $ ExpandDeclForms tailDest scs headValidityPtr vp (Syntax (Stx stxScs loc (List ds)))
+expandDeclForms _dest _scs _vp _stx =
+  error "TODO real error message - malformed module body"
```
gelisam, line 1505: In the previous two functions are more comments explaining how I'm accumulating the scopes now.

src/Expander/Task.hs
```diff
@@ -36,18 +37,19 @@ data ExpanderTask
...
   | AwaitingType SplitTypePtr [AfterTypeTask]
-  | ExpandDecl DeclPtr Scope Syntax DeclValidityPtr
-    -- ^ Where to put it, the scope, the decl, where to put the phase
-  | ExpandMoreDecls ModBodyPtr Scope Syntax DeclValidityPtr
-    -- Depends on the binding of the name(s) from the decl and the phase
+  | ExpandDeclForms DeclTreePtr ScopeSet DeclValidityPtr DeclValidityPtr Syntax
+    -- ^ The produced declaration tree, some already-introduced scopes which
+    -- the syntax can already see, some to-be-introduced scopes which the will
+    -- see, a destination for all the introduced scopes, including those by the
+    -- Syntax's remaining declaration forms.
```
gelisam, line 44: Is this comment too long?

src/Expander/Task.hs
```diff
@@ -36,18 +37,19 @@ data ExpanderTask
   | AwaitingDefn Var Ident Binding SplitCorePtr Ty SplitCorePtr Syntax
     -- ^ Waiting on var, binding, and definiens, destination, syntax to expand
   | AwaitingType SplitTypePtr [AfterTypeTask]
-  | ExpandDecl DeclPtr Scope Syntax DeclValidityPtr
-    -- ^ Where to put it, the scope, the decl, where to put the phase
-  | ExpandMoreDecls ModBodyPtr Scope Syntax DeclValidityPtr
-    -- Depends on the binding of the name(s) from the decl and the phase
```
gelisam, line 42: Now that I'm calling `expandDeclForm` instead of `forkExpandDeclForm`, I no longer need the `ExpandDeclForm` task.

src/Expander.hs
gelisam: `expandModule` and `initializeLanguage` now both use `completely` (indirectly, via `visit`).

src/Expander.hs
```diff
@@ -551,52 +563,56 @@ initializeKernel = do
...
+            subDest <- newDeclTreePtr
+            linkOneDecl dest (Meta subDest)
             inEarlierPhase $
-              forkExpandOneDecl subDest sc pdest =<< addRootScope subDecl
+              expandDeclForms subDest mempty vp =<< addRootScope subDecls
+        )
+      , ( "group"
+        , \dest vp stx -> do
+            (_ :: Syntax, decls) <- mustHaveShape stx
+            expandDeclForms dest mempty vp decls
```
gelisam, line 615: Turns out it was easier to implement `group` directly than to implement it in terms of `no-decls` and `decls`.

david-christiansen: Nice!

src/Expander/Task.hs
david-christiansen: It might be a sign that there's enough arguments to warrant a record here, but I think it's fine. Might be nicer as a bulleted list.

src/Expander.hs
david-christiansen: That's great!  What is `vp` an abbreviation for?

src/Expander.hs
david-christiansen: They are indeed clear enough. Thanks!

src/Expander.hs
gelisam: "validity pointer". Maybe "dvp" would be better?

src/Expander.hs
david-christiansen: Wow! This is some good detective work. I'm really glad you've gotten it cleaned up. Sorry to have built it so strangely to start with.

src/Expander.hs
david-christiansen: What about `outScopes` or `scopeOutDest` or something else that at includes some recognizable words?

gelisam: ok, I've renamed `DeclValidityPtr` to `DeclOutputScopesPtr` and `vp` to `outScopesDest`.

examples/double-define.kl
```diff
@@ -0,0 +1,12 @@
+#lang kernel
+
+(import "define-syntax-rule.kl")
+
+(define-syntax-rule (double-define x y value)
+  (group
```
david-christiansen, line 6: This line right here is a dream come true. Thank you!

src/Pretty.hs
gelisam: Indeed, this implementation is quite wrong. With a few cosmetic fixes, it pretty-prints `group.kl` like this:

```
module "\"/Users/gelisam/working/haskell/klister/examples/group.kl\""
    import #[group.kl:2.16-2.22]<"kernel"> ⇑ 1
    define a : Signal := Signal 1
    meta
        define a : Signal := Signal 10
    define b : Signal := a
    define c : Signal := b
    meta
        define b : Signal := a
        define c : Signal := !!(Var 117)!!
    example@group.kl:14.1-15.1 a : Signal
    example@group.kl:15.1-16.1 b : Signal
    example@group.kl:16.1-17.1 c : Signal
    meta
        example@group.kl:18.3-19.3 a : Signal
        example@group.kl:19.3-20.3 b : Signal
        example@group.kl:20.3-20.14 c : Signal
```

That is, it contains the dreaded exclamation marks. Replacing the implementation with a `foldr` fixes those exclamation marks... but introduces more elsewhere! Any idea what's going on?

```
module "\"/Users/gelisam/working/haskell/klister/examples/group.kl\""
    import #[group.kl:2.16-2.22]<"kernel"> ⇑ 1
    define a : Signal := Signal 1
    meta
        define a : Signal := Signal 10
    define b : Signal := a
    define c : Signal := b
    meta
        define b : Signal := a
        define c : Signal := b
    example@group.kl:14.1-15.1 a : Signal
    example@group.kl:15.1-16.1 b : Signal
    example@group.kl:16.1-17.1 c : Signal
    meta
        example@group.kl:18.3-19.3 a : Signal
        example@group.kl:19.3-20.3 !!(Var 117)!! : Signal
        example@group.kl:20.3-20.14 !!(Var 131)!! : Signal
```

david-christiansen:
I'm on my way to bed with a migraine, so I can only give a little guess,
which is that the definitions aren't being added to the Env at the earlier
phase in which they are valid.

src/Pretty.hs
gelisam: Ah, I finally understand what's going on! Due to its `Env -> (..., Env)` signature, I keep incorrectly assuming that it behaves like a State, by returning a version of the input Env extended with some new variables. I now realize that the output Env only includes the newly-bound variables. For this decl list, this makes things a bit awkward because we do need to thread an Env through, but that Env must only contain the newly-bound variables, so it must start with `mempty`, not with the input Env. That input env must then be added to the accumulated Env each time we call `ppEnv` on one of the decls. After that we can accumulate the variables newly-bound by this decl into the Env we are threading through.

gelisam:
All the items in my TODO are complete; pending approval, I believe this PR is finally ready to merge!

repl/Main.hs
```diff
@@ -95,7 +95,8 @@ mainWithOptions opts =
           mn <- moduleNameFromPath file
           ctx <- mkInitContext mn
           void $ execExpand initializeKernel ctx
-          st <- execExpand (visit mn >> getState) ctx
+          st <- flip execExpand ctx $ do
```
david-christiansen, line 98: Two questions:
1. Should there be a `completely` here?
2. It seems like this `do` is unnecessary - why not either get rid of it, or replace the `>>` with a line break?

src/Expander.hs
```diff
@@ -1191,17 +1195,19 @@ runTask (tid, localData, task) = withLocal localData $ do
...
-      expandOneDeclaration sc dest stx ph
-    ExpandMoreDecls dest sc stx waitingOn -> do
-      readyYet <- view (expanderDeclPhases . at waitingOn) <$> getState
+    ExpandDeclForms dest earlierScopeSet waitingOn outScopesDest stx -> do
+      -- The scopes in waitingOn (from the preceding declaration) haven't yet
+      -- been added to stx, while those in earlierScopeSet (from the
+      -- declarations before that) have. The scopes from both, plus the scopes
+      -- introduced by the declaration forms in stx, must all eventually be
+      -- placed in outScopesDest so that the code which isn't part of this
+      -- declaration group can have access to those declarations.
```
david-christiansen, line 1204: Good comment!

src/Expander.hs
david-christiansen: This is really good stuff.

src/Syntax.hs
```diff
@@ -97,6 +97,22 @@ flipScope p = adjustScope go
...
+    in a2
+  where
+    addUniversalScopes :: HasScopes a => a -> ScopeSet -> a
+    addUniversalScopes =
+      foldlOf (to ScopeSet.contents . _1 . folded)
+              addScope'
+
+    addSpecificScopes :: HasScopes a => a -> ScopeSet -> a
+    addSpecificScopes =
+      ifoldlOf (to ScopeSet.contents .> _2 .> ifolded <. folded)
```
david-christiansen, line 113: This line is quite mysterious to me, but since the code seems to work, I'll assume it's some lens magic that I need to go learn :-)

david-christiansen:
Do you know why Travis stopped working here?

repl/Main.hs
gelisam: Fixed in #38: https://github.com/gelisam/klister/pull/38/files#diff-98af7d285f36c701c8fdfea83eb20dc3R98-R100

src/Syntax.hs
gelisam: `ifoldlOf` is an "indexed" version of `foldlOf`. Here it has this type:

    ifoldlOf :: IndexedFold Phase ScopeSet Scope
             -> (Phase -> a -> Scope -> a)
             -> a -> ScopeSet -> a

An "indexed" Fold is just like a Fold, except it also grabs an index along the way. The `.>` and `<.` point to the place where the index is grabbed, and otherwise act like `.`. So the non-indexed `Fold ScopeSet Scope` would be `to ScopeSet.contents . _2 . folded . folded`, which starts from a `ScopeSet`, uses `ScopeSet.contents` to get a `(Set Scope, Map Phase (Set Scope))`, and then dives into the `Map` and the `Set` to collect all the `Scope`s. The indexed version does the same, but for each `Scope` it found, it also remembers which `Phase` it encountered along the way.

`foldlOf` is like `foldl`, but over the elements of a Fold instead of the elements of a Foldable. That is, it threads a function of type `a -> Scope -> a` through all the `Scope`s in order to compute a final `a`. The indexed version of `foldlOf` does the same, except the function also has access to the `Phase` in addition to the `Scope`.

gelisam:
> Do you know why Travis stopped working here?

There was a transient error about failing to download a script from gitlab.

PR #39, opened 2020-04-12
Refactor: define primitives separately

david-christiansen:
This makes the code a bit easier for me to follow, because things have
names and are not as deeply nested in where blocks.

### DIFF
--- a/src/Expander/Monad.hs
+++ b/src/Expander/Monad.hs
@@ -67,6 +67,9 @@ module Expander.Monad
   , mkInitContext
   -- ** Tasks
   , module Expander.Task
+  , clearTasks
+  , getTasks
+  , setTasks
   , forkAwaitingDefn
   , forkAwaitingMacro
   , forkAwaitingDeclMacro
--- a/src/Expander/Monad.hs
+++ b/src/Expander/Monad.hs
@@ -9,13 +9,26 @@ module Expander.Monad
   , module Expander.DeclScope
   -- * The expander monad
   , Expand(..)
+  , addBinding
+  , addImportBinding
+  , addDefinedBinding
+  , addLocalBinding
+  , addRootScope
+  , addRootScope'
+  , addModuleScope
+  , bind
+  , completely
   , constructorInfo
+  , currentEnv
   , currentPhase
   , currentBindingLevel
+  , currentTransformerEnv
   , datatypeInfo
   , inTypeBinder
   , dependencies
   , execExpand
+  , expandEval
+  , freshBinding
   , freshConstructor
   , freshDatatype
   , freshScope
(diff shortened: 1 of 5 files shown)


david-christiansen:
The force-push was just for a commit message typo

src/Expander/Primitives.hs
```diff
@@ -0,0 +1,578 @@
...
+makeModule :: DeclExpander -> Syntax -> Expand ()
+makeModule expandDeclForms stx =
+  view expanderModuleTop <$> getState >>=
+  \case
+    Just _ ->
+      error "TODO throw real error - already expanding a module"
+    Nothing -> do
+      bodyPtr <- newDeclTreePtr
+      modifyState $ set expanderModuleTop (Just bodyPtr)
+      Stx _ _ (_ :: Syntax, name, body) <- mustHaveEntries stx
```
gelisam, line 468: This line used to ask for `imports` and `exports` in addition to `name` and `body`. `resolveImports imports` and `buildExports exports` were both no-ops, so removing them makes sense, but I'm surprised not to see any corresponding change to any of the examples. Does that mean that the `#%module` primitive is untested?

src/Expander/Monad.hs
```diff
@@ -688,3 +708,106 @@ datatypeInfo datatype = do
...
+  pure a
+
+getTasks :: Expand [(TaskID, ExpanderLocal, ExpanderTask)]
+getTasks = view expanderTasks <$> getState
+
+setTasks :: [(TaskID, ExpanderLocal, ExpanderTask)] -> Expand ()
+setTasks = modifyState . set expanderTasks
+
+clearTasks :: Expand ()
+clearTasks = modifyState $ set expanderTasks []
```
gelisam, line 786: Did you mean to move `getTasks` and `clearTasks` here rather than copy them? `getTasks` and `clearTasks` are still in `src/Expander.hs`.

david-christiansen:
Nice catches!

It does indeed look like #%module isn't being used. This is a bug that
needs fixing! The solution is to make the reader treat #lang more like
Racket does.

I did indeed mean to move those - I'll fix this up then merge.

Thanks!

david-christiansen:
Do you agree that this structure is easier to read?

>

gelisam:
> This makes the code a bit easier for me to follow, because things have
> names and are not as deeply nested in where blocks.
> Do you agree that this structure is easier to read?

The nesting didn't bother me, and the string names of the primitives were good enough for me. But I do like that `src/Expander.hs` got smaller!

src/Expander/Primitives.hs
david-christiansen: I created an issue, #40.

PR #43, opened 2020-04-18
Error form

### LINKED ISSUE #41: Error primitive
We should have something like Haskell's `error`. The argument should be a syntax object with a message in it, so that it can report a source location for the error - this will be useful to make macros that expand to error cases. RE #36 .

david-christiansen:
Programs can now signal run-time errors.

Fixes #41.

### DIFF
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -99,7 +99,7 @@ mainWithOptions opts =
             visit mn
             getState
           case st of
-            Left err -> prettyPrintLn err *> fail ""
+            Left err -> prettyPrintLn err *> exitFailure
             Right result ->
               pure (mn, ctx, result)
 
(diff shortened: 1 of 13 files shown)


david-christiansen:
This is a step on the way to a pattern match macro with guards and the like. It's needed for the "nothing matched" fallback - the error can point at the originating syntax of the scrutinee.

repl/Main.hs
```diff
@@ -99,7 +99,7 @@ mainWithOptions opts =
             visit mn
             getState
           case st of
-            Left err -> prettyPrintLn err *> fail ""
+            Left err -> prettyPrintLn err *> exitFailure
```
david-christiansen, line 102: This change is to stop it from dumping "user error" when it exits.

PR #46, opened 2020-04-21
Add a primitive datatype facility

david-christiansen:
Kernel datatypes must be added with care, but they make it much easier
to extend with special-purpose data (here, the argument to the result
of `make-syntax-introducer`).

I see this as having a few use cases:
1. We'll need a representation of problems to get the stuck macro stuff going (where "problem" is the form of judgment that the macro is attempting to make evident, like "Expression of type T" or "Valid declaration" or "Pattern substituting value of type T1 into type T2", etc
2. We can simplify CoreF, eliminating things like Booleans
3. Many simple one-off datatypes are useful for things like controlling the behavior of a function. Now we don't have to rely on primitives for all these.


david-christiansen:
Just discovered that this needs a bit more work - hold off on review/merge for a bit.

david-christiansen:
OK, seems good to go now.

david-christiansen:
@gelisam Can you restart the failed bit of build? Looks to be a transitory Travis thing.

Issue #36, opened 2020-03-24
pmatch port


david-christiansen:
We should port Oleg's `pmatch`: https://github.com/webyrd/quines/blob/master/pmatch.scm

PR #47, opened 2020-04-22
Changes necessary to get pmatch to work

david-christiansen:
This is the equivalent of Racket/Scheme's make-syntax-introducer and generate-temporaries.


david-christiansen:
There's now a `pmatch` that seems to work (fixes #36).

examples/datatype-macro.kl
```diff
@@ -3,6 +3,7 @@
 (import (shift "n-ary-app.kl" 1))
 (import (shift "lispy-do.kl" 1))
 (import (shift "let.kl" 1))
+(import (shift "quasiquote.kl" 1))
```
gelisam, line 6: huh, why was that import not previously needed even though the file uses `quasiquote/loc`? Presumably, that same mysterious reason is also why `free-identifier-case-test.kl` and `examples/mcond-test.kl` now import `n-ary-app.kl`?

examples/import-list-and-do.kl
```diff
@@ -0,0 +1,6 @@
+#lang "n-ary-app.kl"
+
+(import "list-datatype.kl")
+
+(import "lispy-do.kl")
+
```
gelisam, line 6: This file should probably be part of #48, not this PR. In fact, since #48 claims that code such as this one fails, how come this PR passes CI?

examples/lambda-case.kl
```diff
@@ -0,0 +1,15 @@
...
+(import (shift "quasiquote.kl" 1))
+
+(define-macros
+  ([lambda-case
+    (lambda (stx)
+      (syntax-case stx
+        [(list (_ cases))
+         (pure
+          (quasiquote/loc stx
+            (lambda (x) (case x ,cases))))]))]))
```
gelisam, line 13: It's unfortunate that we can't use our `define-syntax-rule` for such a simple macro! I guess we should add support for the `...` syntax in `quasiquote`.

examples/list-datatype.kl
```diff
@@ -0,0 +1,22 @@
+#lang "n-ary-app.kl"
+
+(import "defun.kl")
+
+(datatype (List A)
+  (nil)
+  (:: A (List A)))
```
gelisam, line 7: does using `::` instead of `cons` make sense if it can't be used infix?

examples/list-datatype.kl
```diff
@@ -0,0 +1,22 @@
+#lang "n-ary-app.kl"
```
gelisam, line 1: To avoid confusion, we should probably rename my `list.kl` example, since it manipulates syntax objects rather than lists.

examples/list-datatype.kl
```diff
@@ -0,0 +1,22 @@
...
+  (case lst
+    [(:: x xs) (:: (f x) (map f xs))]
+    [(nil) (nil)]))
+
+(defun foldr (f base lst)
+  (case lst
+    [(:: x xs)
+     (f x (foldr f base xs))]
+    [(nil)
+     base]))
```
gelisam, line 19: this code makes me very happy, klister finally feels like a real programming language!

examples/datatype-macro.kl
david-christiansen: This is due to #48 getting fixed. Modules were transitively exporting their dependencies' exports! Fixing that meant writing exports correctly.

examples/import-list-and-do.kl
david-christiansen: Because I fixed #48 while I was at it :-) It came up due to definitions from list.kl shadowing things from list-datatype.kl while writing pmatch, even though they shouldn't have been in scope. Reordering imports worked around it, but not in a way that made sense.

examples/lambda-case.kl
david-christiansen: I don't think ... belongs as part of quasiquote, but a pattern language for syntax objects that's along the lines of Dybvig's syntax-case would sure be nice!

examples/list-datatype.kl
david-christiansen: We use prefix + all the time in Lisp-land. Seems analogous.

But really, I've just used syntax from Pie when in doubt. I'm not attached to this name.

examples/list-datatype.kl
david-christiansen: Perhaps syntax-list.kl?

examples/list-datatype.kl
david-christiansen: Yeah, it's getting more and more fun to actually use!

examples/list-datatype.kl
david-christiansen: Done in 049597f. I did "list-syntax.kl" because it scanned more naturally to me on further thought, but I can do `syntax-list.kl` if you prefer.

david-christiansen:
Do you think this is ready for a merge?

gelisam:
> Do you think this is ready for a merge?

If you want, but I haven't reviewed much of it yet. I multi-task, so the lack of new comments doesn't mean I'm done reviewing, it just means I got distracted by something else. I use GitHub's "Approve / Request Changes" button when I'm done.

david-christiansen:
Sounds good! Just wasn't sure :-)

examples/pmatch.kl
```diff
@@ -0,0 +1,145 @@
...
+                                  ,(combine temps)]
+                                 [(else other) ,kf]))))))])]))]))
+
+
+
+(datatype (Nat) (zero) (add1 (Nat)))
+
+(example (pmatch (zero) (else 't)))
+(example (pmatch (zero) ((else x) 't)))
+(example (pmatch (zero) (x 't)))
```
gelisam, line 117: `(else x)` and `x` both match anything? why do we need both?

gelisam: I bet I'm going to accidentally write the pattern `zero` instead of `(zero)` all the time. only having `,x` bind variables would be a lot less error-prone, I think (and would match the original algorithm)

examples/pmatch.kl
```diff
@@ -0,0 +1,145 @@
...
+                    (lambda (e)
+                      (case e
+                        [(just x) (pure `(let (,x ,tgt) ,rhs))]
+                        [(nothing)
+                         (pure `(let (fk (lambda (_) (pmatch-aux ,scrut ,tgt ,cs)))
+                                  (ppat ,tgt ,pat ,rhs (fk 'hi))))])))])])]))]
+   [ppat
+    (lambda (stx)
+      (syntax-case stx
+        [(list (_ tgt pat ks kf))
```
gelisam, line 93: earlier the continuation function was called `fk` and now it's called `kf`. which is it?

examples/pmatch.kl
```diff
@@ -0,0 +1,145 @@
...
+(meta
+  (datatype (Abstracted)
+    (none)
+    (bind Syntax Syntax (Abstracted)))
+
+  (defun abstract (stxs)
+    (syntax-case stxs
+      [() (none)]
+      [(cons e es)
+       (bind 'internal-x e (abstract es))])))
```
gelisam, line 63: `Abstracted`, `none`, and `bind` aren't used anywhere, I'm guessing this is a leftover from a previous implementation?

examples/pmatch.kl
```diff
@@ -0,0 +1,145 @@
...
+   [pmatch-aux
+    (lambda (stx)
+      (syntax-case stx
+        [(list (_ scrut tgt pats))
+         (syntax-case pats
+           [() (pure `(error ',scrut))]
+           [(cons c cs)
+            (syntax-case c
+              [(list (pat rhs))
+               (>>= (else-case pat)
```
gelisam, line 83: Why is the else cases treated separately? I would have expected them to be handled by `ppat`, just like every other pattern.

examples/pmatch.kl
```diff
@@ -0,0 +1,145 @@
...
+            (>>= (make-temporaries (syntax->list args))
+                 (lambda (temps)
+                   (let (temp-names (map fst temps))
+                     (flet (combine (stxs)
+                             (case stxs
+                               [(nil) ks]
+                               [(:: id-and-stx rest)
+                                `(ppat ,(fst id-and-stx) ,(snd id-and-stx) ,(combine rest) ,kf)]))
+                        (pure `(case ,tgt
+                                 [,(cons-list-syntax what (the Syntax (list->syntax temp-names pat)) pat)
```
gelisam, line 107: What is the purpose of the `the Syntax` type annotation? The name `list->syntax` already communicates quite clearly that it is returning a `Syntax`. I am guessing that this is leftover from a previous debugging session?

gelisam: On that note, I had completely forgotten that Klister was a typed language while reading this code. How about defining versions of `define`, `let` and `flet` which encourages including a type signature?

examples/pmatch.kl
david-christiansen: `else` unambiguously binds the variable `x`, while `x` alone may be a macro that expands to a pattern case using the normal `case` machinery.

As far as the comma goes, I am not a fan of the aesthetics of using an antiquotation operator for something that really isn't antiquotation...

examples/pmatch.kl
david-christiansen: To be pedantic, the earlier thing is a function, while this is a local name for a piece of code. But I'll call them all "kf".

examples/pmatch.kl
david-christiansen: Yep! Will delete.

examples/pmatch.kl
david-christiansen: The annotation is going away - it was for sanity while I was afflicted by #48.

For the `define` etc, we would need a way to annotate type schemes, which we don't yet have. See #60 for a design sketch.

examples/pmatch.kl
david-christiansen: Else cases are only valid at the top of a given pattern, while `ppat` is invoked recursively.

examples/pmatch.kl
david-christiansen: Just to be clear - `x` does not presently get to be a macro invocation, but this syntax at least does not rule that out!

examples/pmatch.kl
david-christiansen: Done.

src/Value.hs
```diff
@@ -24,16 +24,22 @@ data MacroAction
   | MacroActionWaitSignal Signal
   | MacroActionIdentEq HowEq Value Value
   | MacroActionLog Syntax
+  | MacroActionIntroducer
   | MacroActionWhichProblem
-  deriving (Eq, Show)
+
+instance Show MacroAction where
+  show _ = "MacroAction..."
```
gelisam, line 31: Meh, if these Show instances aren't going to give any information, should we even bother giving them a Show instance at all?

david-christiansen: Because other things that include a MacroAction do have useful Show instances that are derived.

Issue #48, opened 2020-04-23
Scoping rules broken when importing many modules


david-christiansen:
The modules `examples/list-datatype.kl` and `examples/lispy-do.kl` each type check independently. Thus, importing them should never break.

However, here's two files that break:
```
#lang "n-ary-app.kl"

(import "list-datatype.kl")

(import "lispy-do.kl")
```
gives
```
No progress was possible:
    (ExpandVar Ty {unTy = TFun (Ty {unTy = TMetaVar (MetaPtr 3286)}) (Ty {unTy = TFun (Ty {unTy = TMetaVar (MetaPtr 3272)}) (Ty {unTy = TSyntax})})} (SplitCorePtr 3287) Syntax {_unSyntax = Stx {_stxScopeSet = ScopeSet {_universalScopes = fromList [Scope {scopeNum = 307, scopePurpose = "Module root for /home/davidc/Code/Haskell/stuck-macros/examples/let.kl"},Scope {scopeNum = 308, scopePurpose = "For import at les/let.kl:3.1-4.1"},Scope {scopeNum = 309, scopePurpose = "For import at les/let.kl:4.1-5.1"},Scope {scopeNum = 477, scopePurpose = "For import at les/let.kl:5.1-6.1"},Scope {scopeNum = 478, scopePurpose = "For import at les/let.kl:6.1-8.1"}], _phaseScopes = fromList [(Phase {phaseNum = 0},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"},Scope {scopeNum = 479, scopePurpose = "For macros at les/let.kl:8.1-27.1"}]),(Phase {phaseNum = 1},fromList [Scope {scopeNum = 3, scopePurpose = "Root for phase p1"},Scope {scopeNum = 482, scopePurpose = "For variable \"stx\""},Scope {scopeNum = 484, scopePurpose = "For variable \"_\""},Scope {scopeNum = 485, scopePurpose = "For variable \"args\""},Scope {scopeNum = 486, scopePurpose = "For variable \"body\""}])]}, _stxSrcLoc = SrcLoc {_srcLocFilePath = "/home/davidc/Code/Haskell/stuck-macros/examples/let.kl", _srcLocStart = SrcPos {_srcPosLine = 21, _srcPosCol = 23}, _srcLocEnd = SrcPos {_srcPosLine = 21, _srcPosCol = 26}}, _stxValue = Id "map"}} (Var 1252))
[...]

PR #49, opened 2020-04-24
WIP: Type error display improvements

david-christiansen:
While writing `pmatch`, it was often difficult to figure out what Klister meant when giving type errors. Many of them were due to #48, it turned out, but that was made hard to see. This PR doesn't make it good, but it is at least non-terrible now:
 1. Type errors are aggregated within a compilation unit (that is, a dyamic extent delimited by a `completely`) and reported as a group.
 2. Unification failures display the complete types that failed to unify, as well as their specific sub-components that failed, when these differ.

This is not quite ready for review/merge yet, due to one downside: unification failures do not prevent the elaboration of a term that can't type check, but they should. As things stand now, ill-typed macro code could be executed (though not ill-typed phase 0 code). I need to ponder the right design for fixing this - probably, the stuff in Expander.TC should run in a Validation monad with a bit of local state, rather than an Error monad with the global expander state.

Also, the work is done on top of #47, because I don't expect to finish this before that's been merged, and because I have some little local tests that use `pmatch` in examples.

### DIFF
--- a/examples/datatype-macro.golden
+++ b/examples/datatype-macro.golden
@@ -1,4 +1,4 @@
 (:: Signal 3 (:: Signal 2 (:: Signal 1 (nil)))) : (List Signal)
 (:: Signal 3 (:: Signal 2 (:: Signal 1 (nil)))) : (List Signal)
-#[datatype-macro.kl:48.38-48.39]<a> : Syntax
+#[datatype-macro.kl:49.38-49.39]<a> : Syntax
 Signal 3 : Signal
--- a/examples/datatype-macro.kl
+++ b/examples/datatype-macro.kl
@@ -3,6 +3,7 @@
 (import (shift "n-ary-app.kl" 1))
 (import (shift "lispy-do.kl" 1))
 (import (shift "let.kl" 1))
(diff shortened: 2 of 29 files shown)


david-christiansen:
Having thought about this for a while, I think I have an idea of how to proceed. First of, we must recognize that macros are the source of all compile-time execution, and free variables in macros always refer to either imports or definitions (rather than e.g. `let` bindings). We do not allow modules to load others with type errors. This means that it is sufficient to ensure that no macro that transitively refers to a definition with a type error is ever executed. 

I propose to do that by maintaining a mapping from defined names to their type errors and their dependencies' type errors. When we add a macro to the transformer environment (that is, the handler for `TaskAwaitingMacro`, we first check whether it has type errors or the free variables are definitions that transitively have type errors. If so, we replace the macro with a dummy macro action that explains why it can't run, pointing at the type errors that must be fixed.

This means that modules with type errors at phase 1 or higher may not report all type errors, because a macro failing to run may interrupt expansion of the rest of the module. But I think it's a reasonably simple compromise that's much better than only ever showing one type error at a time!

PR #50, opened 2020-04-27
which-problem

gelisam:
A macro-action which a macro can ask in order to discover whether it should expand to a declaration, an expression, a type, or a pattern.

### DIFF
--- a/src/Core.hs
+++ b/src/Core.hs
@@ -107,6 +107,7 @@ data CoreF pat core
   | CoreIdentEq HowEq core core         -- bound-identifier=? :: Syntax -> Syntax -> Macro Bool
                                         -- free-identifier=?  :: Syntax -> Syntax -> Macro Bool
   | CoreLog core
+  | CoreWhichProblem
   | CoreSyntax Syntax
   | CoreCase core [(SyntaxPattern, core)]
   | CoreIdentifier Ident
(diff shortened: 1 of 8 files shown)


src/Expander.hs
```diff
@@ -1025,72 +1027,79 @@ expandLiteralString _dest str =
 -- If we're stuck waiting on a signal, we return that signal and a continuation
 -- in the form of a sequence of closures. The first closure should be applied to
 -- the result of wait-signal, the second to the result of the first, etc.
-interpretMacroAction :: MacroAction -> Expand (Either (Signal, [Closure]) Value)
```
gelisam, line 1028: mostly whitespace changes:

```
interpretMacroAction (...) = ...
interpretMacroAction (...) = ...
```

to

```
interpretMacroAction prob = \case
  (...) -> ...
  (...) -> ...
```

plus the new `MacroActionWhichProblem` case.

src/Core.hs
```diff
@@ -107,6 +107,7 @@ data CoreF pat core
   | CoreIdentEq HowEq core core         -- bound-identifier=? :: Syntax -> Syntax -> Macro Bool
                                         -- free-identifier=?  :: Syntax -> Syntax -> Macro Bool
   | CoreLog core
+  | CoreWhichProblem
```
gelisam, line 110: Now that I experienced first hand how many places I need to edit in order to add a new macro action, I agree that the next step should be to make it easier to introduce new macro actions :)

david-christiansen:
I just restarted the build - looks like the usual unreliable Travis thing.

Assuming it passes, I think this is good to merge. Great!

PR #51, opened 2020-04-27
Tiny fixes

gelisam:


src/Expander/Error.hs
```diff
@@ -140,11 +140,11 @@ instance Pretty VarInfo ExpansionErr where
     vsep [ pp env stx <> text ":"
          , group $ vsep [ group $ hang 2 $
                           vsep [ text "Used in a position expecting"
-                               , pp env expected
+                               , pp env actual
```
david-christiansen, line 143: It looks to me like the variable names are also wrong here...

gelisam: fixed, thanks

PR #52, opened 2020-04-28
Type error display

david-christiansen:
This is the part of #49 that I think is basically non-controversial and involves no major changes to anything. Might as well get it merged once #47 is in.

This does the following, on top of #47:
1. Unification errors show which sub-terms of the type didn't match, but also the whole type being unified, like GHC's "specifically" clauses
2. Unification errors contain the zonked type rather than one with gratuitous metas in it
3. Minor code changes for readability


david-christiansen:
Yeah, that'd be nice :)

PR #53, opened 2020-04-30
Type case operator that gets stuck on unsolved metas (i.e. THE DREAM ACHIEVED)

david-christiansen:
This is based on #47, because of the various bug fixes and such in it. On top of #47's changes, it has the following:
 * The `expression` constructor of `Problem` has been given a primitive representation of types at the next phase as an argument (the new built-in type `Type`) 
 * The new `type-case` operator allows matching on a `Type` in a monadic context, getting stuck if that type is not yet determined. Type patterns are expansion positions, just like data patterns.

Things to be improved over time:
 * If `type-case` never gets un-stuck, the error message is quite bad. We probably need some provenance tracking for metas to tell it where they are from, at the very least.


examples/which-problem.kl
```diff
@@ -7,12 +7,38 @@
...
+              [(type)         (pure '(Bool))]
+              [(expression t) (pure '(true))]
+              [(pattern)      (pure '(unit))]))))]))
 
 (m)
 
 (example (case (unit)
            ((m) (true))))
+
+(define-macros
```
langston-barrett, line 20: This is so cool :smile:

david-christiansen: We could try to imlement something like Cryptol's zero operator using this. That'd be fun :-)

examples/temporaries.kl
```diff
@@ -0,0 +1,26 @@
+#lang "n-ary-app.kl"
+
+(import "lispy-do.kl")
+(import "list-datatype.kl")
+(import "pair-datatype.kl")
+(import "defun.kl")
+
+
+(define make-temporary
```
langston-barrett, line 9: What is the point of make-temporary?

david-christiansen: It is part of #47.

 The point is to get a fresh name that can't shadow any existing name without having to make a trampoline through the macro expander, which can be nice when programmatically generating an unknown number of binders. The alternative is a CPS'd macro that generates them and then continues with them, but that becomes too hard to follow.

See `generate-temporaries` in Scheme and Racket - this is the unary case of that.

langston-barrett: Okay, thanks for the reference!

package.yaml
```diff
@@ -15,6 +15,7 @@ dependencies:
   - lens
   - megaparsec
   - mtl
+  - n-ary-functor
```
gelisam, line 18: wow, you used my n-ary-functor package! I'm so proud! but... where are you using it? I don't see any `import NAryFunctor`.

david-christiansen:
I used it and then unused it again - will fix!

PR #57, opened 2020-05-02
KLISTERPATH

### LINKED ISSUE #56: KLISTERPATH
In support of #54, it would be nice to import modules by something other than their actual filepath. I'll work on something like PYTHONPATH for Klister.

langston-barrett:
Fixes #56 

Here's an example of it finding the import properly:
```bash
$ cat examples/path.kl

#lang kernel
(import "in-other-dir.kl")
(example works)

$ cat examples/path/in-other-dir.kl

#lang kernel

(define works (lambda (x) x))

$ unset KLISTERPATH

$ cabal run klister -- run examples/path.kl
Up to date
Module not found on $KLISTERPATH: in-other-dir.kl
ExitFailure 1

$ export KLISTERPATH=$PWD/examples/path

$ cabal run klister -- run examples/path.kl
Up to date
Unknown: #[path.kl:3.10-3.15]<works>
ExitFailure 1
```

### DIFF
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -527,10 +527,6 @@ primImportModule dest outScopesDest importStx = do
   linkDeclOutputScopes outScopesDest (ScopeSet.singleUniversalScope sc)
   where
     importSpec :: Syntax -> Expand ImportSpec
-    importSpec (Syntax (Stx scs srcloc (String s))) =
-      ImportModule . Stx scs srcloc <$> liftIO (moduleNameFromLocatedPath srcloc (T.unpack s))
-    importSpec (Syntax (Stx scs srcloc (Id "kernel"))) =
-      return $ ImportModule (Stx scs srcloc (KernelName kernelName))
     importSpec stx@(Syntax (Stx _ _ (List elts)))
       | (Syntax (Stx _ _ (Id "only")) : spec : names) <- elts = do
           subSpec <- importSpec spec
@@ -546,7 +542,7 @@ primImportModule dest outScopesDest importStx = do
         Stx _ _ p <- mustBeIdent prefix
         return $ PrefixImports subSpec p
       | otherwise = throwError $ NotImportSpec stx
-    importSpec other = throwError $ NotImportSpec other
+    importSpec modStx = ImportModule <$> mustBeModName modStx
     getRename s = do
       Stx _ _ (old', new') <- mustHaveEntries s
       old <- mustBeIdent old'
--- a/src/KlisterPath.hs
+++ b/src/KlisterPath.hs
@@ -1,0 +1,91 @@
+{-
+Module           : KlisterPath
+Description      : Implementation of $KLISTERPATH
+Copyright        : (c) Klister Collaborators 2020
+License          : See GH#55
+Maintainer       : Langston Barrett <langston@galois.com>
+Stability        : provisional
+-}
+
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+
+module KlisterPath
+  ( KlisterPath
+  , KlisterPathError(..)
+  , ppKlisterPathError
+  , getKlisterPath
+  , makeKlisterPath
+  , importFromKlisterPath
+  ) where
+
+import Data.Functor ((<&>))
+import qualified Data.Set as Set
+import Data.Set (Set)
+import qualified Data.Text.Prettyprint.Doc as PP
+import Data.Traversable (for)
+import System.Directory (canonicalizePath, listDirectory)
+import System.FilePath.Posix ((</>))
+import System.Environment (lookupEnv)
+
+-- | A list of directories to search for modules
+--
+-- Invariant: All paths are canonical.
+newtype KlisterPath = KlisterPath { unKlisterPath :: Set FilePath }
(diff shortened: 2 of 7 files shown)


langston-barrett:
@david-christiansen I don't understand why the import doesn't bring `works` into scope. If I use the `only` import form it tells me that `works` is not available at `p0`.

langston-barrett:
This is flawed when importing a file which is in both the same diretory as the importing file and that directory is on the `KLISTERPATH`:
```
Too many versions of this module on $KLISTERPATH: n-ary-app.kl
Candidates:
/home/siddharthist/code/klister/stdlib/n-ary-app.kl
/home/siddharthist/code/klister/stdlib/n-ary-app.kl
```
We should maintain the invariant that the entries in `KlisterPath` are unique.

david-christiansen:
Like Racket, and unlike Haskell, Klister modules have to explicitly export bindings that they want to be available elsewhere. Try adding `(export works)` to the end of `in-other-dir.kl`.

src/KlisterPath.hs
```diff
@@ -0,0 +1,91 @@
...
+    split _ [] = []
+    split d s = x : split d (drop 1 y)
+      where (x, y) = span (/= d) s
+
+data KlisterPathError
+  = ModuleNotFound String
+  | TooManyFound String [FilePath]
+  deriving (Eq, Show)
+
+ppKlisterPathError :: KlisterPathError -> PP.Doc ann
```
david-christiansen, line 60: Why isn't this a `Pretty` instance? That would be more consistent with the rest of the code, at least.

src/KlisterPath.hs
```diff
@@ -0,0 +1,91 @@
...
+ppKlisterPathError =
+  \case
+    ModuleNotFound modName ->
+      PP.pretty $ "Module not found on $KLISTERPATH: " <> modName
+    TooManyFound modName found ->
+      PP.vsep $
+        [ PP.pretty $
+            "Too many versions of this module on $KLISTERPATH: " <> modName
```
david-christiansen, line 68: Perhaps this is better phrased as "ambiguous module name" rather than "too many versions"? The latter implies that related names are for related code, while I think that in reality, the names might coincidentally coincide.

I like that it actually gives the list of what it found :-)

src/KlisterPath.hs
```diff
@@ -0,0 +1,91 @@
...
+
+-- | Fetch the Klister path from the environment variable @KLISTERPATH@.
+getKlisterPath :: IO KlisterPath
+getKlisterPath =
+  makeKlisterPath =<<
+    maybe Set.empty (Set.fromList . (split ':')) <$>
+      lookupEnv "KLISTERPATH"
+  where
+    -- How is this not in the Prelude...?
+    split :: Eq a => a -> [a] -> [[a]]
```
david-christiansen, line 50: It seems to me that this will fail on directory names that contain a colon. OTOH, I don't think POSIX `PATH` works right in that way either, and AFAIK Windows doesn't even allow colons in directory names, so we're probably fine :-)

src/Expander.hs
```diff
@@ -111,7 +111,8 @@ expandModule thisMod src = do
   modifyState $ set expanderCurrentBindingTable mempty
   startDefTypes <- view expanderDefTypes <$> getState
   local (set (expanderLocal . expanderModuleName) thisMod) do
-    lang <- mustBeModName (view moduleLanguage src)
+    kPath <- klisterPath
+    lang <- mustBeModName kPath (view moduleLanguage src)
```
david-christiansen, line 115: Shouldn't `mustBeModName` be doing the `klisterPath` lookup? That'll prevent repeated code at each call site, and I don't expect that each caller has an interest in varying the path.

src/Expander.hs
```diff
@@ -527,8 +528,12 @@ primImportModule dest outScopesDest importStx = do
   linkDeclOutputScopes outScopesDest (ScopeSet.singleUniversalScope sc)
   where
     importSpec :: Syntax -> Expand ImportSpec
-    importSpec (Syntax (Stx scs srcloc (String s))) =
-      ImportModule . Stx scs srcloc <$> liftIO (moduleNameFromLocatedPath srcloc (T.unpack s))
+    importSpec (Syntax (Stx scs srcloc (String s))) = do
```
david-christiansen, line 531: Perhaps this should be using `mustBeModName`?

Like:
```haskell
mportSpec mod@(Syntax (Stx scs srcloc (String s))) =
  ImportModule <$> mustBeModName mod
```

```

src/Expander.hs
langston-barrett: Sure, good idea!

src/KlisterPath.hs
langston-barrett: Yeah, I don't think there's much to be gained by handling edge cases here :-)

src/KlisterPath.hs
langston-barrett: Will fix.

src/KlisterPath.hs
langston-barrett: Agreed!

langston-barrett:
Documentation thunked to a follow-up, since we don't have an pre-existing usage docs: https://github.com/gelisam/klister/issues/62

PR #59, opened 2020-05-02
stdlib and prelude

### LINKED ISSUE #54: Prelude/stdlib
We have a lot of useful syntax, datatypes, and functions them scattered throughout examples. We should collect these into a cohesive, organized set of modules that could be easily imported.

### LINKED ISSUE #64: Module search order
@david-christiansen suggested on #59 that imports be resolved thusly:

> What about resolving ambiguities thusly: first, check for adjacent files with the given name. Then, check for stdlib. Only after that, fall back on KLISTERPATH.

langston-barrett:
Progress on #54, requires #57, fixes #64

### DIFF
--- a/src/KlisterPath.hs
+++ b/src/KlisterPath.hs
@@ -1,10 +1,21 @@
 {-
 Module           : KlisterPath
-Description      : Implementation of $KLISTERPATH
+Description      : Searching for imports
 Copyright        : (c) Klister Collaborators 2020
 License          : See GH#55
 Maintainer       : Langston Barrett <langston@galois.com>
 Stability        : provisional
+
+Klister executes the following procedure when searching for imports:
+
+1. Check for files with the given name in the same directory as the importing
+   file
+2. Check for modules in the stdlib
+3. Check for modules in directories listed in the @KLISTERPATH@ environment
+   variable
+
+The stdlib modules are installed by Cabal, and accessed via the generated module
+"Paths_klister".
 -}
 
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
@@ -15,18 +26,22 @@ module KlisterPath
   , KlisterPathError(..)
   , ppKlisterPathError
   , getKlisterPath
-  , makeKlisterPath
-  , importFromKlisterPath
+  , findModule
   ) where
 
+import Control.Lens (view)
 import Data.Functor ((<&>))
 import qualified Data.Set as Set
 import Data.Set (Set)
 import qualified Data.Text.Prettyprint.Doc as PP
-import Data.Traversable (for)
 import System.Directory (canonicalizePath, listDirectory)
+import System.FilePath (takeDirectory)
 import System.FilePath.Posix ((</>))
 import System.Environment (lookupEnv)
+
+import ModuleName
+import Syntax.SrcLoc
+import Paths_klister (getDataDir)
 
 -- | A list of directories to search for modules
 --
@@ -53,15 +68,21 @@ getKlisterPath =
       where (x, y) = span (/= d) s
 
 data KlisterPathError
-  = ModuleNotFound String
+  = ModuleNotFound String FilePath KlisterPath
   | TooManyFound String [FilePath]
-  deriving (Eq, Show)
+  | TooManyFoundInStdlib String [FilePath]
+  deriving (Show)
 
 ppKlisterPathError :: KlisterPathError -> PP.Doc ann
 ppKlisterPathError =
   \case
-    ModuleNotFound modName ->
-      PP.pretty $ "Module not found on $KLISTERPATH: " <> modName
+    ModuleNotFound modName parentDir (KlisterPath kPath) ->
+      PP.vsep $
+        [ PP.pretty $ "Module not found: " <> modName
+        , PP.pretty $ "Searched in directory at " <> parentDir
+        , PP.pretty $ "And on $KLISTERPATH:"
+        , PP.vsep $ map PP.pretty $ Set.toList $ kPath
+        ]
     TooManyFound modName found ->
       PP.vsep $
         [ PP.pretty $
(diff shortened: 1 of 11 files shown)


langston-barrett:
Honestly, in the spirit of small, atomic changes, we could just merge this as-is and add more stuff to the prelude and stdlib in follow-up PRs.

david-christiansen:
I like having small atomic changes :-)

I have a concern about this way of doing it. I am afraid that it ties Klister to running from the root of the repository.

Shouldn't we be using Cabal's `data-files` feature to arrange for the stdlib to be available, so that Klister need not be run from the source repository?

stdlib/prelude.kl
```diff
@@ -0,0 +1,23 @@
+#lang "n-ary-app.kl"
+
+[define id [lambda (x) x]]
+[define const [lambda (x y) x]]
+[define compose [lambda (f g x) (f (g x))]]
+(define compose* (flet (comp (f g x) (f (g x))) comp))
+
+(export #%module #%app
```
david-christiansen, line 8: I actually think that `"prelude.kl"` should export the contents of `n-ary-app` again at phase 1. That is, it should import it shifted, and then repeat the same export list shifted again. The nice thing about this is that it makes them available in macros as well without any extra ceremony (kind of like how `#lang racket` exports the contents of `racket/base` at phase 1).

But this is something that can easily be added later!

langston-barrett:
I think that "stdlib" should be removed from Main.hs, then your KLISTERPATH determines where to find the prelude and standard library. Does that address your concern?

langston-barrett:
I like the idea of installing the stdlib using data-files.

david-christiansen:
> I think that "stdlib" should be removed from Main.hs, then your KLISTERPATH determines where to find the prelude and standard library. Does that address your concern?

It does not!

I'd like Klister programs that use the standard library to just work, regardless of the CWD or the setting of environment variables. In other words, the Klister executable should be able to find the `stdlib` directory to consider as part of its search path.

I think there's three reasonable ways to achieve this:
1. Require that the stdlib be near the executable. This means that users can't stick the executable in `/usr/bin`, so I'm not a big fan.
2. Use Cabal's `data-files` feature, which will point source-local builds at the `stdlib` in the source tree, and installed ones at something like `/usr/local/share/klister/stdlib`, but it means that we basically just ignore it and let Cabal solve it. Idris does this. In this solution, it's almost like the stdlib is baked into the compiler, which I like for the simplicity of it.
3. Use XDG directories, so it's something like `~/.local/share/klister`, perhaps with fallback to `/usr/local/share` or something. This lets people mess with their stdlib more easily.

I like 2, because the standard library is standard, and really part of the language, and should just always be there.

gelisam:
> I like having small atomic changes :-)

me too! large changes take me forever to review (sorry it's taking me so long)

david-christiansen:
One further issue to think of: with KLISTERPATH, I can break a library by adding a local file with a name that conflicts. The import in the library, which may be a purely internal thing, might become ambiguous with something I've just created.

What about resolving ambiguities thusly: first, check for adjacent files with the given name. Then, check for stdlib. Only after that, fall back on KLISTERPATH.

Or we could indicate these situations by expanding the concrete syntax of module paths a la Racket.

langston-barrett:
> What about resolving ambiguities thusly: first, check for adjacent files with the given name. Then, check for stdlib. Only after that, fall back on KLISTERPATH.

I like this! Eventually, we could also have logical module paths like Haskell or Coq, rather than a flat namespace. I think that would complement the search order.

I will make the change to use Cabal data files before merging.

langston-barrett:
> I think there's three reasonable ways to achieve this:

One other way (maybe this fell outside of "reasonable" :smile:) would be to bake the standard library into the Klister executable using Template Haskell. This has the advantage of always "just working" for the user, and has some performance advantages (no need to touch the filesystem, do expansion, nor parse when importing a stdlib module).

langston-barrett:
> Use Cabal's data-files feature, which will point source-local builds at the stdlib in the source tree, and installed ones at something like /usr/local/share/klister/stdlib, but it means that we basically just ignore it and let Cabal solve it. Idris does this. In this solution, it's almost like the stdlib is baked into the compiler, which I like for the simplicity of it.

For precedent on this approach, we can look to Cryptol: https://github.com/GaloisInc/cryptol/blob/7e841e915f34a21bbfad67070c7338662ac28601/cryptol.cabal

stdlib/prelude.kl
langston-barrett: Yes, let's go ahead and do this in a follow-up. I mostly moved these files here as a kind of proof-of-concept that importing the stdlib would work correctly. In another PR, we can make it actually useful.

src/KlisterPath.hs
```diff
@@ -53,39 +67,74 @@ getKlisterPath =
...
 ppKlisterPathError :: KlisterPathError -> PP.Doc ann
 ppKlisterPathError =
   \case
-    ModuleNotFound modName ->
-      PP.pretty $ "Module not found on $KLISTERPATH: " <> modName
+    ModuleNotFound modName parentDir (KlisterPath kPath) ->
+      PP.vsep $
+        [ PP.pretty $ "Module not found: " <> modName
+        , PP.pretty $ "Searched in directory at " <> parentDir
+        , PP.pretty $ "And on $KLISTERPATH:"
```
david-christiansen, line 82: Nice :)

src/KlisterPath.hs
```diff
@@ -53,39 +67,74 @@ getKlisterPath =
...
 findInKlisterPath (KlisterPath paths) moduleName =
-  Set.fromList . concat <$>
-    (for (Set.toList paths) $ \path ->
-      map (path </>) . filter (== moduleName) <$> listDirectory path)
+  concat <$> mapM (findInDirectory moduleName) (Set.toList paths)
 
-importFromKlisterPath ::
-  KlisterPath {- ^ Path to search -} ->
+stdlibPath :: IO String
+stdlibPath = (</> "stdlib") <$> getDataDir
```
david-christiansen, line 112: Minor, minor nitpick: would this be clearer with `<&>`?

Issue #60, opened 2020-05-02
More support for passing types around in the expander


david-christiansen:
Right now, almost all types (with the exception of annotations using `the`) are inferred. Even with macrology,`the` is insufficient, because it only supports ascribing types, rather than type schemes.

Here's a sketch of a design for definition annotations.

Desiderata:
1. Allow partial specification of types/schemes. We don't want to have an all-or-nothing approach, because part of the definition may come from one macro, and part from another. Something like Haskell's `PartialTypeSignatures` seems much more important here.
2. Respect lexical scoping. A hack like `ScopedTypeVariables` is not great, because definition forms really bind type variables simultaneously in the equivalent of a System F forall and big-lambda. 
3. Don't require large algorithmic changes in the type checker. Right now, the generalization machinery uses the notion of binding levels that I learned from @sestoft rather than scanning the context for free variables, and I'd like to stick with that mechanism for the sake of both efficiency and implementation simplicity.

Here's how I think we can accomplish these goals.

First off, we enrich each syntactic form that's subject to generalization (that is, `define`, `flet`, `let`, and `example`) with an extra set of parens that binds rigid variables, which we model as fresh type constants in the expander, with an encoding like the one used for names of datatypes. Example syntax in the kernel language:
```
(define (A) id (the (-> A A) (lambda (x) x)))
```
We can make this optional in non-kernel languages using a little macro. These bound type variables can only unify with themselves and with metas from their own scope (which means our notion of binding level might need to become a bit more informative, and instead talk about nested scopes rather than just being natural numbers - I'll think about this a bit). When it comes time to generalize, these turn into bound variables from the type scheme. Things to be aware of: 1. rigid type vars that don't get used don't get generalized - so it's "free" to bind a few. 2. unconstrained metas still get generalized, so ``(define (A) (lambda (x) (lambda (y) (pair (the A x) y))))`` will be a `(forall (A B) (-> A (-> B (Pair A B))))`.

Next, we need a way to provide partial type information to sub-invocations of macros (see @gelisam's [talk](https://youtu.be/nUvKoG_V_U0?t=2441) for the `setPartialType` operator and its usefulness). I think we can do this with another expression form, `with-unknown-type`. An expression
```
(with-unknown-type (M) EXPR)
```
elaborates `EXPR`, but with `M` bound to a new type metavariable. This will allow something like:
```
(with-unknown-type (A)
  (the (-> (List A) (List A)) (my-macro)))
```
[...]

PR #67, opened 2020-05-03
examples/contract: Simple function contracts

langston-barrett:
Allows defining simple boolean pre- and post-conditions for arguments and return values.
```lisp
(defun/contract id-bool (x) (true? true?) x)
```
means "I require that the argument `x` is `true?`, and will return a value that is `true?`.

### DIFF
--- a/examples/contract.kl
+++ b/examples/contract.kl
@@ -1,0 +1,78 @@
+#lang "n-ary-app.kl"
+[import "defun.kl"]
+[import "bool.kl"]
+[import [shift "n-ary-app.kl" 1]]
+[import [shift "defun.kl"1]]
+
+(meta
+  (defun contract-violation (contract)
+    (list-syntax ('error (quote '"Contract violation!")) contract)))
+
+(meta -- type annotation
+  (example (the (-> Syntax Syntax)
+                contract-violation)))
+
+(meta
+  (defun enforce-single (arg contract body)
+    (pure
+      (list-syntax
+        ('if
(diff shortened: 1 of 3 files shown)


examples/contract.kl
```diff
@@ -0,0 +1,78 @@
...
+[import "defun.kl"]
+[import "bool.kl"]
+[import [shift "n-ary-app.kl" 1]]
+[import [shift "defun.kl"1]]
+
+(meta
+  (defun contract-violation (contract)
+    (list-syntax ('error (quote '"Contract violation!")) contract)))
+
+(meta -- type annotation
```
david-christiansen, line 11: Ewww - I hope we get #60 done soon!

examples/contract.kl
```diff
@@ -0,0 +1,78 @@
...
+               (lambda (new-body)
+                 (pure
+                   (list-syntax
+                     ('defun f args new-body)
+                     stx))))]
+         [_ (syntax-error '"bad syntax" stx)])]]))
+
+(defun const (x y) x)
+(define any (const (true)))
+(defun true? (b) (if b (true) (false)))
```
david-christiansen, line 74: AKA `id` :-) But this makes sense for readability.

examples/contract.kl
langston-barrett: Indeed! It would also be nice to have a `defun/ann` form like `defun/contract` where the contracts are replaced by type annotations.

PR #71, opened 2020-05-06
Add with-unknown-type

david-christiansen:
This is half of #60. See that issue for a description of the new operator.

### DIFF
--- a/src/Expander/Monad.hs
+++ b/src/Expander/Monad.hs
@@ -213,12 +213,14 @@ newtype ExpansionEnv = ExpansionEnv (Map.Map Binding EValue)
   deriving (Semigroup, Monoid)
 
 data EValue
-  = EPrimExprMacro (Ty -> SplitCorePtr -> Syntax -> Expand ()) -- ^ For special forms
+  = EPrimExprMacro (Ty -> SplitCorePtr -> Syntax -> Expand ())
+    -- ^ For special forms
   | EPrimTypeMacro (SplitTypePtr -> Syntax -> Expand ()) (TypePatternPtr -> Syntax -> Expand ())
     -- ^ For type-level special forms - first as types, then as type patterns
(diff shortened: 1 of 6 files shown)


src/Expander/Monad.hs
```diff
@@ -213,12 +213,14 @@ newtype ExpansionEnv = ExpansionEnv (Map.Map Binding EValue)
...
 data EValue
-  = EPrimExprMacro (Ty -> SplitCorePtr -> Syntax -> Expand ()) -- ^ For special forms
+  = EPrimExprMacro (Ty -> SplitCorePtr -> Syntax -> Expand ())
+    -- ^ For special forms
   | EPrimTypeMacro (SplitTypePtr -> Syntax -> Expand ()) (TypePatternPtr -> Syntax -> Expand ())
     -- ^ For type-level special forms - first as types, then as type patterns
   | EPrimModuleMacro (Syntax -> Expand ())
   | EPrimDeclMacro (DeclTreePtr -> DeclOutputScopesPtr -> Syntax -> Expand ())
   | EPrimPatternMacro (Either (Ty, Ty, PatternPtr) (Ty, TypePatternPtr) -> Syntax -> Expand ())
+  | EPrimUniversalMacro (MacroDest -> Syntax -> Expand ())
```
langston-barrett, line 223: What is a "universal" macro?

david-christiansen:
Universal in the sense that it works in any syntactic context, be it
expression, type, decl, etc.

src/Expander/Monad.hs
david-christiansen: Universal in the sense that it works in any syntactic context, be it expression, type, decl, etc.

langston-barrett: What is its meaning in a pattern context? Looks like an error? And in a declaration context? What about a module context? (Not trying to be annoying, just trying to understand :smile:)

PR #72, opened 2020-05-07
n-ary arrow

gelisam:
So that the example in the extended abstract looks nicer.

### DIFF
--- a/examples/which-problem.kl
+++ b/examples/which-problem.kl
@@ -1,8 +1,9 @@
 #lang "n-ary-app.kl"
 
 (import (shift "n-ary-app.kl" 1))
-(import "quasiquote.kl")
 (import (shift "quasiquote.kl" 1))
+(import "quasiquote.kl")
+(import "define-syntax-rule.kl")
 
 (define-macros
   ([m (lambda (stx)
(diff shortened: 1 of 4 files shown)


examples/which-problem.kl
```diff
@@ -28,13 +30,11 @@
...
-                       (pure (list-syntax ('lambda
-                                           (list-syntax ('_) stx)
-                                           (list-syntax ('mega-const e) e))
-                                          stx))]
+                       (pure `(lambda (_)
+                                (mega-const ,e)))]
                       [(else x) (pure e)])])))]))]))
 
-(example (the (-> (Bool) (-> (Bool) (-> (Bool) (-> (Bool) (Unit))))) (mega-const (unit))))
+(example (the (-> (Bool) (Bool) (Bool) (Bool) (Unit)) (mega-const (unit))))
```
gelisam, line 37: I think it would look even nicer if there weren't parentheses around `Bool`, `Unit` and `unit`, is there a particular reason for having them?

gelisam:
Next, I'm thinking of sugaring `(define-macros ([mega-const (lambda (stx) (syntax-case stx [(list (_ e)) ...]))]))` to `(define-macro (mega-const e) ...)`

gelisam:
...it was easier than I thought!

examples/which-problem.kl
david-christiansen: Yes - it's to have a uniform syntax for type construction, especially once we get higher-kinded stuff. The idea is that the core language is as simple and uniform as possible, and then we can define macros to make it prettier.  FWIW, this is one of the reasons for having patterns and type patterns be macro expansion positions.

Try something like this:
```klister
(define-macros
  ([Unit* (lambda (stx) (syntax-case stx [(ident u) (quasiquote/loc stx Unit)]))]))
(export (rename Unit* Unit))
```

david-christiansen:
This is nice to have - thanks!

PR #73, opened 2020-05-09
make parens optional for builtin types and constructors

gelisam:
Ideally I would also wrap `datatype` so that the nullary types and constructors it generates also have optional parentheses, but I only did builtins for now because that's all we need to make the extended abstract's example look nice.

### DIFF
--- a/stdlib/n-ary-app.kl
+++ b/stdlib/n-ary-app.kl
@@ -109,47 +109,5 @@
                                           (cons-list-syntax 'my-arrow args stx))
                                         stx))])]))]))
 
-(meta
-  (define optional-parens
-    (lambda (symbol)
-      (lambda (stx)
-        (syntax-case stx
-          [(ident x)
-           (pure (list-syntax (symbol) stx))]
-          [(cons x xs)
-           (pure (cons-list-syntax symbol xs stx))])))))
-
--- TODO: wrap 'datatype' so that parentheses are also optional for
--- the nullary types and constructors it creates
-(define-macros
-  ([my-Unit (optional-parens 'Unit)]
-   [my-unit (optional-parens 'unit)]
-   [my-Bool (optional-parens 'Bool)]
-   [my-true (optional-parens 'true)]
-   [my-false (optional-parens 'false)]
-   [my-nothing (optional-parens 'nothing)]))
-
-(export (rename ([my-arrow ->]
-                 [my-Unit Unit] [my-unit unit]
-                 [my-Bool Bool] [my-true true] [my-false false]
-                 [my-nothing nothing])
-                my-arrow
-                my-Unit my-unit
-                my-Bool my-true my-false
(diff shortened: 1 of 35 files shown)


stdlib/n-ary-app.kl
```diff
@@ -109,23 +109,47 @@
...
+                 [my-nothing nothing])
+                my-arrow
+                my-Unit my-unit
+                my-Bool my-true my-false
+                my-nothing))
 (export #%module #%app
         lambda define example define-macros quote meta
         the Syntax Macro
         with-unknown-type
-        if true false
```
gelisam, line 117: Apparently we were previously exporting `true` and `false` twice, should that be an error?

At some point I was both exporting `my-true` as true in the above `export (rename ...)` and `true` as `true` here, thus leading to mysterious errors. So I think exporting different values with the same name should definitely be an error.

langston-barrett: Yeah, sounds like an error to me. #74

stdlib/n-ary-app.kl
```diff
@@ -109,23 +109,47 @@
                                           (cons-list-syntax 'my-arrow args stx))
                                         stx))])]))]))
 
-(export (rename ([my-arrow ->]) my-arrow))
+(meta
```
langston-barrett, line 112: I don't think this is the right place for this code, as it's not related to nary application. Perhaps we could have a new module, and re-export these identifiers from `prelude.kl`?

stdlib/n-ary-app.kl
```diff
@@ -109,23 +109,47 @@
...
+   [my-Bool (optional-parens 'Bool)]
+   [my-true (optional-parens 'true)]
+   [my-false (optional-parens 'false)]
+   [my-nothing (optional-parens 'nothing)]))
+
+(export (rename ([my-arrow ->]
+                 [my-Unit Unit] [my-unit unit]
+                 [my-Bool Bool] [my-true true] [my-false false]
+                 [my-nothing nothing])
+                my-arrow
```
langston-barrett, line 136: Does this also export `my-arrow` etc? Would we want to use those names in other modules?

stdlib/n-ary-app.kl
gelisam: Good point. Moved to `stdlib/builtins.kl`.

`n-ary-app.kl` kind of became our unofficial prelude, and so it re-exports a ton of stuff which have nothing to do with n-ary functions, and a ton of examples use `#lang "n-ary-app.kl"`. I've changed it to only export its new macros, and I've changed all the examples to use `#lang "prelude.kl"` instead.

stdlib/n-ary-app.kl
gelisam: No, `(rename ([a b]) a)` only exports `b`, it does not also export `a`.

PR #77, opened 2020-05-09
examples: test for import renaming

langston-barrett:
We didn't yet have a test for this!

### DIFF
--- a/examples/import-import-renaming.kl
+++ b/examples/import-import-renaming.kl
@@ -1,0 +1,4 @@
+#lang kernel
+(import (rename "bool.kl" (not negb)))
+(import "import-renaming.kl")
+(example (negb (false)))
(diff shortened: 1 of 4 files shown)


examples/import-import-renaming.kl
```diff
@@ -0,0 +1,4 @@
+#lang kernel
+(import (rename "bool.kl" (not negb)))
+(import "import-renaming.kl")
+(example (negb (false)))
```
gelisam, line 4: which of the two `negb` imports should take precedence? the renamed one or the one from `import-renaming. kl`?

langston-barrett: Oh, I didn't intend to leave the first in there actually :sweat_smile: copy/paste error

david-christiansen:
Sorry for the slow response!

PR #78, opened 2020-05-20
fix the llet example

gelisam:


src/Expander.hs
```diff
@@ -706,9 +706,8 @@ runTask (tid, localData, task) = withLocal localData $ do
           case ty of
             Ty (TMetaVar ptr) | ptr == ptr' -> stillStuck tid task
             _ -> forkAwaitingTypeCase loc dest (Ty (TMetaVar ptr')) env cases kont
-        other -> do
-          result <- expandEval $ withEnv env $ doTypeCase loc (Ty other) cases
-          forkContinueMacroAction dest result kont
+        _ -> do
+          forkInterpretMacroAction dest (MacroActionTypeCase env loc ty cases) kont
```
gelisam, line 710: This fixes the problem, but feels wrong. This delegates to another implementation of `TypeCase`:

https://github.com/gelisam/klister/blob/40796716ae67ac707159aa505702355ec5056e33/src/Expander.hs#L1192-L1202

Both implementations seem to do the same thing, but I guess the other one is less buggy. The fact that there was duplicated code in the first place is already suspicious. But what really feels wrong here is that this is the code which runs the `AwaitingTypeCase` task... and it now does so by forking a `TypeCase`. I feel like this should be the other way around: forking a `TypeCase` should schedule an `AwaitingTypeCase` task. But I haven't quite figured out how to move the implementation from there to here.

gelisam: At least I do understand why the error message we got was "expecting syntax, got macro action". We were calling `forkContinueMacroAction`, which can be used in `(>>= mx f)` to send the value produced by `mx` to `f`, and can also be used when we're done evaluating the macro, to send the Syntax back to the expander. But we don't yet have a value (Syntax or not), we've only picked a branch; that is, we only have a macro action.

gelisam: I have now pushed a much better fix, in which `interpretMacroAction` now delegates to `runTask` instead of the other way around. This matches the way in which `interpretMacroAction` implements `AwaitSignal`: by unconditionally forking a task which waits on the signal, regardless of whether the signal is already available or not.

PR #75, opened 2020-05-09
stdlib: import defun.kl, export from prelude at p0

langston-barrett:
This one seems useful enough to put in the prelude.

PR #80, opened 2020-05-24
ambiguous-identifier bug?

gelisam:
I encountered a similar error message to [the one @langston-barrett did](https://github.com/gelisam/klister/pull/75#issuecomment-626227462), so I turned it into a minimal example. Seems like a bug to me, but I wouldn't be surprised if it was just an expected but unintuitive behaviour.

I don't want to merge this PR, it's just the easier to demonstrate and reproduce the problem with a PR than with an issue.


examples/m3.kl
```diff
@@ -0,0 +1,6 @@
+#lang kernel
+
+(import (shift "m2.kl" 1))
+(import "m2.kl")
+
+(example (m2))
```
gelisam, line 6: This fails with

    Ambiguous identifier in phase p0 #[m2.kl:9.15-9.17]<m1>
        ScopeSet {_universalScopes = fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"},Scope {scopeNum = 2, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/m2.kl"},Scope {scopeNum = 3, scopePurpose = "For import at ples/m2.kl:3.1-4.1"},Scope {scopeNum = 9, scopePurpose = "For import at ples/m2.kl:4.1-6.1"}], _phaseScopes = fromList [(Phase {phaseNum = 1},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}
        ScopeSet {_universalScopes = fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"},Scope {scopeNum = 2, scopePurpose = "Module root for /Users/gelisam/working/haskell/klister/examples/m2.kl"},Scope {scopeNum = 3, scopePurpose = "For import at ples/m2.kl:3.1-4.1"},Scope {scopeNum = 9, scopePurpose = "For import at ples/m2.kl:4.1-6.1"}], _phaseScopes = fromList [(Phase {phaseNum = 0},fromList [Scope {scopeNum = 0, scopePurpose = "Root for phase p0"}])]}

examples/m1.kl
```diff
@@ -0,0 +1,10 @@
+#lang kernel
+
+(import (shift kernel 1))
+
+(define-macros
+  ([m1
+    (lambda (stx)
+      (pure ''1))]))
```
gelisam, line 8: note that the bug disappears if `m1` is moved to `m2.kl`.

examples/m3.kl
```diff
@@ -0,0 +1,6 @@
+#lang kernel
+
+(import (shift "m2.kl" 1))
+(import "m2.kl")
```
gelisam, line 4: Only the non-shifted import is used, but the shifted import is needed in order to trigger the bug. Normally, importing the same module with and without a shift does not cause any problem, but in this case it does.

gelisam:
I implemented the potential solution we discussed (add a scope when `visit` visits a module at a particular phase) in 0818ff4, but it's causing a lot of tests to fail with `Unknown: <my-lam>`, so that's probably not the right solution. Probably because the use sites aren't tagged with that new scope. Any other ideas?

david-christiansen:
That solution had two parts - the first is to add the fresh scope to the
binders, getting generativity as you've done. The second was to also add it
to all syntax literals/quotations in the fully-expanded module just prior
to evaluation. Without the second step, it makes sense that it can't find
them.

Den tir. 2. jun. 2020 kl. 21.30 skrev Samuel Gélineau <
notifications@github.com>:

gelisam:
All right, I have added the scope to all the syntax objects. This helped a lot, most of the tests are passing now. There are still two failures, in `which-problem` and `pmatch`, which I have not yet investigated.

gelisam:
Taking a closer look at the `pmatch` failure, the problematic symbol is `#%app`, the candidates all have the `p0` and `quasiquote.kl` roots as universal scope sets, plus a `quasiquote.kl at phase pN` as a phase N, for N=1, 2, and 3. There should thus be only 3 candidates, but in fact there are 25, because the `quasiquote.kl at phase pN`s have different `scopeNum`s. I think this means `visit` is visiting each module more often than we think instead of reusing the cached copy if we've already seen a module at a particular phase.

In any case, the use site also has the `quasiquote.kl` root as a universal scope sets, plus the `p0` root at phase 0, a `quasiquote.kl at phase p1`, and plenty more. The `quasiquote.kl at phase p1`'s `scopeNum` matches that of exactly one candidate, so I don't understand why there is any ambiguity.

gelisam:
Oh! It's because we're at phase 0, so none of the `quasiquote.kl at phase pN`s are taken into account, and so all the candidates look identical, they all have the two roots and nothing else. Does that mean the `quasiquote.kl at phase pN` should be added universally?

gelisam:
all the tests pass when I add it universally, so... signs point to "yes"?

david-christiansen:
Yes, the idea was to do it at all phases simultaneously.

On a binding site, adding a scope restricts the identifiers that can refer to it. At a use site, it gives the identifier permission to refer to more things. The idea here is to make it so that each instantiation of a module has unique binding-reference relationships, _at all phases simultaneously_. Does that make sense?

src/Expander.hs
```diff
@@ -226,11 +226,189 @@ visit modName = do
               getState
   unless visitedp $ do
     let m' = shift i m -- Shift the syntax literals in the module source code
-    evalResults <- inPhase p $ evalMod m'
+    sc <- freshScope $ T.pack $ "For module-phase " ++ shortShow modName ++ "-" ++ shortShow p
+    let syntaxScopeSets :: Traversal' Syntax ScopeSet
```
david-christiansen, line 230: This seems generally useful to me - perhaps it should live in Syntax.hs?

langston-barrett:
Can we go ahead and merge this? Maybe with a follow-up issue to move the traversals to other modules? I'd like to work more on #75 which is blocked on this bug.

gelisam:
> Can we go ahead and merge this? Maybe with a follow-up issue to move the traversals to other modules? I'd like to work more on #75 which is blocked on this bug.

I was working on higher-kinded types because it seemed more important, but if you're blocked, I'll switch to this instead.

langston-barrett:
@gelisam No need to totally switch gears, it just looked like the only remaining task was a small refactoring so I was wondering if we could postpone that.

gelisam:
> No need to totally switch gears, it just looked like the only remaining task was a small refactoring so I was wondering if we could postpone that.

There are quite a few things I don't like in this PR: the names of the `m1`, `m2`, and `m3` modules, the lack of explanation of what bug they're trying to reproduce, the super-long Traversals (maybe some data-generic programming can simplify them?), and the temporary nature of the debugging info I added to the error message. I've fixed the last of those, I'll try to address the other 3 issues today, and if it that ends up taking too much time, maybe we can indeed merge this as-is in order to unblock you, and add the rest as issues?

langston-barrett:
@gelisam Sounds good to me!

src/Expander.hs
gelisam: even better, I've managed to replace these very long Traversal definitions with a generic `allScopeSets` which can produce a `Traversal` of all the `ScopeSet`s in any type which has a `Data` instance.

gelisam:
One more down, two (or rather three, there's a conflict now) to go! Better make a list.

- [x] polish the error message
- [x] use Data instead of long traversals
- [ ] better names for `m1`, `m2` and `m3`
- [ ] comments explaining what `m1`, `m2` and `m3` are testing
- [ ] fix conflict

gelisam:
@langston-barrett All done! only waiting for a review.

examples/import-scoping.kl
```diff
@@ -0,0 +1,33 @@
+#lang kernel
+
+-- This tests a tricky case around imports and scoping.
```
langston-barrett, line 3: Thanks for this detailed explanation!

PR #81, opened 2020-05-24
deep-matching on syntax objects

gelisam:
`pmatch` is very convenient for deep-matching on datatypes, here is a version for syntax objects. That should make writing macros more pleasant!

### DIFF
--- a/examples/define-syntax-rule.kl
+++ b/examples/define-syntax-rule.kl
@@ -1,10 +1,12 @@
 #lang "prelude.kl"
 
 (import (shift "prelude.kl" 1))
+(import (shift "defun.kl" 1))
 (import (shift "do.kl" 1))
 (import (shift "list-syntax.kl" 1))
 (import (shift "quasiquote.kl" 1))
 (import (shift "syntax.kl" 1))
+(import (shift "temporaries.kl" 1))
 
 (define-macros
   ((define-macro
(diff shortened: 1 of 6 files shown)


examples/define-syntax-rule.kl
```diff
@@ -1,47 +1,110 @@
...
+  (lambda (stx)
+    (pure `(,'quote ,stx))))
+
+(define-macro (m2 arg)
+  (pure `(,'quote ,arg)))
+
+(example (m1 foo bar baz))  -- '(m1 foo bar baz)
+(example (m2 foo))          -- 'foo
+
+
```
gelisam, line 44: The following implementation is way more complicated than it needs to be. I wrote it first, and then I wrote `syntax-pmatch`. I then tried to rewrite `define-syntax-rule` in terms of `syntax-pmatch`, but I encountered #80 . So I'll leave it as-is for now.

gelisam:
One feature I would like to implement next is the `foo ...` pattern to match on a list of items. But in order to make that useful, we'd need a corrsponding `,(foo ...)` pattern in quasiquote, and I've never quite managed to understand how that function works.

gelisam:
hmm, did I do something dumb performance-wise? A few tests are running much slower now:

    define-syntax-rule: was 1.13s, now 7.87s
    which-problem: was 2.15s, now 17.42s
    double-define: was 1.48s, now 10.63s

david-christiansen:
Did you ever get some insight into the slowdowns? I suspect that it's a combination of the deep matching taking time to be expanded into large code, combined with said large code running slowly, but it would be interesting to know how much of it is our expander and how much is our interpreter.

PR #85, opened 2020-07-06
give every type an argument list

gelisam:
possibly an empty one. in practice, the type constructor determines the
length of that list, so TFun always has 2 arguments and TString always
has zero. thus, this more-uniform representation is less precise than
before, in that it makes more illegal states representable, but it paves
the way for the implementation of higher-kinded types. for example, it
is now possible to represent TFun applied to zero or one argument;
currently, writing either of those in a Klister program is a kind error,
but soon we can make those types of kind `* -> * -> *` and `* -> *`.

### DIFF
--- a/src/Evaluator.hs
+++ b/src/Evaluator.hs
@@ -311,20 +311,15 @@ doTypeCase blameLoc (Ty v0) ((p, rhs0) : ps) = match (doTypeCase blameLoc (Ty v0
     match :: Eval Value -> TypePattern -> Core -> TyF Ty -> Eval Value
     match next (TypePattern t) rhs scrut =
       case (t, scrut) of
-        (TSyntax, TSyntax) -> eval rhs
-        (TSignal, TSignal) -> eval rhs
-        (TFun (n1, x1) (n2, x2), TFun dom cod) ->
-           withExtendedEnv n1 x1 (ValueType dom) $
-           withExtendedEnv n2 x2 (ValueType cod) $
-           eval rhs
-        (TMacro (n, x), TMacro a) -> withExtendedEnv n x (ValueType a) (eval rhs)
-        (TType, TType) -> eval rhs
-        (TDatatype dt xs, TDatatype dt' args)
-          | dt == dt', length xs == length args ->
-            withManyExtendedEnv [(n, x, ValueType arg) | (n, x) <- xs, arg <- args] $
-            eval rhs
-        (TSchemaVar i, TSchemaVar j)
-          | i == j ->
+        -- unification variables never match; instead, type-case remains stuck
+        -- until the variable is unified with a concrete type constructor or a
+        -- skolem variable.
+        (TyF (TMetaVar _) _, _) -> next
+        (_, TyF (TMetaVar _) _) -> next
+
+        (TyF ctor1 args1, TyF ctor2 args2)
+          | ctor1 == ctor2 && length args1 == length args2 ->
+            withManyExtendedEnv [(n, x, ValueType arg) | (n, x) <- args1, arg <- args2] $
             eval rhs
         (_, _) -> next
     match _next (AnyType n x) rhs scrut =
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -382,12 +382,12 @@ initializeKernel = do
   where
     typePrims :: [(Text, (SplitTypePtr -> Syntax -> Expand (), TypePatternPtr -> Syntax -> Expand ()))]
     typePrims =
-      [ ("Syntax", Prims.baseType $ TyF TSyntax [])
-      , ("String", Prims.baseType $ TyF TString [])
-      , ("Signal", Prims.baseType $ TyF TSignal [])
+      [ ("Syntax", Prims.baseType tSyntax)
+      , ("String", Prims.baseType tString)
+      , ("Signal", Prims.baseType tSignal)
       , ("->", Prims.arrowType)
       , ("Macro", Prims.macroType)
-      , ("Type", Prims.baseType $ TyF TType [])
+      , ("Type", Prims.baseType tType)
       ]
 
     funPrims :: [(Text, Scheme Ty, Value)]
@@ -504,7 +504,7 @@ initializeKernel = do
                  then throwError $ WrongDatatypeArity stx dt arity (length args)
                  else do
                    argDests <- traverse scheduleType args
-                   linkType dest $ TyF (TDatatype dt) argDests
+                   linkType dest $ tDatatype dt argDests
           patImpl =
             \dest stx -> do
               Stx _ _ (me, args) <- mustBeCons stx
@@ -520,7 +520,7 @@ initializeKernel = do
                          | (sc, n, x) <- varInfo
                          ]
                   linkTypePattern dest $
-                    TypePattern $ TyF (TDatatype dt) [(varStx, var) | (_, varStx, var) <- varInfo]
+                    TypePattern $ tDatatype dt [(varStx, var) | (_, varStx, var) <- varInfo]
       let val = EPrimTypeMacro tyImpl patImpl
       b <- freshBinding
       bind b val
@@ -1053,7 +1053,7 @@ expandOneForm prob stx
             List xs -> expandOneExpression t dest (addApp List stx xs)
         ETypeVar i -> do
           dest <- requireTypeCtx stx prob
-          linkType dest $ TyF (TSchemaVar i) []
+          linkType dest $ tSchemaVar i
         EIncompleteDefn x n d -> do
           (t, dest) <- requireExpressionCtx stx prob
           forkAwaitingDefn x n b d t dest stx
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -393,7 +393,7 @@ initializeKernel = do
     funPrims :: [(Text, Scheme Ty, Value)]
     funPrims =
       [ ( "string=?"
-        , Scheme 0 (tString `tFun` tString `tFun` Prims.primitiveDatatype "Bool" [])
+        , Scheme 0 $ tFun [tString, tString] (Prims.primitiveDatatype "Bool" [])
         , ValueClosure $ HO $
           \(ValueString str1) ->
             ValueClosure $ HO $
@@ -403,7 +403,7 @@ initializeKernel = do
                 else primitiveCtor "false" []
         )
       , ( "string-append"
-        , Scheme 0 (tString `tFun` tString `tFun` tString)
+        , Scheme 0 $ tFun [tString, tString] tString
         , ValueClosure $ HO $
           \(ValueString str1) ->
             ValueClosure $ HO $
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -86,7 +86,7 @@ import qualified ScopeSet
 expandExpr :: Syntax -> Expand SplitCore
 expandExpr stx = do
   dest <- liftIO $ newSplitCorePtr
-  t <- Ty . TMetaVar <$> freshMeta
+  t <- tMetaVar <$> freshMeta
   forkExpandSyntax (ExprDest t dest) stx
   expandTasks
   children <- view expanderCompletedCore <$> getState
@@ -382,18 +382,18 @@ initializeKernel = do
   where
     typePrims :: [(Text, (SplitTypePtr -> Syntax -> Expand (), TypePatternPtr -> Syntax -> Expand ()))]
     typePrims =
-      [ ("Syntax", Prims.baseType TSyntax)
-      , ("String", Prims.baseType TString)
-      , ("Signal", Prims.baseType TSignal)
+      [ ("Syntax", Prims.baseType $ TyF TSyntax [])
+      , ("String", Prims.baseType $ TyF TString [])
+      , ("Signal", Prims.baseType $ TyF TSignal [])
       , ("->", Prims.arrowType)
       , ("Macro", Prims.macroType)
-      , ("Type", Prims.baseType TType)
+      , ("Type", Prims.baseType $ TyF TType [])
       ]
 
     funPrims :: [(Text, Scheme Ty, Value)]
     funPrims =
       [ ( "string=?"
-        , Scheme 0 $ Ty (TFun (Ty TString) (Ty (TFun (Ty TString) (Prims.primitiveDatatype "Bool" []))))
+        , Scheme 0 (tString `tFun` tString `tFun` Prims.primitiveDatatype "Bool" [])
         , ValueClosure $ HO $
           \(ValueString str1) ->
             ValueClosure $ HO $
@@ -403,7 +403,7 @@ initializeKernel = do
                 else primitiveCtor "false" []
         )
       , ( "string-append"
-        , Scheme 0 $ Ty (TFun (Ty TString) (Ty (TFun (Ty TString) (Ty TString))))
+        , Scheme 0 (tString `tFun` tString `tFun` tString)
         , ValueClosure $ HO $
           \(ValueString str1) ->
             ValueClosure $ HO $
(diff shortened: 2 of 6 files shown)


gelisam:
This is only a small step towards higher-kinded types, but since this change already touches so many lines of code, I think it's worth reviewing and merging it before moving on to the next step.

src/Evaluator.hs
```diff
@@ -311,20 +311,15 @@ doTypeCase blameLoc (Ty v0) ((p, rhs0) : ps) = match (doTypeCase blameLoc (Ty v0
     match :: Eval Value -> TypePattern -> Core -> TyF Ty -> Eval Value
     match next (TypePattern t) rhs scrut =
       case (t, scrut) of
-        (TSyntax, TSyntax) -> eval rhs
-        (TSignal, TSignal) -> eval rhs
```
gelisam, line 315: `(TString, TString)` was missing, so previously a type-case on `String` would mysteriously fail to match:

```
-- Internal error during expansion! This is a bug in the implementation.
-- Pattern info not added
(define-macro (m)
  (>>= (which-problem)
       (lambda (prob)
         (case prob
           [(expression t)
            (type-case t
              [String
               (pure '"foo")])]))))

(example (the String (m)))
```

With this PR it, all the type constructors are treated generically, so we can no longer forget to add a case... but a type-case on `String` still fails with the same error message. I am thus guessing that the error occurs earlier.

src/Expander.hs
```diff
@@ -513,14 +513,14 @@ initializeKernel = do
...
-                  sch <- trivialScheme $ Ty TType
+                  sch <- trivialScheme tType
                   modifyState $
                     set (expanderPatternBinders . at (Right dest)) $
                     Just [ (sc, n, x, sch)
                          | (sc, n, x) <- varInfo
                          ]
                   linkTypePattern dest $
-                    TypePattern $ TDatatype dt [(varStx, var) | (_, varStx, var) <- varInfo]
+                    TypePattern $ TyF (TDatatype dt) [(varStx, var) | (_, varStx, var) <- varInfo]
```
gelisam, line 523: The diffs in this file up to this point illustrate a few different situations in which `TyF` is used. Some functions ask for a `Ty`, which wraps a `TyF Ty`, while other functions as for a `TyF a` for an `a` which isn't `Ty`. Those were getting pretty long an unreadable, so I decided to define a bunch of helper functions, such as `tFun` and `tSyntax`, for constructing such values more easily.

I had a design choice: should I make those helpers for the `Ty` or the `TyF` variant? Making them for the `TyF` variant would improve all the call sites, at the cost of extra `Ty` wrappers in the places where `Ty` is needed. I decided instead fo make them for the `Ty` variant, because `Ty` expressions like the one on L396 are long and nested, and so the extra readability helps a lot more than in the `TyF` expressions, which tend to only be one level deep.

src/Expander.hs
```diff
@@ -382,18 +382,18 @@ initializeKernel = do
...
       , ("Macro", Prims.macroType)
-      , ("Type", Prims.baseType TType)
+      , ("Type", Prims.baseType $ TyF TType [])
       ]
 
     funPrims :: [(Text, Scheme Ty, Value)]
     funPrims =
       [ ( "string=?"
-        , Scheme 0 $ Ty (TFun (Ty TString) (Ty (TFun (Ty TString) (Prims.primitiveDatatype "Bool" []))))
+        , Scheme 0 (tString `tFun` tString `tFun` Prims.primitiveDatatype "Bool" [])
```
gelisam, line 396: note: infix `tFun` is right-associative, like `(->)`.

src/Expander/TC.hs
```diff
@@ -36,36 +36,34 @@ setTVLevel ptr l = do
...
 normType t = return t
 
 normAll :: Ty -> Expand Ty
-normAll t =
-  normType t >>= fmap Ty .
-  \case
-    (Ty (TFun a b)) -> TFun <$> normType a <*> normType b
-    (Ty (TMacro a)) -> TMacro <$> normType a
-    (Ty (TDatatype dt tArgs)) ->
-      TDatatype dt <$> traverse normType tArgs
```
gelisam, line 57: This time the omission of `TString` was intentional, as it has no recursive positions on which to call `normType`. The new code handles such argument-less type constructors just fine, and makes it impossible to forget to add new type constructors which do have recursive positions.

src/Expander/TC.hs
```diff
@@ -36,36 +36,34 @@ setTVLevel ptr l = do
...
+  Ty <$> TyF ctor <$> traverse normType tArgs
 
 metas :: Ty -> Expand [MetaPtr]
-metas t =
-  normType t >>=
-  \case
-    Ty (TMetaVar x) -> pure [x]
-    Ty (TFun a b) -> (++) <$> metas a <*> metas b
-    Ty (TMacro a) -> metas a
-    Ty (TDatatype _ ts) -> concat <$> traverse metas ts
```
gelisam, line 67: same here.

src/Expander/TC.hs
```diff
@@ -102,22 +100,29 @@ inst :: Scheme Ty -> [Ty] -> Expand Ty
...
+    instTyF :: TyF Ty -> Expand (TyF Ty)
+    instTyF (TyF ctor tArgs) = do
+      let TyF ctor' tArgsPrefix = instCtor ctor
+      tArgsSuffix <- traverse instTy tArgs
+      pure $ TyF ctor' (tArgsPrefix ++ tArgsSuffix)
 
-    inst' (TFun a b) = TFun <$> instNorm a <*> instNorm b
-    inst' (TMacro a) = TMacro <$> instNorm a
-    inst' (TDatatype dt tArgs) = TDatatype dt <$> traverse instNorm tArgs
-    inst' (TSchemaVar i) = pure . unTy $ ts !! fromIntegral i
```
gelisam, line 114: and here.

src/Expander/TC.hs
```diff
@@ -102,22 +100,29 @@ inst :: Scheme Ty -> [Ty] -> Expand Ty
...
+    instTy t = do
       t' <- normType t
-      Ty <$> inst' (unTy t')
+      Ty <$> instTyF (unTy t')
+
+    instTyF :: TyF Ty -> Expand (TyF Ty)
+    instTyF (TyF ctor tArgs) = do
+      let TyF ctor' tArgsPrefix = instCtor ctor
+      tArgsSuffix <- traverse instTy tArgs
+      pure $ TyF ctor' (tArgsPrefix ++ tArgsSuffix)
```
gelisam, line 114: The fact that we have to concatenate two lists is new.

Previously, the types were guaranteeing that a `TSchemaVar` is never applied to type arguments. So when we encountered a `TSchemaVar`, that was it, we returned `ts !! fromIntegral i` and we did not have any type arguments to recur on.

But now, according to the types, a `TSchemaVar` can have type arguments. So when `instCtor` encounters a `TSchemaVar`, it once again returns `ts !! fromIntegral` as before, and that might be e.g. a `TFun` with two type arguments. Or maybe `instCtor` encounters a `TFun`, and the two type arguments are instead in `tArgs` (`tArgsSuffix` after they are instantiated). So we have two sources of type arguments which we need to combine into one.

In practice, a `TSchemaVar` will not have type arguments, because higher-kinded types are not yet supported. So one of the two lists will be empty, and `++` is a complicated way to pick the non-empty one. Thinking a bit ahead, however, once we do support higher-kinded types, we might have `f Int` with `f` expanding to `Either String`; in this case, we want to concatenate `[String]` with `[Int]` to get `[String, Int]`, that is, `Either String Int`. So while it currently doesn't matter whether we write `tArgsPrefix ++ tArgsSuffix` or `tArgsSuffix ++ tArgsPrefix`, once we have higher-kinded types, we will want `tArgsSuffix ++ tArgsPrefix`, so we might as well use that now.

src/Expander/TC.hs
```diff
@@ -102,22 +100,27 @@ inst :: Scheme Ty -> [Ty] -> Expand Ty
...
-    inst' otherTy = pure otherTy
+    instCtor :: TypeConstructor -> TyF Ty
+    instCtor (TSchemaVar i) = unTy $ ts !! fromIntegral i
+    instCtor ctor           = TyF ctor []
 
 
 specialize :: Scheme Ty -> Expand Ty
 specialize sch@(Scheme n _) = do
-  freshVars <- replicateM (fromIntegral n) $ Ty . TMetaVar <$> freshMeta
+  freshVars <- makeTypeMetas n
```
gelisam, line 123: there was already a function, `makeTypeMetas`, which did exactly the same thing.

src/Evaluator.hs
david-christiansen: I suspect without having looked that the `String` builtin doesn't expand to the right thing in a type-pattern context, if this is what you're seeing. In any case, making it generic is definitely an improvement!

src/Expander.hs
david-christiansen: Sounds good.

What about making the helpers into a type class that can have instances for both cases?

src/Expander.hs
david-christiansen: Ak! Scary! Fixity declarations for backticked names is a fairly obscure Haskell feature, isn't it?

I'd probably have made it be `[Ty] -> Ty -> Ty` instead, to remove all doubt at call sites, but this does work.

src/Expander/TC.hs
david-christiansen: Great!

src/Expander/TC.hs
david-christiansen: This makes sense. But why is the comment here, rather than in the source file?

src/Expander/TC.hs
david-christiansen: :+1:

src/Evaluator.hs
```diff
@@ -311,20 +311,15 @@ doTypeCase blameLoc (Ty v0) ((p, rhs0) : ps) = match (doTypeCase blameLoc (Ty v0
...
-        (TSchemaVar i, TSchemaVar j)
-          | i == j ->
+        -- unification variables never match; instead, type-case remains stuck
+        -- until the variable is unified with a concrete type constructor or a
+        -- skolem variable.
+        (TyF (TMetaVar _) _, _) -> next
+        (_, TyF (TMetaVar _) _) -> next
+
+        (TyF ctor1 args1, TyF ctor2 args2)
+          | ctor1 == ctor2 && length args1 == length args2 ->
```
david-christiansen, line 321: This is an interesting design choice when it comes to later on. Do we allow matching at kinds other than `*`? In that case, do we statically rule out ill-kinded typecase operations, just as we type-check datatype patterns?

In any case, the `length` calls are redundant here unless we don't kind-check at all, right?

src/Expander.hs
david-christiansen: E.g. `tFun [tString, tString] (tPrimData "Bool" [])`

src/Expander/TC.hs
```diff
@@ -177,13 +176,14 @@ generalizeType ty = do
...
+    genVarsCtor _ ctor =
+      pure ctor
 
 
 makeTypeMetas :: Natural -> Expand [Ty]
-makeTypeMetas 0 =
-  pure []
-makeTypeMetas n =
-  (:) <$> (Ty . TMetaVar <$> freshMeta) <*> makeTypeMetas (n - 1)
+makeTypeMetas n = replicateM (fromIntegral n) $ tMetaVar <$> freshMeta
```
david-christiansen, line 187: The `replicateM` here is much nicer. Thanks!

src/Type.hs
```diff
@@ -25,18 +25,25 @@ newMetaPtr = MetaPtr <$> newUnique
 instance Show MetaPtr where
   show (MetaPtr i) = "(MetaPtr " ++ show (hashUnique i) ++ ")"
 
-data TyF t
+data TypeConstructor
```
david-christiansen, line 28: If we wanted to, we could index `TypeConstructor` by the number of type arguments, and get some more static safety for the built-in ones. I don't think it's worth it, but it is definitely a possibility.

And we could do something like this:
```
{-# LANGUAGE DataKinds #-}

data ArgCount = None | One | Two | Many
data TypeConstructor :: ArgCount -> Type where
  TSyntax :: ArgCount None
  -- ...
  TFun :: ArgCount Two
  -- ...
  TDatatype :: ArgCount Many

data TyApp =
  App0 (TypeConstructor None) | App1 (TypeConstructor One) (TyApp) | App2 (TypeConstructor Two) TyApp TyApp | AppMany (TypeConstructor Many) [TyApp]
```
This basically treats TypeConstructor as one type per number of args, with the advantage of being able to create a little existential package when we want to talk about them generically rather than having to build the explicit sum type (basically using the iso between Either A B and (Sigma b : Bool . if b then A else B).

This is probably not worth the added complication, but you had mentioned being a bit unhappy with the lack of static safety so I though it was worth mentioning.

src/Type.hs
```diff
@@ -25,18 +25,25 @@ newMetaPtr = MetaPtr <$> newUnique
...
+  | TMacro
   | TType
-  | TDatatype Datatype [t]
+  | TDatatype Datatype
   | TSchemaVar Natural
   | TMetaVar MetaPtr
+  deriving (Eq, Show)
+makePrisms ''TypeConstructor
+
+data TyF t = TyF
```
david-christiansen, line 41: Why is this called TyF now? The name before was because it was the pattern functor in a typical fixed point construction, but here it feels like the more relevant concept to readers of the code is that it's an application of a type constructor to some arguments. It's certainly also a pattern functor - but it seems that we now have a less-generic name available than we had before.

david-christiansen: (Purple bike shed please!)

src/Type.hs
```diff
@@ -73,23 +80,39 @@ newtype Ty = Ty
   deriving (Eq, Show)
 makePrisms ''Ty
 
+tSyntax :: Ty
```
david-christiansen, line 83: As I mentioned elsewhere, this could perhaps be easier to use by creating a type class and a couple of bidirectional pattern synonyms, right? Then users could basically ignore the newtype constructor.

src/Type.hs
```diff
@@ -73,23 +80,39 @@ newtype Ty = Ty
...
-  alphaCheck TSyntax TSyntax = pure ()
-  alphaCheck TSignal TSignal = pure ()
-  alphaCheck (TFun a1 b1) (TFun a2 b2) = do
-    alphaCheck a1 a2
-    alphaCheck b1 b2
-  alphaCheck (TMacro a) (TMacro b) =
-    alphaCheck a b
-  alphaCheck (TDatatype a args1) (TDatatype b args2) = do
-    guard (a == b)
+  alphaCheck (TyF ctor1 args1) (TyF ctor2 args2) = do
```
david-christiansen, line 112: I love short code. Especially code that lacks bugs like "forgot to compare TString"!

src/Expander.hs
gelisam: good idea! pushed.

src/Expander.hs
gelisam: Sure, if you prefer. That doesn't work with `TyF`, so I've changed the method from `tFun :: arg -> arg -> a` to `tFun1 :: arg -> arg -> a`, and I have defined `tFun :: [Ty] -> Ty -> Ty` in terms of it.

src/Expander/TC.hs
gelisam: because I think comments which explain why a _change_ was made belong either in commit messages or PR comments. I can turn it into a comment which explains why the code _does_ what it does if you like.

gelisam: added.

david-christiansen: Looks good!

PR #86, opened 2020-07-27
use [Kind] instead of a Natural arity whenever possible

gelisam:
another small step towards higher-kinded types


src/Expander/Primitives.hs
```diff
@@ -5,6 +5,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS -Wno-name-shadowing #-}
```
gelisam, line 8: There was a warning about the `typeArgs` variable on line 129, which shadows some `typeArgs` global. I don't find name-shadowing warnings to be particularly useful, so I decided to hush disable them instead of fixing them.

src/Expander.hs
```diff
@@ -287,7 +288,10 @@ evalMod (Expanded em _) = execWriterT $ do
             Map.insert cn (ConstructorInfo argTypes dt)
           lift $ modifyState $
             over (expanderWorld . worldDatatypes . at p . non Map.empty) $
-            Map.insert dt (DatatypeInfo arity [c | (_, c, _) <- ctors ])
+            Map.insert dt $ DatatypeInfo
+              { _datatypeArgKinds     = replicate (fromIntegral arity) KStar
```
gelisam, line 292: this `arity` comes from the `Data` constructor above; I guess the next step will be to push the `[Kind]` representation a bit further and use it in `Data` as well.

src/Expander.hs
```diff
@@ -543,7 +547,7 @@ initializeKernel = do
 
       let info =
             DatatypeInfo
-            { _datatypeArity = arity
+            { _datatypeArgKinds = replicate (fromIntegral arity) KStar
```
gelisam, line 550: this `arity` is an argument of `addDatatypePrimitive`; I guess another next step will be to push the `[Kind]` representation a bit further and use at the `addDatatypePrimitive` call sites as well.

src/Expander/Primitives.hs
```diff
@@ -146,7 +148,7 @@ datatype dest outScopesDest stx = do
 
   let info =
         DatatypeInfo
-        { _datatypeArity = fromIntegral arity
+        { _datatypeArgKinds = replicate arity KStar
```
gelisam, line 151: the `arity` comes from the number of type parameters given when declaring the datatype, e.g. `(datatype (Either A B) ...` generates a datatype `Either` with arity 2 and thus of kind `* -> * -> *`. That's Haskell's behaviour when `PolyKinds` is off, so I'm thinking of leaving it as-is. One next step would be to allow kind annotations, e.g. `(datatype (StateT (the * s) (the (-> * *) m) (the * a)) ...`.

src/Expander/TC.hs
```diff
@@ -1,7 +1,8 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE ViewPatterns #-}
-module Expander.TC (unify, freshMeta, inst, specialize, varType, makeTypeMetas, generalizeType, normType) where
+{-# OPTIONS -Wno-unused-top-binds #-}
```
gelisam, line 4: There was a warning saying that `setTVLevel` is dead code because it is not used inside the file but also not exported. Because of its symmetry with `setTVKind` (now `setTVLinkage`), it seemed like a plausibly-useful function, so I decided to leave it in and to disable the warning.

src/Expander/TC.hs
```diff
@@ -26,10 +28,10 @@ derefType ptr =
     Just var -> pure var
 
 
-setTVKind :: MetaPtr -> VarKind Ty -> Expand ()
-setTVKind ptr k = do
+setTVLinkage :: MetaPtr -> VarLinkage Ty -> Expand ()
+setTVLinkage ptr k = do
```
gelisam, line 32: renaming "TVKind" to "TVLinkage" in order to avoid confusion between "kind" meaning "is this a type variable which is linked to another, known type expression, or is this an unlinked type variable about which we know nothing?" and "kind" meaning "the type of types".

src/Expander/TC.hs
```diff
@@ -97,8 +99,8 @@ freshMeta = do
   return ptr
 
 inst :: Scheme Ty -> [Ty] -> Expand Ty
-inst (Scheme n ty) ts
-  | length ts /= fromIntegral n =
+inst (Scheme argKinds ty) ts
+  | length ts /= length argKinds =
```
gelisam, line 103: currently, type arguments always have kind `*`, so it is not necessary to check that each element of `ts` has the right kind.

src/Expander.hs
```diff
@@ -988,12 +992,12 @@ expandOneForm prob stx
           saveExprType dest t
         EConstructor ctor -> do
           ConstructorInfo args dt <- constructorInfo ctor
-          DatatypeInfo arity _ <- datatypeInfo dt
+          DatatypeInfo argKinds _ <- datatypeInfo dt
           case prob of
             ExprDest t dest -> do
-              argTys <- makeTypeMetas arity
+              argTys <- traverse makeTypeMeta argKinds
```
gelisam, line 998: I am passing the kind to `makeTypeMeta`, but it currently ignores it; eventually, we'll want to associate a kind to each type variable (and perhaps even to infer their kinds).

src/Expander/TC.hs
```diff
@@ -149,34 +151,34 @@ notFreeInCtx var = do
...
-      StateT (Natural, Map MetaPtr Natural) Expand TypeConstructor
+      StateT (Natural, Map MetaPtr Natural, [Kind]) Expand TypeConstructor
     genVarsCtor vars (TMetaVar v)
       | v `elem` vars = do
-          (i, indices) <- get
+          (i, indices, argKinds) <- get
           case Map.lookup v indices of
             Nothing -> do
-              put (i + 1, Map.insert v i indices)
+              put (i + 1, Map.insert v i indices, argKinds ++ [KStar])
```
gelisam, line 181: I could get rid of the `Natural`, it can be recovered from the length of the list.

gelisam: I guess I should use a `Seq` instead of a list in order to avoid quadratic performance here...

src/Kind.hs
```diff
@@ -0,0 +1,10 @@
+module Kind where
+
+
+data Kind
+  = KStar
+  | KFun Kind Kind
+  deriving (Show, Eq)
```
gelisam, line 7: if we do kind inference, eventually this will have to be split into a non-recursive type, like everything else.

tests/Golden.hs
```diff
@@ -13,7 +13,7 @@ import Data.Foldable (for_)
 import Data.Text.Lazy (Text)
```
gelisam, line 13: I ran the tests, expecting to only see small differences due to the pretty-printing now using `∀(α : *)` instead of `∀α`, but the test output was so unreadable that I couldn't even tell if those were the differences or not! So I spent a bit of time improving the output format in the failure case.

gelisam:
I guess launchpad.net is having issues? The build fails with

```
Get:2 http://ppa.launchpad.net/hvr/ghc/ubuntu xenial/main amd64 ghc-8.8.1 amd64 8.8.1-12~16.04 [89.3 MB]

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I guess we should simply try again later.

src/Expander.hs
gelisam: done.

src/Expander.hs
gelisam: done.

src/Expander/Primitives.hs
david-christiansen: I wish there was a warning for shadowing names with compatible types only. In the absence of that, I think this is reasonable.

src/Expander/Primitives.hs
david-christiansen: Ideally we'd have kind inference here rather than requiring the annotation. Annotations are nice, but not really in the spirit of the ML-Haskelly language we're building. In any case, `the` should annotate use sites rather than binding sites. What about
`(datatype (StateT s (m (-> * *)) a) ...)` where no annotation means `*` if we absolutely need annotations?

src/Expander/TC.hs
david-christiansen: Good idea

src/Expander/TC.hs
david-christiansen: This seems like the kind of thing that will be easy to forget later...

src/Expander/TC.hs
david-christiansen: I suspect that the list just won't be a problem here. A quadradic overhead on a list of at most 10 elements is still tiny.

src/Kind.hs
david-christiansen: Indeed!

david-christiansen:
This is still a working draft, right?

src/Expander/Primitives.hs
gelisam: In Haskell, I often need to explicitly give kind annotations, because its kind inference defaults to `*` when there is nothing in the type definition which constrains the kind, even when there is some function using that type later in the file which constraints the kind. so it that's also the behaviour we want for kind inference, I do think we need to allow for optional kind annotations.

gelisam:
> This is still a working draft, right?

Yes and no; I am obviously not done with type inference, but personally I like to keep PRs small instead of accumulating commit after commit in a giant feature branch. It makes a bigger difference when reviewing than when creating PRs, though, so I'll let you decide whether to merge what I have so far or to keep this PR open until I'm done with type inference.

src/Expander.hs
david-christiansen: Yeah, I suspect we'll actually want to associate a fresh kind variable with each type variable once we make some more progress.

tests/Golden.hs
david-christiansen: That's wonderful! Thank you!

src/Pretty.hs
```diff
@@ -292,16 +293,22 @@ instance PrettyBinder VarInfo [CompleteDecl] where
...
+  pp env (Scheme argKinds t) =
     text "∀" <>
     (align $ group $
      vsep [ group $
-            vsep (map text (take (fromIntegral n) typeVarNames)) <> text "."
+            vsep (zipWith ppArgKind typeVarNames argKinds) <> text "."
           , pp env t
           ])
+    where
+      ppArgKind varName kind = parens (text varName <+> text ":" <+> pp env kind)
```
david-christiansen, line 311: I wonder if this wouldn't be more readable if we only showed non-`*` kinds.

src/Expander/TC.hs
david-christiansen: Fixed in #92 .

src/Kind.hs
david-christiansen: Implemented in #92.

david-christiansen:
The merge conflicts for this branch are fixed in #92, which is built on it.

gelisam:
Closing in favour of #92, which fixes the conflicts.

Issue #55, opened 2020-05-02
License file


langston-barrett:
We need a license file, or our code may not be effectively open source! From http://choosealicense.com/no-license:

> You’ll have to check with your own legal counsel regarding your particular project, but generally speaking, the absence of a license means that default copyright laws apply. This means that you retain all rights to your source code and that nobody else may reproduce, distribute, or create derivative works from your work.

FWIW, I really don't care what license we have.

gelisam:
I don't care either. I usually put "PublicDomain".

david-christiansen:
Public Domain is tricky in some jurisdictions, like Germany, so I'm not a fan of that. I think that the goals of public domain are better met with CC0.

Strategically, I like making infrastructure GPL because it results in more users (rather than corporations) controlling their computers. But this language implementation is more of a research prototype that we want to get adopted and read, so how about MIT? 

We should get @xplat's permission for that too.

langston-barrett:
> so how about MIT?

Sure!

gelisam:
> so how about MIT?

sure, I'm fine with anything. Thanks for the CC0 suggestion, I did hear that public domain had issues but I didn't take the time to research the alternatives.

langston-barrett:
Fixed in #88

PR #88, opened 2020-08-11
Add BSD license

david-christiansen:
I talked to the legal team at Galois and BSD is preferred. This work for you all?


langston-barrett:
Fixes #55

Issue #79, opened 2020-05-23
Integers, not signals


gelisam:
Now that we have type-aware macros, the signals we were playing with at the beginning as a proof of concept have outlived their usefulness. Let's drop the signalling machinery, but keep the parser and treat them as proper integers. A few primitive operations like `+`, `-`, `*`, and `integer=?` would be nice.

david-christiansen:
Here's how I think the design should work:
 - As you said, have a primitive `Integer` type, with primitive functions reflected from Haskell's `Num`, `Eq`, and `Ord`.
 - When the expander encounters an integer literal `i`, it should expand to `(#%int-literal i)`. The kernel language's `#%int-literal` will then do the right thing, including the type checking. This allows DSLs do do things like changing the meanings of integer literals, similarly to how they do it today with `#%app`. The `#%int-literal` should be added such that macros can expand to a `let-syntax` that binds it to something that's locally interesting.
 - Our syntax pattern language will need a case for int literals as well.

david-christiansen:
This has mostly been done as of #89 and #99 . The only thing missing is the `#%int-literal` hook.

gelisam:
Since we're usually doing whatever Racket does, what's the motivation for straying and using `#%int-literal` (and presumably `#%string-literal`) instead of `#%datum`?

Personally, I like `#%int-literal` better than `#%datum`. One of my plans for Klister is to eventually support macros with more precise types than `Syntax -> Macro Syntax`, and at that point `#%int-literal`'s type can be specialized to `Int -> Macro Syntax`.

david-christiansen:
I think that there's a good reason for doing this here.

Racket, being dynamically-typed, mostly just has one thing to do with literals: wrap them in a `quote`. When designing a DSL, one might want to additionally have `#%datum` check for more specfics and do something different, but that's more of a corner case. In Klister, however, we will need something like a sum type for literals with a case for each type if we want to give them a uniform expander hook like `#%datum`. Might as well just give them each their own little expander hook and avoid the sum type, I'd think.

PR #89, opened 2020-08-15
Remove signals, replace with integer literals

langston-barrett:
Work towards #79


src/LiteralInteger.hs
```diff
@@ -0,0 +1,24 @@
...
+import Control.Monad
+
+import Alpha
+import ShortShow
+
+
+-- | A newtype wrapping user-supplied literal integers.
+--
+-- Useful to have a newtype in case we allow, e.g., integer literals in another
+-- base someday.
```
langston-barrett, line 14: Do you buy this?

david-christiansen: Would the idea be to replace the newtype with a sum type, with a constructor for each base, so that we can print syntax objects as the user wrote them? Seems fine enough, but also like something we don't need to really plan a lot for.

src/Syntax.hs
```diff
@@ -29,7 +29,7 @@ makeLenses ''Stx
 data ExprF a
   = Id Text
   | String Text
-  | Sig Signal
+  | Lit LiteralInteger
```
david-christiansen, line 32: I'm not a big fan of this constructor name, because strings or other bits we might want can also be fairly characterized as  "literals". I found it a bit confusing at first myself. What about `Int` or `Integer` instead?

src/LiteralInteger.hs
langston-barrett: Removed

src/Expander.hs
```diff
@@ -1146,17 +1131,16 @@ expandDeclForms _dest _scs _outScopesDest _stx =
...
 expandLiteralString dest str = do
   linkExpr dest (CoreString str)
 
 data MacroOutput
-  = StuckOnSignal Signal [Closure]
-  | StuckOnType SrcLoc Ty VEnv [(TypePattern, Core)] [Closure]
+  = StuckOnType SrcLoc Ty VEnv [(TypePattern, Core)] [Closure]
   | Done Value
 
 -- If we're stuck waiting on a signal, we return that signal and a continuation
```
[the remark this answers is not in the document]
langston-barrett: Updated

src/Syntax.hs
langston-barrett: Changed to `LitInt` to avoid confusion with Haskell's `Integer`, and because it's a tad more descriptive.

david-christiansen:
LGTM. Thanks!

src/Syntax.hs
```diff
@@ -31,7 +30,7 @@ makeLenses ''Stx
 data ExprF a
   = Id Text
   | String Text
-  | Sig Signal
+  | LitInt Integer
```
gelisam, line 33: For consistency, I think we should either use `String` and `Int` or `LitString` and `LitInt`.

langston-barrett: I'd be fine with tackling that in a follow-up. I chose `LitInt` to avoid clashing with Haskell's built-in `Integer`.

gelisam: What was the clash? `String` is, of course, also a built-in type, but it's a type, so it doesn't conflict with `ExprF`'s `String` data constructor. I would have expected the same to happen with `Integer`.

langston-barrett: @gelisam Yep, I just have a stylistic preference to avoid reusing common built-in names, because it can be confusing.

PR #91, opened 2020-08-16
Module comments for Core, SplitCore, Syntax

langston-barrett:
Let me know if I've characterized these well!

### DIFF
--- a/src/Core.hs
+++ b/src/Core.hs
@@ -1,3 +1,12 @@
+{-|
+Module           : Core
+Description      : An extensible datatype for expressions
+
+The various instantiations of the openly recursive 'CoreF' datatype represent
+expressions in different states of expansion, with the fixpoint 'Core'
+representing fully-expanded expressions.
+-}
+
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveFunctor #-}
@@ -7,6 +16,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
+
 module Core where
 
 import Control.Lens hiding (elements)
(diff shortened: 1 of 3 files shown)


src/Core.hs
```diff
@@ -1,3 +1,12 @@
+{-|
+Module           : Core
+Description      : An extensible datatype for expressions
+
+The various instantiations of the openly recursive 'CoreF' datatype represent
+expressions in different states of expansion, with 'Core' representing
```
david-christiansen, line 6: As a minor quibble, I might write "with the fixed point 'Core' representing fully-expanded expressions.'

src/Core.hs
```diff
@@ -286,6 +296,7 @@ instance (Phased typePat, Phased pat, Phased core) => Phased (CoreF typePat pat
   shift i (CoreSyntax stx) = CoreSyntax (shift i stx)
   shift i other = bimap (shift i) (shift i) other
 
+-- | A fully expanded expression, ready to be evaluated.
```
david-christiansen, line 299: I think "fully-expanded" should have the hyphen, as above.

src/SplitCore.hs
```diff
@@ -1,8 +1,22 @@
+{-|
+Module           : SplitCore
+Description      : An intermediate datatype for partially expanded expressions.
+
+'SplitCore' is an intermediary between 'Syntax' (what the user types) and 'Core'
```
david-christiansen, line 5: I wouldn't characterize it as an intermediary - it's got basically nothing to do with `Syntax`. I'd instead characterize it as a partially-constructed AST in which zero or more sub-trees may be absent that's represented as an explicit pointer graph in the `Map` component.

src/Syntax.hs
```diff
@@ -1,8 +1,17 @@
+{-|
+Module           : Syntax
+Description      : User-facing syntax of Klister
+
+'Syntax' is the result of parsing Klister code. It is transformed into
+Klister\'s core language by the expander.
```
david-christiansen, line 6: This is true, but incomplete. Some Syntax also results from the expansion of user macros.

I'd instead say that this is the user-facing syntax for Klister, and that it can result either from parsing user code or from steps of macro expansion.

langston-barrett:
Thanks for the review and clarification @david-christiansen!

PR #92, opened 2020-08-23
Kind inference

david-christiansen:
Based on the work in #86, I hacked up some kind inference and higher-kinded types when I had insomnia. It still needs some careful attention to make sure kinds are checked where they need to be, as well as some better tests and exampes, but I figured I'd get the approach in here at least for comment.


david-christiansen:
Assuming I didn't mess up the tests, I think this is good to go now.

gelisam:
I tried resolving the conflicts using github's in-browser merge tools. Resolving conflicts without a typechecker felt like doing trapeze without a net! I'll definitely be using the more traditional out-of-browser approach next time.

examples/fix.kl
```diff
@@ -0,0 +1,43 @@
...
+
+(define length
+  (flet (length (xs)
+          (case (out xs)
+            [(nil) (in (zero))]
+            [(cons y ys) (in (succ (length ys)))]))
+    length))
+
+
+(define l1 (the (Fix (ListF String)) (in (cons "hello" (in (cons "world" (in (nil))))))))
```
gelisam, line 27: does the `the` imply that our type inference algorithm unable to infer `l1`'s type?

gelisam:
Urg, the "trapeze without a net" conflict resolution I did was completely bogus. Do over! With a net this time!

david-christiansen:
I did it over!

examples/fix.kl
david-christiansen: It does not - that was there for readability. I just pushed a version without it to make sure it keeps working.

gelisam:
fixed the conflicts.

examples/monad.kl
```diff
@@ -0,0 +1,231 @@
...
+(import "lambda-case.kl")
+(import (shift "prelude.kl" 1))
+(import (shift "let.kl" 1))
+(import (shift "quasiquote.kl" 1))
+
+
+(example (lambda (x) x))
+
+(datatype (Functor F A B)
+  (functor (-> (-> A B) (-> (F A) (F B)))))
```
gelisam, line 13: Hmm, that's not quite the definition I expected! I am guessing that we cannot yet have a field of type `forall a b. something`?

david-christiansen:
That's right. But every occurrence of "functor" can freely instantiate the
type variables, so it all ends up working out. Manual dictionary encodings
require either this hack or higher-rank types (and I heard from a little
bird that this was what led to some of the early work on rank 2
polymorphism in Haskell).

examples/monad.kl
```diff
@@ -0,0 +1,231 @@
...
+(example (perhaps string-append (just "applicative") (nothing)))
+
+
+(datatype (Monad M A B)
+  (monad (Applicative M A B)
+         (-> (M A) (-> A (M B))
+             (M B))))
+
+(define macro-monad
+  (monad macro-applicative (lambda (m f) (>>= m f))))
```
gelisam, line 146: I see that the `lambda` is necessary because `bindMacro` is a macro, not a function. I don't remember why we defined it this way, is there a good reason?

david-christiansen:
Because we didn't support primitive values back then. It could be one now,
I think...

gelisam:
hmm, looks like the macOS test is flaky. merging anyway, I'll debug that separately.

PR #98, opened 2020-08-26
stdlib: Add quasiquote.kl, list.kl

langston-barrett:
Progress on #54

### DIFF
--- a/examples/import-list-and-do.kl
+++ b/examples/import-list-and-do.kl
@@ -1,6 +1,6 @@
 #lang "prelude.kl"
 
-(import "list-datatype.kl")
+(import "list.kl")
 
 (import "lispy-do.kl")
 
--- a/examples/list-datatype.golden
+++ b/examples/list-datatype.golden
@@ -1,0 +1,1 @@
+(:: Signal 3 (:: Signal 2 (:: Signal 1 (nil)))) : (List Signal)
--- a/examples/list-datatype.kl
+++ b/examples/list-datatype.kl
@@ -1,22 +1,5 @@
 #lang "prelude.kl"
 
-(import "defun.kl")
+(import "list.kl")
 
-(datatype (List A)
-  (nil)
-  (:: A (List A)))
-
-(defun map (f lst)
-  (case lst
-    [(:: x xs) (:: (f x) (map f xs))]
-    [(nil) (nil)]))
-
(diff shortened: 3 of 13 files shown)


langston-barrett:
I'm not entirely sure what's up with the remaining failure:
```
    bool:                                              Expected 3 entries between parentheses, but got
  #[bool.kl:26.24-27.68]<(#%app binary-and (true) (and (true) (true)))>
FAIL (0.04s)
      user error
```

david-christiansen:
Probably it's using n-ary function application in a file where the language doesn't have that.

david-christiansen:
Looking at this a little more, it looks like it's picking the wrong `#%app` in the expansion of the `and` macro. 

When I switch line 26-27 to:
```
                (pure (list-syntax ('binary-and x
                                               (cons-list-syntax 'and xs stx)) stx))))))))

```
it works. This makes me think that it's a bug in the quasiquote macro, where it's picking the wrong lexical context, but I don't know for sure.

david-christiansen:
Yep, that was it. This fixes it:
```
modified   stdlib/quasiquote.kl
@@ -49,14 +49,14 @@
                                                       (list-syntax ('quasiquote s)
                                                                    s))
                                                     e)
-                                        (list-syntax ('quote (replace-loc e 'foo)) 'foo))
+                                        (list-syntax ('quote e) e))
                                       e)]]]]]
               [_ [pure
                   (list-syntax ('list-syntax
                                   (map-syntax (lambda (s)
                                                 (list-syntax ('quasiquote s) s))
                                               e)
-                                  (list-syntax ('quote (replace-loc e 'foo)) 'foo))
+                                  (list-syntax ('quote e) e))
                                 e)]])])]
         [_ (syntax-error [list-syntax [[quote "bad syntax"] stx] stx] stx)])]])]
 
```

There's no way to send a PR to a PR as far as I can see, unfortunately.

langston-barrett:
@david-christiansen Thanks for taking a detailed look! I'll add that to this branch soonish.

david-christiansen:
It looks to me like a golden file needs updating.

david-christiansen:
Ah, the golden file is right for your branch, but not for the result of merging in `master`, where `Signal` is no more. What about merging master and then running CI again?

PR #99, opened 2020-08-26
Add integer primitive operations

david-christiansen:
This makes integers more useful, now that we have them as of #89.


src/Expander.hs
```diff
@@ -411,8 +411,34 @@ initializeKernel = do
             \(ValueString str2) ->
               ValueString (str1 <> str2)
         )
+      , ( "integer->string"
```
gelisam, line 414: `int-ops` tests everything but `integer->string`, is that intentional?

stdlib/prelude.kl
```diff
@@ -11,7 +11,7 @@
 (export -- primitive types
         Syntax
         ->
-        Integer
+        Integer + - * / abs negate > >= < <= = /=
```
gelisam, line 14: `integer->string` is again omitted. probably a mistake?

src/Expander.hs
david-christiansen: Nope, that was an oversight. Thanks!

stdlib/prelude.kl
david-christiansen: Yep :)

langston-barrett:
Looks great, thanks @david-christiansen!

PR #104, opened 2020-08-28
Basic support for an IO monad

david-christiansen:
This PR adds support for an IO monad. It will need updating once #92 is merged, but nothing major (mostly just extending the kind checker to cover the new cases, and replacing `Naturals` that give type scheme arities with `[Kind]`).

# Design choices

I had to make a few design choices for the language and the implementation, and I figure these are worth calling out explicitly for feedback.

### IO monad
Effects are monadic, as in Haskell. So far, I've only implemented one effect: printing output.

### Ports a la Scheme
Rather than exposing Unix file handles directly as Haskell does, I've implemented a port abstraction a la Scheme. Ports are basically just generalizations of handles, and they give us a hook on which to implement things like handles that point at in-memory strings rather than having to round-trip through a temp file.

### No `main`

Instead of a C- or Haskell-like designated entry point named `main`, there's a `run` declaration form that causes an IO action to be executed in the same contexts that an example gets displayed (that is, when testing or running the module that it's in). This allows programmers a bit more freedom about code structure, and it frees us from worrying about how to plumb `main` through all the hygiene machinery.

### Handle everything at the value level

CoreF is changed here, but only to clarify some naming. Instead of adding primitive syntax for IO ops the way we have it for macro ops, I did everything, including `bind` and `pure`, as values. There's a value constructor for IO actions, which just directly embeds a Haskell `IO` action. I think this is a win, due to much shorter code and not requiring users to eta-expand things that they want to use as functions, but it's a bit inconsistent. I think it's worth it, though.

### Testing: golden tests run IO actions
Golden tests run the IO actions, capturing their output and redirecting it to the golden files.


src/Module.hs
```diff
@@ -192,6 +192,7 @@ data Decl ty scheme decl expr
   | DefineMacros [(Ident, MacroVar, expr)]
   | Meta decl
   | Example SrcLoc scheme expr
+  | Run expr
```
langston-barrett, line 195: Perhaps this should also take a `SrcLoc` so we can say what failed if there is a failure?

david-christiansen: Given that this is a private implementation detail rather than a public interface to a library, I'd rather add that kind of thing as it becomes necessary (that is, as part of improving error reporting) rather than now, where it doesn't get used.

langston-barrett: There are a few places in this PR where it _could_ get used pretty easily - like in the "Impossible - type checker bug" error message. Shouldn't hold up merge, but IMO it would be nice to report a source location in messages like that.

david-christiansen: Yeah, that would be nice :)

david-christiansen: As always, your feedback has led to an improvement :-)

src/Type.hs
```diff
@@ -98,22 +100,26 @@ instance ShortShow a => ShortShow (TyF a) where
...
+  tInteger    :: a
+  tString     :: a
+  tOutputPort :: a
+  tFun1       :: arg -> arg -> a
+  tMacro      :: arg -> a
+  tIO         :: arg -> a
+  tType       :: a
+  tDatatype   :: Datatype -> [arg] -> a
+  tSchemaVar  :: Natural -> a
+  tMetaVar    :: MetaPtr -> a
```
langston-barrett, line 113: Ah, this is why I complain about aligned source code... It's not easy to glance at this and tell what's changed! :grumpy: :opinionated:

david-christiansen:
A shame we can't tell the web interface to use a better diff algorithm.

src/Type.hs
david-christiansen: A shame we can't tell the web interface to use a better diff algorithm.

gelisam:
Can we merge this? I am planning to use Klister to solve a few Advent of Code challenges this December, in the hope that using Klister in anger will encourage me to smoothen the rough edges, and IO actions would definitely help with that kind of code.

src/Type.hs
gelisam: try clicking on the gear icon and checking "hide whitespace changes", it makes it very clear which of these lines have changed ;)

PR #106, opened 2020-08-28
Expose more string primitives

david-christiansen:
This should make it reasonable to write some actual programs once IO support is merged.

### DIFF
--- a/examples/string.golden
+++ b/examples/string.golden
@@ -2,3 +2,40 @@
 "hello world" : String
 (true) : Bool
 (false) : Bool
+"cde" : (Maybe String)
+(nothing) : (Maybe String)
+"ab" : (Maybe String)
+(nothing) : (Maybe String)
+3 : Integer
+0 : Integer
+(false) : Bool
+(true) : Bool
+(false) : Bool
+(true) : Bool
+(false) : Bool
+(true) : Bool
+(true) : Bool
+(false) : Bool
+(false) : Bool
+(true) : Bool
+(true) : Bool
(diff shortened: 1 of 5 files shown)


examples/string.golden
```diff
@@ -2,3 +2,40 @@
 "hello world" : String
 (true) : Bool
 (false) : Bool
+"cde" : (Maybe String)
```
langston-barrett, line 5: Shouldn't this have a `just` if it's a `Maybe String`?

src/Expander.hs
```diff
@@ -417,6 +417,46 @@ initializeKernel = do
...
+        , ValueClosure $ HO $
+          \(ValueInteger (fromInteger -> start)) ->
+            ValueClosure $ HO $
+            \(ValueInteger (fromInteger -> len)) ->
+              ValueClosure $ HO $
+              \(ValueString str) ->
+                if | start < 0 || start       >= T.length str -> primitiveCtor "nothing" []
+                   | len   < 0 || start + len >  T.length str -> primitiveCtor "nothing" []
+                   | otherwise ->
+                     ValueString $ T.take len $ T.drop start str
```
langston-barrett, line 432: I think this is missing a `just`?

examples/string.golden
david-christiansen: Absolutely! Good catch!

src/Expander.hs
david-christiansen: Yep!

PR #109, opened 2020-09-10
Add deep pattern matching to core language

### LINKED ISSUE #107: Deep pattern matching in kernel language
After working a bit with pattern macros, I think we should have deep pattern matching in the core language rather than as a library.

This is mainly because it makes pattern macros much more useful. Right now, we can't define e.g. `list` as a pattern macro, because it won't work here. I don't think it's so hard to implement, either, especially if we're not caring too much about optimal performance (although proper case trees are not that hard to do either, and probably worth it).

Another thing to consider is whether pattern macros should actually expand to patterns, or whether we should expose some sort of more fundamental interface (e.g. success/failure continuations), which would enable things like view patterns or guards to be defined, I think. But that's for another time, I suppose.

david-christiansen:
Fixes #107.

Macros in pattern position complicate the semantics a little bit,
because pattern variables are lexically identical to any other kind of
identifier. There's no easy way to tell whether a user intends to
shadow an existing macro with a pattern variable or to invoke that
macro in a pattern position.

The solution here is perhaps overly simplistic: bare identifiers in
pattern position are ALWAYS pattern variables - even if they share a
name with a constructor or a macro. The rule is simple, but not very
flexible.

Other alternative approaches could include:

 1. Lexically distinguish pattern variables from other identifiers,
 e.g. by wrapping them in <> or requiring them to start or end with a
 given character. This is the Haskell/Idris approach.

 2. Have some more rules to check for exisiting bindings, and specify
 when they override the local binding. This is the Agda approach.

 3. Allow macros to specify at binding time whether they should be
 considered in pattern contexts, or not. This is the Racket approach (see
 define-match-expander).

I think that any of those can build on top of this basic run-time
support for deep patterns, as well as the expander's handling of
bindings and the like. Our experience in Idris, which went from option
2 to option 1 a few years back, was that simple rules were better than
more flexible yet more complicated rules.

### DIFF
--- a/src/Core.hs
+++ b/src/Core.hs
@@ -70,15 +70,15 @@ data ConstructorPatternF pat
 makePrisms ''ConstructorPatternF
 
 newtype ConstructorPattern =
-  ConstructorPattern { _constructorPattern :: ConstructorPatternF ConstructorPattern }
+  ConstructorPattern { unConstructorPattern :: ConstructorPatternF ConstructorPattern }
   deriving (Data, Eq, Show)
-makeLenses ''ConstructorPattern
+makePrisms ''ConstructorPattern
 
 instance Phased a => Phased (ConstructorPatternF a) where
   shift i = fmap (shift i)
 
 instance Phased ConstructorPattern where
-  shift i = over constructorPattern (shift i)
+  shift i = over _ConstructorPattern (shift i)
 
 instance Phased TypePattern where
   shift _ = id
@@ -370,7 +370,7 @@ instance (AlphaEq typePat, AlphaEq pat, AlphaEq core) => AlphaEq (CoreF typePat 
 
 instance AlphaEq ConstructorPattern where
   alphaCheck p1 p2 =
-    alphaCheck (view constructorPattern p1) (view constructorPattern p2)
+    alphaCheck (unConstructorPattern p1) (unConstructorPattern p2)
 
 instance AlphaEq a => AlphaEq (ConstructorPatternF a) where
   alphaCheck (CtorPattern c1 vars1)
@@ -578,7 +578,7 @@ instance ShortShow Core where
   shortShow (Core x) = shortShow x
 
 instance ShortShow ConstructorPattern where
-  shortShow = shortShow . view constructorPattern
+  shortShow = shortShow . unConstructorPattern
 
 instance ShortShow a => ShortShow (ConstructorPatternF a) where
   shortShow (CtorPattern ctor vars) =
--- a/src/Core.hs
+++ b/src/Core.hs
@@ -63,14 +63,22 @@ data TypePattern
   | AnyType Ident Var
   deriving (Data, Eq, Show)
 
-data ConstructorPattern
-  = ConstructorPattern !Constructor [(Ident, Var)]
-  | AnyConstructor Ident Var
+data ConstructorPatternF pat
+  = CtorPattern !Constructor [pat]
+  | PatternVar Ident Var
+  deriving (Data, Eq, Foldable, Functor, Show, Traversable)
+makePrisms ''ConstructorPatternF
+
+newtype ConstructorPattern =
+  ConstructorPattern { _constructorPattern :: ConstructorPatternF ConstructorPattern }
   deriving (Data, Eq, Show)
-makePrisms ''ConstructorPattern
+makeLenses ''ConstructorPattern
+
+instance Phased a => Phased (ConstructorPatternF a) where
+  shift i = fmap (shift i)
 
 instance Phased ConstructorPattern where
-  shift _ = id
+  shift i = over constructorPattern (shift i)
 
 instance Phased TypePattern where
   shift _ = id
@@ -361,12 +369,16 @@ instance (AlphaEq typePat, AlphaEq pat, AlphaEq core) => AlphaEq (CoreF typePat 
 
 
 instance AlphaEq ConstructorPattern where
-  alphaCheck (ConstructorPattern c1 vars1)
-             (ConstructorPattern c2 vars2) = do
+  alphaCheck p1 p2 =
+    alphaCheck (view constructorPattern p1) (view constructorPattern p2)
+
+instance AlphaEq a => AlphaEq (ConstructorPatternF a) where
+  alphaCheck (CtorPattern c1 vars1)
+             (CtorPattern c2 vars2) = do
     alphaCheck c1 c2
     for_ (zip vars1 vars2) (uncurry alphaCheck)
-  alphaCheck (AnyConstructor _ x1)
-             (AnyConstructor _ x2) = do
+  alphaCheck (PatternVar _ x1)
+             (PatternVar _ x2) = do
     alphaCheck x1 x2
   alphaCheck _ _ = notAlphaEquivalent
 
@@ -566,12 +578,15 @@ instance ShortShow Core where
   shortShow (Core x) = shortShow x
 
 instance ShortShow ConstructorPattern where
-  shortShow (ConstructorPattern ctor vars) =
+  shortShow = shortShow . view constructorPattern
+
+instance ShortShow a => ShortShow (ConstructorPatternF a) where
+  shortShow (CtorPattern ctor vars) =
     "(" ++ shortShow ctor ++
     " " ++ intercalate " " (map shortShow vars) ++
     ")"
-  shortShow (AnyConstructor ident _var) =
-    "(AnyConstructor " ++ shortShow ident ++ " )"
+  shortShow (PatternVar ident _var) =
+    "(PatternVar " ++ shortShow ident ++ " )"
 
 instance ShortShow TypePattern where
   shortShow (TypePattern t) =
--- a/src/Evaluator.hs
+++ b/src/Evaluator.hs
@@ -279,7 +279,7 @@ withScopeOf scope expr = do
 doDataCase :: SrcLoc -> Value -> [(ConstructorPattern, Core)] -> Eval Value
 doDataCase loc v0 [] = throwError (EvalErrorCase loc v0)
 doDataCase loc v0 ((pat, rhs) : ps) =
-  match (doDataCase loc v0 ps) (eval rhs) [(view constructorPattern pat, v0)]
+  match (doDataCase loc v0 ps) (eval rhs) [(unConstructorPattern pat, v0)]
   where
     match _fk sk [] = sk
     match fk sk ((CtorPattern ctor subPats, tgt) : more) =
@@ -288,7 +288,7 @@ doDataCase loc v0 ((pat, rhs) : ps) =
           | c == ctor ->
             if length subPats /= length args
               then error $ "Type checker bug: wrong number of pattern vars for constructor " ++ show c
-              else match fk sk (zip (map (view constructorPattern) subPats) args ++ more)
+              else match fk sk (zip (map unConstructorPattern subPats) args ++ more)
           | otherwise -> fk
         _otherValue -> fk
     match fk sk ((PatternVar n x, tgt) : more) =
--- a/src/Evaluator.hs
+++ b/src/Evaluator.hs
@@ -278,22 +278,21 @@ withScopeOf scope expr = do
 
 doDataCase :: SrcLoc -> Value -> [(ConstructorPattern, Core)] -> Eval Value
 doDataCase loc v0 [] = throwError (EvalErrorCase loc v0)
-doDataCase loc v0 ((pat, rhs) : ps) = match (doDataCase loc v0 ps) pat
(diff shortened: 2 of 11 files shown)


src/Core.hs
```diff
@@ -63,14 +63,22 @@ data TypePattern
...
+  = CtorPattern !Constructor [pat]
+  | PatternVar Ident Var
+  deriving (Data, Eq, Foldable, Functor, Show, Traversable)
+makePrisms ''ConstructorPatternF
+
+newtype ConstructorPattern =
+  ConstructorPattern { _constructorPattern :: ConstructorPatternF ConstructorPattern }
   deriving (Data, Eq, Show)
-makePrisms ''ConstructorPattern
+makeLenses ''ConstructorPattern
```
gelisam, line 75: Since `ConstructorPattern` is a newtype, `makePrisms` and `makeLenses` both create an `Iso`, the only difference is the name; `makePrism` calls it `_ConstructorPattern`, while `makeLenses` names it `constructorPattern`. I like the former, because this allows me to use the more conventional name `unConstructorPattern` for the field name.

src/Core.hs
```diff
@@ -63,14 +63,22 @@ data TypePattern
...
+
+instance Phased a => Phased (ConstructorPatternF a) where
+  shift i = fmap (shift i)
 
 instance Phased ConstructorPattern where
-  shift _ = id
+  shift i = over constructorPattern (shift i)
```
gelisam, line 81: That's just a more complicated version of `id`, no?

src/Evaluator.hs
```diff
@@ -278,22 +278,21 @@ withScopeOf scope expr = do
 
 doDataCase :: SrcLoc -> Value -> [(ConstructorPattern, Core)] -> Eval Value
 doDataCase loc v0 [] = throwError (EvalErrorCase loc v0)
-doDataCase loc v0 ((pat, rhs) : ps) = match (doDataCase loc v0 ps) pat
+doDataCase loc v0 ((pat, rhs) : ps) =
+  match (doDataCase loc v0 ps) (eval rhs) [(view constructorPattern pat, v0)]
   where
```
gelisam, line 283: I think a type signature with comments would make this a lot more comprehensible:

```haskell
match
  :: Eval Value  -- ^ failure continuation
  -> Eval Value  -- ^ success continuation, to be ran in an extended environment
  -> [(ConstructorPatternF ConstructorPattern, Core)]  -- ^ [(pattern, scrutinee)]
  -> Eval Value
```

src/Evaluator.hs
```diff
@@ -278,22 +278,21 @@ withScopeOf scope expr = do
...
-              else withManyExtendedEnv [ (n, x, v) | (n, x) <- vars
-                                       | v <- args
-                                       ] $
-                   eval rhs
-          | otherwise -> next
-        _otherValue -> next
-    match _next (AnyConstructor n x) =
-      withExtendedEnv n x v0 $ eval rhs
+              else match fk sk (zip (map (view constructorPattern) subPats) args ++ more)
+          | otherwise -> fk
```
gelisam, line 292: the `otherwise` branch is redundant, we'll fall through to `_otherValue` regardless

src/Expander/Monad.hs
```diff
@@ -267,9 +268,10 @@ data ExpanderState = ExpanderState
   , _expanderTasks :: [(TaskID, ExpanderLocal, ExpanderTask)]
   , _expanderOriginLocations :: !(Map.Map SplitCorePtr SrcLoc)
   , _expanderCompletedCore :: !(Map.Map SplitCorePtr (CoreF TypePatternPtr PatternPtr SplitCorePtr))
-  , _expanderCompletedPatterns :: !(Map.Map PatternPtr ConstructorPattern)
+  , _expanderCompletedPatterns :: !(Map.Map PatternPtr (ConstructorPatternF PatternPtr))
```
gelisam, line 271: At some point, we should probably switch to a weak map, we're probably keeping data in those `Map`s long after they are no longer needed...

src/Expander/Monad.hs
```diff
@@ -267,9 +268,10 @@ data ExpanderState = ExpanderState
...
   , _expanderCompletedTypePatterns :: !(Map.Map TypePatternPtr TypePattern)
-  , _expanderPatternBinders :: !(Map.Map (Either PatternPtr TypePatternPtr) [(Scope, Ident, Var, SchemePtr)])
+  , _expanderPatternBinders :: !(Map.Map PatternPtr (Either [PatternPtr] (Scope, Ident, Var, SchemePtr)))
+  , _expanderTypePatternBinders :: !(Map.Map TypePatternPtr [(Scope, Ident, Var, SchemePtr)])
```
gelisam, line 274: So when pattern-matching on types, we must still do it a single layer at a time, presumably because of the interaction with the getting-stuck mechanism? If so, it seems like we will still end up writing a macro desugaring deep type-pattern matching into shallow type-pattern matching. But since that macro is manipulating syntax, not types, I guess we don't have the same boot-strapping problem?

src/SplitCore.hs
```diff
@@ -64,7 +64,7 @@ newTypePatternPtr = TypePatternPtr <$> newUnique
 data SplitCore = SplitCore
   { _splitCoreRoot         :: SplitCorePtr
   , _splitCoreDescendants  :: Map SplitCorePtr (CoreF TypePatternPtr PatternPtr SplitCorePtr)
-  , _splitCorePatterns     :: Map PatternPtr ConstructorPattern
+  , _splitCorePatterns     :: Map PatternPtr (ConstructorPatternF PatternPtr)
   , _splitCoreTypePatterns :: Map TypePatternPtr TypePattern
```
gelisam, line 68: Hmm, this datatype seems inconsistent (but it was already inconsistent before this PR, so we don't have to fix it now). One concern I have with those `Ptr`s is that it's not clear whether we should be looking them up in the global `ExpanderState` or in this local `Map PatternPtr`. I see that for `CorePtr`, we defined a dedicated `SplitCorePtr` in order to clarify that we should look it up in this local `Map SplitCorePtr`, not in the global `ExpanderState`. I assume the same is true for the `TypePatternPtr`s and the `PatternPtr`s, but for those we did not create a special `SplitCorePatternPtr` newtype. Why not?

gelisam: I should probably clarify that when I'm saying that "it's not clear whether we should be looking them up in the global ExpanderState or in this local `Map PatternPtr`", I mean that it won't be clear at the use site, where the definition of `SplitCore` won't be right in our face. Looking at this definition it _is_ quite clear what the intent is, my concern is that we chose types which only imprecisely reflect that intent.

src/Expander.hs
```diff
@@ -926,28 +926,58 @@ runTask (tid, localData, task) = withLocal localData $ do
...
+                       [ sc'
+                       | (sc', _, _, _) <- vars
+                       ]
+          withLocalVarTypes
+            [ (var, varStx, t)
+            | (_sc, varStx, var, t) <- vars
+            ] $
+            expandOneExpression ty dest rhs'
+    AwaitingTypePattern patPtr ty dest stx -> do
+      ready <- isJust . view (expanderCompletedTypePatterns . at patPtr) <$> getState
```
gelisam, line 963: a slightly more lensy way of expressing this would be

    ready <- has (expanderCompletedTypePatterns . ix patPtr) <$> getState

src/Core.hs
david-christiansen: Sounds good! Though I don't get the difference here - why does `unConstructorPattern` clash with `constructorPattern`?

src/Core.hs
david-christiansen: It's a more future-proof one. If the thing inside becomes shiftable, then this will do it.

src/Expander/Monad.hs
david-christiansen: Don't we clear out those tables between modules? If we don't, then we should.

src/Expander/Monad.hs
david-christiansen: Matching on types is still a monadic action due to the capturing of the continuation.

If we make the structure of the captured continuations more complicated, then we can support deep matching on types directly, and I think this would probably be a nice improvement. But it's not in this PR.

src/SplitCore.hs
david-christiansen: This is worth thinking about.

So far, we create the `Split...` by wrapping its record constructor around the `Map` from the global state, and accept that it contains junk. This is indeed a bit confusing.

src/Expander.hs
david-christiansen: Is lensiness better here? I'm trying to use lenses in a way that don't make the code incomprehensible to those who haven't marinated their brains in it :)

src/Core.hs
gelisam: It's not a clash between a field named `unFoo` and a lens named `foo`, it's that if the field name is `unFoo` instead of `_foo`, then it doesn't begin with an underscore and so no lens will be generated.

src/Expander/Monad.hs
gelisam: ah, that's much simpler than using week references, good idea.

src/Expander.hs
gelisam: lensy idioms often work better with nested structures; for example, `has (myMap1 . ix k1 . myMap2 . ix k2) s1` seems better to me than

    case view (myMap1 . at k1) s1 of
      Nothing -> False
      Just s2 -> isJust $ view (myMap2 . at k2) s2

but since we're not dealing with a deep structure here, it doesn't really matter. besides, I wasn't really trying to suggest an improvement, I was only trying to spread my knowledge of lensy tricks, in the hope you find that information useful!

src/Expander.hs
```diff
@@ -926,28 +926,58 @@ runTask (tid, localData, task) = withLocal localData $ do
...
+      ready <- do
+        let patternComplete ptr =
+              (view (expanderCompletedPatterns . at ptr) <$> getState) >>=
+              \case
+                Nothing ->
+                  pure False
+                Just layer ->
+                  and <$> traverse patternComplete layer
+        patternComplete patPtr
+        -- TODO traverse the tree and get the layers out, collecting the set of pointers to later accumulate the binders
```
gelisam, line 938: I don't quite understand this TODO. Are you saying that there's some piece of functionality which is missing from this PR, or that one day it would be nice to get rid of `expanderPatternBinders` and to instead put that information in `expanderCompletedPatterns`?

src/Expander.hs
```diff
@@ -1058,26 +1095,23 @@ expandOneForm prob stx
                   else for (zip args' foundArgs) (uncurry schedule)
               linkExpr dest (CoreCtor ctor argDests)
               saveExprType dest t
-            PatternDest _exprTy patTy dest -> do
-              Stx _ loc (_cname, patVars) <- mustBeCons stx
+            PatternDest exprTy patTy dest -> do
```
gelisam, line 1098: I don't think this `exprTy` is used anywhere, is it? Here, we just forward it along in the `PatternDest` of the nested patterns.

gelisam (approved):

src/Expander.hs
gelisam: It was indeed unused, I removed it in a subsequent PR.

gelisam (approved):

src/Evaluator.hs
david-christiansen: Done!

src/Evaluator.hs
david-christiansen: Done!

src/Expander.hs
david-christiansen: It was to write the code immediately below it :-)

src/Expander.hs
david-christiansen: Nice!

src/Expander.hs
david-christiansen: That makes sense! Thanks!

david-christiansen: Does this mean that this version of `pruneLevel` would be equivalent?
```
pruneLevel :: Traversable f => BindingLevel -> f MetaPtr -> Expand ()
pruneLevel l = traverse_ reduce
  where
    reduce ptr =
      modifyState $
      over (expanderTypeStore . ix ptr . varLevel) (min l)
```
The current one is:
```
pruneLevel :: Traversable f => BindingLevel -> f MetaPtr -> Expand ()
pruneLevel l = traverse_ reduce
  where
    reduce ptr =
      modifyState $
      over (expanderTypeStore . at ptr) $
      fmap (over varLevel (min l))
```

david-christiansen:
Travis is failing due to hpack not being available. I promise the tests still pass locally!

src/Expander.hs
gelisam: yes!

PR #110, opened 2020-10-01
remove unused expression type from {Type,}PatternDest

gelisam:
based off of #109, so the diff will be easier to see once that PR is merged.

### DIFF
--- a/examples/deep-patterns.golden
+++ b/examples/deep-patterns.golden
@@ -1,0 +1,8 @@
+(add1 (add1 (zero))) : Nat
+(add1 (add1 (zero))) : Nat
+(add1 (add1 (zero))) : Nat
+(add1 (add1 (zero))) : Nat
+(true) : Bool
+(false) : Bool
+(false) : Bool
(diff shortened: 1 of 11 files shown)


david-christiansen:
It seems I don't have permissions to restart Travis anymore :-(

david-christiansen (approved):
This looks good. I added that field originally because I thought it might be useful for macros, but I haven't managed to come up with a concrete use case, so I think it's good to get rid of it.

gelisam:
> It seems I don't have permissions to restart Travis anymore :-(

I didn't change anything, and I don't see a setting to re-enable it. Are you sure you used to have permission? I vaguely remember that one of the reasons I suggested creating a klister group on github was so that I wouldn't be the only person who has permission to use Travis.

gelisam:
I tried running `curl -sSL https://github.com/sol/hpack/raw/master/get-hpack.sh` locally and it worked fine, and yet when I restarted Travis, it failed the same way again. So I think something changed on the Travis side.

The tests passed when I ran them manually, so I'll merge despite the CI failure.

PR #113, opened 2020-10-02
Replace hpack with a standard cabal file

### LINKED ISSUE #111: Get rid of hpack
From my perspective, `hpack` is causing more problems than it solves.

With common stanzas, the repetetiveness of the cabal file format is much reduced, and I keep having to remember to run hpack when I'm building the project. It's also been causing some intermittent CI issues.

How do you all feel about replacing it with a `.cabal` file with a common stanza?

david-christiansen:
Fixes #111. This touches Travis, so I want to make sure that passes before I merge :)


david-christiansen:
The tests pass with GHC 8.6.5, 8.8.4, and 8.10.2 locally, and Stack does its thing as well for me. I'll wait for Travis to be happy, then mark this as review-ready.

david-christiansen:
The PR ends up also increasing the Travis build matrix. I had to mess with the Travis file anyway, so I figured I'd add a newer GHC to the mix while I was at it.

klister.cabal
```diff
@@ -0,0 +1,120 @@
+cabal-version: 2.2
+
+name:           klister
```
langston-barrett, line 3: nit: In general, I prefer non-aligned code. Consider what the diff would look like if we added the famous `really-long-cabal-field-everything-needs-to-be-reindented-now: true` field to this config.

klister.cabal
```diff
@@ -0,0 +1,120 @@
...
+    megaparsec           >= 7.0.5 && < 9.1,
+    mtl                  >= 2.2.2 && < 2.3,
+    prettyprinter        >= 1.2 && < 1.8,
+    text                 ^>= 1.2,
+    transformers         ^>= 0.5
+
+library
+  import: deps
+  exposed-modules:
+      Alpha
```
langston-barrett, line 41: Indentation here is 4 but in `common deps`/`build-depends:` it's 2. Any reason for the discrepancy?

langston-barrett (approved):
Two tiny nits but nothing that should hold up merging.

gelisam (approved):

gelisam:
Merging so I can restart the other PRs which failed because travis couldn't download hpack. We can address the nitpicks in a later PR if needed.

PR #114, opened 2020-10-12
make Hedgehog tests 10 times faster

gelisam:
by running 10 times less tests. these take a lot more time to run than
all our other tests, so reducing those makes interactively running the
tests a lot more pleasant.


david-christiansen:
Looks good to me. Is there a reason CI didn't run?

gelisam:
> Is there a reason CI didn't run?

Huh. https://travis-ci.org/github/gelisam/klister says they _did_ run and pass, they just don't show up on github for some reason. I'll assume this is a transient issue, but let's keep our eyes open to see if it keeps happening, and if so I'll investigate further.

PR #116, opened 2020-10-20
Open and close syntax objects

david-christiansen:
Operations for looking inside syntax objects, hopefully more primitive than the current `syntax-case`.


david-christiansen:
For instance,
```
(syntax-case stx
  [(list (a b c)) ...])
```
becomes
```
(case (open-syntax stx)
  [(list-contents (:: a (:: b (:: c (nil))))) ...])
```

langston-barrett (commented):
It would be good to have an example file exercising this functionality

david-christiansen:
I agree!

I mostly put this here so @gelisam could use it, because I've been too slow getting the proper suite of examples working.

gelisam:
> I mostly put this here so @gelisam could use it

Hmm, I don't remember what I needed it for...

src/Expander.hs
```diff
@@ -486,6 +534,15 @@ initializeKernel = do
       , ("Bool", 0, [("true", []), ("false", [])])
       , ("Problem", 0, [("declaration", []), ("expression", [tType]), ("type", []), ("pattern", [])])
       , ("Maybe", 1, [("nothing", []), ("just", [tSchemaVar 0])])
+      , ("List", 1, [("nil", []), ("::", [tSchemaVar 0, Prims.primitiveDatatype "List" [tSchemaVar 0]])])
+      , ("Syntax-Contents"
+        , 1
+        , [ ("list-contents", [Prims.primitiveDatatype "List" [tSchemaVar 0]])
+          , ("integer-contents", [tInteger])
+          , ("string-contents", [tString])
+          , ("identifier-contents", [tString])
```
gelisam, line 543: Hmm, I would have expected `identifier-contents` to contain an opaque `Identifier` type holding both an inspectable name and the uninspectable scopes, so that closing a syntax after opening it is the identity. As is, whenever you encounter an identifier you have to be careful to also keep the original `Syntax` so you can re-attach the same scopes when you close it back.

gelisam (approved):

src/Expander.hs
david-christiansen: We discussed this on a call :-)

PR #122, opened 2020-11-24
GitHub Action config

gelisam:
from https://kodimensional.dev/github-actions


gelisam:
the stack-based CI build is now working, but it only tests one GHC version, the one pinned by `stack.yaml`.

the cabal-based CI build failed because it couldn't find a build plan with ghc-8.10.2, which is strange because our cabal file says we did test with that version. I will try to build with that version locally in order to double-check.

gelisam:
```
$ cabal new-test
cabal: Cannot test the package klister-0.1 because none of the components are
available to build: the test suite 'klister-tests' is not available because
the solver did not find a plan that included the test suites. Force the solver
to enable this for all packages by adding the line 'tests: True' to the
'cabal.project.local' file.
```

What? Since when is it necessary to explicitly enable tests in order to run `cabal new-test`?

gelisam:
apparently I need `cabal new-test --enable-tests`??

gelisam:
In any case, a build plan is easily found. So why can't the GitHub Action find it?

My build plan found `hedgehog-1.0.3`, but the GitHub Action error message implies that it didn't consider anything more recent than `hedgehog-1.0.2`. So maybe I just need to throw in a `cabal new-update`?

gelisam:
Strange, I see that `cabal update` (we don't need the `new-` prefix anymore, the new behaviour is the default) did run on CI, but hedgehog-1.0.3 still wasn't considered. Maybe the caching system is somehow causing `cabal update`'s changes to `~/.cabal` to be ignored?

gelisam:
Tests fails on windows, presumably because the golden files end with linux newlines. We weren't testing windows before, so this isn't a new problem introduced by this PR, it's only a problem discovered by this PR. So I think I'll just disable windows testing for now. I guess I should open a ticket to make the tests pass on Windows, but I think the project is small enough that Windows support is not worth the effort until somebody explicitly asks for it.

gelisam:
All tests are green! Caching seems to be working pretty well, too; the previous builds were 8 minutes each, while these took only 2 minutes each.

Issue #128, opened 2020-12-01
"expected" and "actual" are still swapped?


gelisam:
I remember that our error messages used to say "expected ACTUAL but found EXPECTED". We fixed that, but the error message is still backwards:

```
-- Type mismatch at unknown location. Expected Syntax but got (List Integer)
(example (the (List Integer) 'foo))
```

If I had written the type-checker myself, I would definitely have used unification so that I wouldn't have to think about when to check and when to synthesize. And then I would expect to see bugs like the above, because there is no way to know which side of a unification is the expected type and which side is the actual type. But since @david-christiansen wrote the type-checker, I'm pretty sure he used a bidirectional algorithm, in which we do know which side is expected and which side is actual. So what's going on?

Also, why is the location unknown? The very similar `(example (+ 'foo 1))` does give a location.

david-christiansen:
The type checker is not written bidirectionally. It's a common variation on the usual Hindley-Milner style.

In textbook H-M, the type checker is really a type inferencer. That is, you give it a context and a term, and it monadically returns a type. A downside of this approach is that it creates tons of metavariables, because if something's type is not restricted, it has to create a meta to say so. Then, the calling context unifies the returned thing with the type that it had to have anyway. The version used in Klister is one where the type checker  takes a context, a term, and a type, and either succeeds or fails monadically. The behavior of the inferencer can be recovered by passing this version a fresh meta, then checking what it's been solved to afterwards. But now a calling context can pass a type inwards, which saves on fresh metas but also allows some more control over error messages.

The bug here is just two arguments flipped to unify somewhere.

david-christiansen:
As far as unknown location goes, the type checker presently keeps track of source locations in a table indexed by SplitCorePtr. Some operator somewhere isn't filling in the table for itself.

david-christiansen:
Fixed in #129.

Issue #42, opened 2020-04-18
Golden testing for expected failures


david-christiansen:
The `non-examples` directory contains files we expect to fail. We should augment the testing infrastructure to ensure they continue to fail in the way we expect.

doyougnu:
started on this:

```
  which-problem:                                     OK (1.20s)
    incorrect-context:                                 FAIL (0.04s)
      tests/Golden.hs:53:
      expected:
      #[incorrect-context.kl:5.10-5.22]<(define x 5)>:
        Used in a position expecting an expression
        but is valid in a position expecting a top-level declaration or example
      ExitFailure 1
      
      got:
      #[incorrect-context.kl:5.10-5.22]<(define x 5)>:
        Used in a position expecting an expression
        but is valid in a position expecting a top-level declaration or example
    bound-vs-free:                                     OK (0.04s)
    meta-macro:                                        OK (0.14s)
    integer-syntax:                                    OK (0.05s)
    import:                                            OK (0.03s)
```

but then I noticed that `cabal test --test-option='--accept'` returns:
```
Running 1 test suites...
Test suite klister-tests: RUNNING...
Invalid option `--accept'

Usage: klister-tests [-p|--pattern PATTERN] [-t|--timeout DURATION] 
                     [-l|--list-tests] [-j|--num-threads NUMBER] [-q|--quiet] 
                     [--hide-successes] [--color never|always|auto] 
                     [--ansi-tricks ARG] [--hedgehog-replay ARG] 
                     [--hedgehog-show-replay ARG] [--hedgehog-tests ARG] 
                     [--hedgehog-discards ARG] [--hedgehog-shrinks ARG] 
                     [--hedgehog-retries ARG]

Test suite klister-tests: FAIL
```

which means that `tasty-golden` isn't being used at all. Is there any reason? Are is it time to actually use it?

doyougnu:
it looks pretty good to me:
```
    custom-module-test:                                OK (0.78s)
    which-problem:                                     OK (1.02s)
    incorrect-context:                                 FAIL (0.04s)
      Test output was different from 'examples/failing-examples/incorrect-context.golden'. It was:
      #[incorrect-context.kl:5.10-5.22]<(define x 5)>:
        Used in a position expecting an expression
        but is valid in a position expecting a top-level declaration or example
    bound-vs-free:                                     OK (0.04s)
    meta-macro:                                        OK (0.16s)
    integer-syntax:                                    OK (0.05s)
[...]

PR #129, opened 2020-12-02
Improve type error messages

david-christiansen:
Fixes both problems pointed out in #128, as well as a few more similar ones.


david-christiansen:
We don't have a good mechanism for negative tests yet, unfortunately. (#42)

david-christiansen:
I guess to make the robot happy, I should explicitly say

Fixes #128

gelisam (approved):

PR #131, opened 2020-12-21
investigate macos test issues

gelisam:
I don't have a solution yet, but the error only occurs on CI, so I have to open a PR in order to investigate further.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,19 +13,6 @@ jobs:
     strategy:
       matrix:
         include:
-            # latest GHC, default OS
-          - cabal: "3.2"
-            ghc: "8.10.2"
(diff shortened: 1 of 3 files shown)


gelisam:
I expected the test to fail with a clearer error message (comparing the `Show` representation of the strings instead of the string themselves, to be able to see any differences in the unicode or newline encoding), but instead the test passed! Did I accidentally find a fix, or is this a sporadic failure?

gelisam:
Nope, it's consistent: the test fails with `goldenVsString`, and passes with my reimplementation. I don't know what the difference is, but I guess we should keep mine!

gelisam:
oops, I forgot to revert `temporarily focus on the failing macOS test`, I'll do it directly on `master`

PR #132, opened 2020-12-21
address warnings

gelisam:
should be fairly uncontroversial, I think.


david-christiansen:
Thank you! And also for all the heroic CI migration work!

Issue #105, opened 2020-08-28
We should have syntax parameters


david-christiansen:
This would make a number of macros nicer to write. See http://scheme2011.ucombinator.org/papers/Barzilay2011.pdf for details.

gelisam:
With regular, non-syntax parameters, each parameter behaves like `Reader`, in that the effect of [`parameterize`](https://docs.racket-lang.org/reference/parameters.html#%28form._%28%28lib._racket%2Fprivate%2Fmore-scheme..rkt%29._parameterize%29%29) is to change the value of the parameter during the dynamic extend of the `parameterize` body.

With _syntax_ parameters, the dynamic-extend of the body of [`syntax-parameterize`](https://docs.racket-lang.org/reference/stxparam.html#%28form._%28%28lib._racket%2Fstxparam..rkt%29._syntax-parameterize%29%29) is quite short, as it is typically inside a macro definition and thus just returns a syntax object. What makes syntax parameters more useful than regular parameters in this circumstance is that the values of the syntax parameters gets attached to the returned syntax object, so that any macro invocation it contains also sees the values given to `syntax-parameterize`.

gelisam:
We discussed the following implementation technique.

The expander's state is divided into two. The `ExpanderState` evolves over time as e.g. new definitions get added, while `ExpanderLocal` holds data which would normally be in a `Reader` if we were processing the tasks as we encountered them, but is instead stored with each task (in `ExpanderState`'s `_expanderTasks :: [(TaskID, ExpanderLocal, ExpanderTask)]` field) so that we can evaluate the tasks out of order.

Thus, we could attach the current value of all the syntax parameters to `ExpanderLocal`, and they should propagate correctly from the macro action task which executes `syntax-parameterize` to the syntax-object creation task which captures those values. That syntax-object creation task might already be associating the `ExpanderLocal` to the syntax object for scoping reasons.

gelisam:
In #166 , I suggested the name "ReaderRef" instead of "parameter", to avoid confusion with parameters-vs-arguments. Thus, correspondingly, I would like to suggest the name "SyntaxReaderRef" for syntax parameters, even though the name "syntax parameter" is a lot less ambiguous than the name "parameter".

PR #135, opened 2021-01-26
investigate the flaky macOS test

gelisam:
this is not a real PR, it's just a way to trigger the test on CI, because I cannot reproduce the problem on my machine.

### DIFF
--- a/klister.cabal
+++ b/klister.cabal
@@ -28,6 +28,7 @@ common deps
     bifunctors           >= 5.5.5 && < 5.6,
     containers           >= 0.6 && < 0.7,
     directory            >= 1.3.3 && < 1.4,
+    exceptions,
     extra                >= 1.6.18 && < 1.8,
     filepath             >= 1.4.2 && < 1.5,
     lens                 >= 4.17.1 && < 4.20,
--- a/repl/Main.hs
+++ b/repl/Main.hs
@@ -74,7 +74,7 @@ mainWithOptions opts =
   case optCommand opts of
     Repl (ReplOptions Nothing) -> do
       ctx <- mkInitContext (KernelName kernelName)
-      void $ execExpand ctx initializeKernel
+      void $ execExpand ctx (initializeKernel stdout)
       repl ctx initialWorld
     Repl (ReplOptions (Just file)) -> do
       (_mn, ctx, result) <- expandFile file
@@ -97,7 +97,7 @@ mainWithOptions opts =
   where expandFile file = do
           mn <- moduleNameFromPath file
           ctx <- mkInitContext mn
-          void $ execExpand ctx initializeKernel
+          void $ execExpand ctx (initializeKernel stdout)
           st <- execExpand ctx $ completely $ do
             void $ visit mn
             getState
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -54,7 +54,7 @@ import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import Data.Traversable
 import System.Directory
-import System.IO (stdout)
+import System.IO (Handle)
 
 import Binding
 import Core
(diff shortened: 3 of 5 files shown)


gelisam:
I managed to reproduce the problem by re-running the test in CI about 5 times. I cannot reproduce it locally on my mac.

gelisam:
It failed on the same test as in #92 (except for the timings):

```
    which-problem:             OK    io:                        FAIL (0.06s)
      tests/Golden.hs:45:
      expected:
      #<output port> : Output-Port
      #<IO action> : (IO Integer)
      hello world!
      
      got:
      #<output port> : Output-Port
      #<IO action> : (IO Integer)
       (1.32s)hello world!
```

that is, it is the `io` test which is failing, but its output is interleaved with `which-problem`'s for some reason. The test must have failed before that interleaving happened, because the error message is only printed if the test fails. Because of the interleaving, it is not clear what the problem is. Let's try changing the output format next.

gelisam:
Here's what the failure looks like in this easier-to-debug format:

      expected:
      "#<output port> : Output-Port\n#<IO action> : (IO Integer)\nhello world!\n"expected:
      "#<output port> : Output-Port\n#<IO action> : (IO Integer)\nhello world!\n"
      got:
      "#<output port> : Output-Port\n#<IO action> : (IO Integer)\n (1.31s)hello world!\n\n"
      got:
      "#<output port> : Output-Port\n#<IO action> : (IO Integer)\n (1.31s)hello world!\n\n"

I printed each line twice so that when that string with repeated lines gets interleaved with the ` (1.31s)` from the previous test, it only overlaps with one of the twp occurrences of the string, so we can read the other occurrence with no overlay. However, surprise! the ` (1.31s)` appears twice, which means it is part of the string itself! This is definitely not supposed to happen.

gelisam:
The `(1.31s)` is captured right before the `hello world!`, which is captured via a call to `capture_`, which saves whatever's written to `stdout` to a String. `capture_` is very careful to flush `stdout` before it starts capturing, so the printing of `(1.31s)` would have to occur after that `hFlush` call in order to be captured. maybe the test framework is running multiple threads, and is printing the time for the previous test while the next test is running?

gelisam:
Oh! I think I know what's happening. `hspec` is probably trying to improve performance by running multiple tests in parallel, and the printing the results sequentially. That is probably a good strategy in most cases, but since some of our tests are using a global resource (namely`stdout`), this is causing non-determinism in our case.

gelisam:
Darn, looks like parallelism is off by default and that we need to opt in using the `parallel` wrapper: https://hspec.github.io/parallel-spec-execution.html

Okay, then maybe the way hspec supports this feature is by having a pool of worker threads, plus a reporting thread printing the results sequentially, and when `parallel` is turned off, the pool only has a single worker thread, but there is still an extra reporting thread? And that this is normally fine because most tests don't hijack `stdout`?

gelisam:
Silly me, we're using Tasty, not hspec!

gelisam:
Yup, Tasty's readme boasts that it "runs tests in parallel but report results in a deterministic order". Can we turn that off?

gelisam:
Yes, with `--num-threads=1` or `setEnv "TASTY_NUM_THREADS" "1"`.

gelisam:
I used this setting locally, and surprisingly, the tests don't seem to be taking any longer to run. Maybe there is something about my machine which is somehow causing tasty to always use 1 thread regardless? This would explain why I have not been able to reproduce the problem locally.

gelisam:
If that doesn't work, we could avoid hijacking `stdout` by implementing parameters (#105) and running tests in an environment in which `stdout` points to a string buffer. That sounds more fun too ;)

gelisam:
Using a single thread didn't work. Parameters it is!

gelisam:
Closing in favour of #105.

gelisam:
Wait, no, let's keep this ticket open, so that once parameters are implemented, we remember both to change the test suite to use them instead of `silently_`, and to run the test suite several times in order to confirm that the problem is gone. Seems to happen fairly frequently, here are the results of the last 9 runs:

    fail
    pass
    pass
    pass
    pass
    fail
    pass
    pass
    fail

gelisam:
We didn't implement parameters after all. Instead, we added an extra argument to `initializeKernel` specifying which Handle to use for `stdout`. Once we implement parameters, we should get rid of that extra argument.

gelisam:
I ran the tests a bunch of times to make sure they aren't still flaky:

    pass
    pass
    pass
    pass
    pass
    pass
    pass

PR #136, opened 2021-03-07
Move not, binary-{and,or}, and, or from examples/ to stdlib/

langston-barrett:
Progress on #54

### DIFF
--- a/examples/bool.kl
+++ b/examples/bool.kl
@@ -1,40 +1,5 @@
 #lang "prelude.kl"
 
-(import (shift "prelude.kl" 1))
-(import (shift "quasiquote.kl" 1))
-
-(define not
-  (lambda (b)
-    (if b (false) (true))))
-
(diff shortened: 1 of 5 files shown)


gelisam (approved):

gelisam:
> Unfortunately once https://github.com/gelisam/klister/pull/276 lands these functions need to be type aware

why? these functions operate on booleans, e.g. `binary-or` is a two-argument version of `or`, not a bitwise or, so it doesn't have anything to do with the `Word8`s introduced in #276.

are you saying that while in Haskell, [(||)](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:-124--124-) and [(.|.)](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Bits.html#v:.-124-.) are different operators from different names, in Klister you would like them to have the same name?

PR #138, opened 2021-03-08
fix golden tests and lambda case

gelisam:
the existing implementation of lambda-case was strangely-limited: only a single pattern-rhs pair was allowed. with this PR, several are now allowed.

I added a new file to test this feature, and in doing so, I found a flaw in my reimplementation of the golden tests accidentally dropped an important feature: generating the golden files when they don't yet exist. this PR fixes that too.


gelisam:
Darn, while we can easily fix the implementation in this repo, the incorrect implementation is immortalized in [our TyDe presentation](https://youtu.be/FyeWwYfqTHo?t=142) :(

langston-barrett (approved):

PR #139, opened 2021-03-14
explain lambda-case and mega-const

gelisam:
I added two comments, each explaining what a macros does using a commented-out example of what _code_ the macro expands to.

Doing so made me realize that the `(example <expr>)` form has two flaws. First, since the output is in the golden file, it is cumbersome for the reader to jump between the example and its output. I think I would prefer `(example <expr> <expected-output>)`, similarly to [`check-equal?`](https://docs.racket-lang.org/rackunit/api.html?q=syntax-%3E#%28def._%28%28lib._rackunit%2Fmain..rkt%29._check-equal~3f%29%29).

Second, if `(<macro> <args>)` outputs `(f 1 2 3)`, then `(example (<macro> <args>))` outputs the result of the call `(f 1 2 3)`, not the code `'(f 1 2 3)`. When giving an example of what a macro does, like I did in those two comments, showing the code is often more helpful. I think we should thus have a variant of `example` which does that.

### DIFF
--- a/examples/lambda-case.kl
+++ b/examples/lambda-case.kl
@@ -4,6 +4,14 @@
 (import (shift "quasiquote.kl" 1))
 
 (define-macros
+  -- (lambda-case
+  --   [(nil) 'nil]
+  --   [(:: a b) 'cons])
+  -- =>
+  -- (lambda (x)
+  --   (case x
+  --     [(nil) 'nil]
+  --     [(:: a b) 'cons]))
   ([lambda-case
     (lambda (stx)
       (syntax-case stx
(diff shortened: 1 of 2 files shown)


david-christiansen:
As far as `example` goes, I agree that it's not great for a test suite, but I like it's current behavior for interactive programming (e.g. instead of a REPL).

Can we write `check-equal?` as a macro? Something like:
```
(check-equal NAME PRED ACTUAL EXPECTED)
====>
(example (if (PRED ACTUAL EXPECTED) (success 'NAME) (failure 'NAME ACTUAL EXPECTED)))
```
with a suitable sum type for the success and failure constructors? I think this would yield readable output with a minimum of work, and it could later be made into something with an effect on failure.

david-christiansen (approved):

PR #143, opened 2021-03-29
interpret ",@" as unquote-splicing

gelisam:
quasiquote does not yet make use of this syntax, but it's a first step in that direction.


gelisam:
implements #142

langston-barrett (approved):

PR #146, opened 2022-07-08
Fix type-case bug

### LINKED ISSUE #145: "Type pattern info not added" when matching on builtins of kind Type
(define-macro (type-name)
      (>>= (which-problem)
           (lambda (prob)
             (case prob
               [(expression t)
                (type-case t
                  [(-> a b)
                   (pure '"arrow")]
                  [String
                   (pure '"string")])]))))
    
    --   Internal error during expansion! This is a bug in the implementation.
    --   Type pattern info not added
    (example (the String (type-name)))

The same happens with the types `Syntax`, `Integer`, and `Type`, but not with `(-> a b)`, `(Macro a)`, `(IO a)`, `Unit`, `Bool`, `(Maybe a)`, nor `(List a)`, so the bug seems to be about matching on builtin types of kind `Type`.

gelisam:
Fixes #145

### DIFF
--- a/examples/which-problem.golden
+++ b/examples/which-problem.golden
@@ -4,3 +4,4 @@
 #<closure> : (Bool → (Bool → (Bool → (Bool → Unit))))
 (both #<closure> #<closure>) : ∀(α : *) (β : *) (γ : *). (Both (α → (β → (γ → Syntax))))
 (both #<closure> #<closure>) : ∀(α : *) (β : *) (γ : *). (Both (α → (β → (γ → Syntax))))
+"string" : String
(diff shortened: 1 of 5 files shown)


david-christiansen:
Shouldn't the comment in which-problem.kl be deleted now that you've fixed it?

gelisam:
> Shouldn't the comment in which-problem.kl be deleted now that you've fixed it?

fixed, thanks!

gelisam:
I'm thinking of improving this further by merging `_expanderCompletedTypePatterns` and `_expanderTypePatternBinders` into a single Map, to make illegal states unrepresentable.

gelisam:
Nah, there are too many other similar pairs of fields in ExpanderState. I don't want to make a big refactoring which would make _everything_ unrepresentable, and I don't want to make a small refactoring which would let readers wonder why type patterns are represented one way while regular patterns are represented in a different way. I'll leave things as-is, improving `linkTypePattern` was good enough.

david-christiansen:
It's nice seeing more work on Klister!

PR #147, opened 2022-07-31
Fix variable binding in type patterns

gelisam:
When using the pattern `(-> a b)`, both `a` and `b` were bound to the same type! The cause was using list comprehension instead of parallel list comprehension (`[(x,y) | x <- ..., y <- ...]` instead of `[(x,y) | x <- ... | y <- ...]`), thus resulting in the list `[(x1,y1), (x1,y2), (x2,y1), (x2,y2)]` instead of `[(x1,y1), (x2,y2)]`.

### DIFF
--- a/examples/which-problem.golden
+++ b/examples/which-problem.golden
@@ -7,3 +7,4 @@
 "String" : String
 "String -> String" : String
 "String -> String -> String" : String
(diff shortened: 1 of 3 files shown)


david-christiansen:
Nice catch!

PR #153, opened 2022-08-03
wrap literals in #%integer-literal or #%string-literal

### LINKED ISSUE #79: Integers, not signals

gelisam:
this makes it possible for a #lang to e.g. interpret every integer literal as something else, e.g. a polymorphic `fromInteger n`.

Fixes #79.


gelisam:
This is a draft PR because the test doesn't pass. That test happens to use a quasiquote, but it turns out that quasiquotes don't currently support integers, and that adding support for integers requires adding an `integer-syntax` primitive and adding an `integer` pattern to `syntax-case`.

david-christiansen:
That makes sense!

gelisam:
I have now added the missing `integer-syntax` primitive, the `integer` pattern, integer support in quasiquote, and now the tests pass, this PR is now ready for review!

examples/custom-literals.kl
```diff
@@ -0,0 +1,21 @@
+#lang "prelude.kl"
+
+(import (shift "prelude.kl" 1))
+(import "define-syntax-rule.kl")
+
+
+(define-macro (my-integer n)
+  (pure `(+ 1 (#%integer-literal ,n))))
```
david-christiansen, line 8: You sneaky person. I was about to react negatively to the test, and then I saw what was going on :-)

david-christiansen (approved):
LGTM! Thanks!

examples/custom-literals.kl
gelisam: It was the simplest possible non-default interpretation of integer literals I could think of. Do you have a less sneaky non-default implementation in mind? We don't have type classes yet, so I couldn't use `(fromInteger (#%integer-literal ,n))`.

david-christiansen:
Nah, this is a great test, I just had fun reading it!

lør. 6. aug. 2022 21.19 skrev Samuel Gélineau ***@***.***>:

Issue #156, opened 2022-08-06
implicit-conversion example


gelisam:
I want to implement an example which demonstrates that it is possible to define a `#lang` with a custom type system. More specifically, I want a type system which, as Conor McBride exhorts, does more than just rejecting bad programs, and in addition generates terms.

My current plan is to implement Scala's "implicit conversion" feature. That is, a function from `A` to `B` can be added to the list of implicit conversion functions currently in scope, and then if within that scope an `A` is used where a `B` is expected, the conversion function is automatically inserted. Here is some example code I am hoping to support:

```racket
#lang "implicit-conversion.kl"

(example
  (let-implicit string-length
    (+ 1 "foo")))
```

Which would become

```racket

#lang "implicit-conversion.kl"

(example
  (+ 1 (string->length "foo")))
```

Since my goal is only to demonstrate that this is possible, not to use the `#lang` in anger, I only plan to implement the minimum set of functionality to support the above example. In particular, I am not planning to include lambdas in the `#lang`.

gelisam:
Implemented by #157.

Issue #155, opened 2022-08-06
type-equality primitive


gelisam:
It is possible to write a macro which checks whether types like `(-> Integer String)` are equal, but for types like `(-> (Maybe Integer) String)`, this is harder because the user can later introduce more datatypes which the macro doesn't know about.

I propose to provide a primitive macro which determines whether two types are known to be equal, or known to be apart. If the answer depends on what type unification variables will eventually resolve to, the macro is of course stuck until enough information is available.

david-christiansen:
This seems like it's clearly useful. But rather than a macro, I'd make it an effect in the macro monad. That is, a type like`Type -> Type -> Macro Bool` rather than `Syntax -> Macro Syntax`, and not bound in a transformer environment.

PR #157, opened 2022-08-08
implicit conversion example

gelisam:
A `#lang` which provides a `let-implicit` form which makes a function eligible to be used implicitly, to convert a value to the expected type as needed. For example,

```racket
#lang "implicit-conversion.kl"

(example
  (let-implicit string-length
    (+ 1 "foo")))
```

becomes

```racket

#lang "implicit-conversion.kl"

(example
  (+ 1 (string->length "foo")))
```

Implements #156.


gelisam:
This is only a draft PR even though the test passes, because I am unhappy with how the implementation is too tailored to this specific example:

- [ ] Type equality (when checking whether to use a value as-is or to introduce an implicit conversion) only considers the types `Integer` and `String`. With #155, it would be possible to support even more types.
- [ ] `implicit-let` cannot be nested. This is because `implicit-let` assumes that its body is a `Reader something`, but it calls `runReader` and thus returns something which cannot be the body of a second `implicit-let`.
- [ ] Each example must contain exactly one `implicit-let`. Otherwise, the `Reader` implementation detail will leak, because `example` doesn't call `runReader`.
- [ ] The first argument of `implicit-let` must be special. The `#lang` provides a special `+` which is wrapped in a `Reader`, but is careful to provide a `string-length` which is not wrapped in a `Reader`. This means `string-length` cannot be used explicitly in the code, it can only be used implicitly via `let-implicit`. Instead, `string-length` and every function made available inside the `#lang` should be wrapped in a `Reader`, and so `let-implicit` should expect its first argument to be a function which is wrapped in a Reader.
- [ ] The search for an appropriate implicit conversion function is rigger. Since we know that exactly one implicit function is in scope, we use that one without checking if it has the type we need.
- [ ] I was not planning to implement lambdas, but since I want the first argument to `let-implicit` to be ordinary code returning a function, I should probably implement lambdas instead of limiting the user to the builtin functions provided by the `#lang`.

examples/pmatch.kl
```diff
@@ -23,12 +23,6 @@
     (nothing)
     (just A)))
 
-(meta
-  (defun syntax->list (stx)
-    (syntax-case stx
-      [() (nil)]
-      [(cons x xs) (:: x (syntax->list xs))])))
-
```
gelisam, line 31: moved to `stdlib/list.kl`

examples/pmatch.kl
```diff
@@ -45,12 +39,6 @@
             [(else other) (pure (nothing))])]
          [_ (pure (nothing))]))))
 
-(meta
-  (defun list->syntax (stx-lst stx)
-    (case stx-lst
-      [(nil) (empty-list-syntax stx)]
-      [(:: x xs) (cons-list-syntax x (list->syntax xs stx) stx)])))
-
```
gelisam, line 53: moved to `stdlib/list.kl`

examples/reader-test.kl
```diff
@@ -0,0 +1,63 @@
+#lang "prelude.kl"
```
gelisam, line 1: Should I put those examples in `reader.kl` instead? I created it because I saw a bunch of similarly-named `*-test.kl` files, but I now realize that those are all exceptions for tests which can't be written in the same file as the code they exercise.

stdlib/list.kl
```diff
@@ -7,10 +7,6 @@
 (import "n-ary-app.kl")
 (import "defun.kl")
 
-(datatype (List A)
-  (nil)
-  (:: A (List A)))
-
```
gelisam, line 13: clashed with the List from `kernel`.

stdlib/list.kl
```diff
@@ -54,4 +65,25 @@
...
+(example
+  (reverse (list 1 2 3)))
+
+(example
+  (syntax->list '(a b c)))
+
+(example
+  (list->syntax (list 'a 'b 'c) 'here))
+
+(export List nil :: foldr map filter snoc reverse syntax->list list->syntax list)
```
gelisam, line 89: re-export `List(..)` from `kernel`.

gelisam:
I don't plan to implement type-equality nor syntax parameters as part of this PR, those improvements will come later. This PR is thus finally ready for review!

gelisam:
Merging this without review, because this doesn't change the Klister language itself, I'm just using the language for a personal project and adding my code to `examples`, the small list of all Klister code which has ever been written.

PR #161, opened 2022-09-08
mutual recursion

### LINKED ISSUE #160: mutually-recursive functions
In #158 , @sorawee asked about the Klister equivalent of the following Racket code:

```racket
(define (even? n) (if (zero? n) #true (odd? (sub1 n))))
(define (odd? n) (if (zero? n) #false (even? (sub1 n))))
```

Currently, there is no Klister equivalent because Klister does not support mutually-recursive functions. Klister's scoping rules require identifiers to be defined before they are used, so the above code would definitely be rejected. Klister does have a `flet` builtin for defining recursive functions, and the prelude defines a `defun` wrapper around `define` and `flet` for defining top-level recursive functions:

```racket
#lang "prelude.kl"
(import "list.kl")

(defun fact (n)
  (if (= n 0)
    1
    (* n (fact (- n 1)))))

-- (list 1 2 6 24 120)
(example (map fact (list 1 2 3 4 5)))
```

It thus seems possible to define a similar `defuns` macro for defining several mutually-recursive macros:

```racket
#lang "prelude.kl"
(import "list.kl")

(defuns
  [even? (n)
   (if (= n 0) true (odd? (- n 1)))]
  [odd? (n)
   (if (= n 0) false (even? (- n 1)))])

-- (list false true false true false)
(example (map even? (list 1 2 3 4 5)))

-- (list true false true false true)
(example (map odd? (list 1 2 3 4 5)))
```

For example, the macro could desugar these mutually-recursive functions to a single recursive function which receives and produces `Either`s:

```racket
#lang "prelude.kl"
(import "list.kl")

(datatype (Either A B)
  (left A)
  (right B))

(defun even?/odd? (n/n)
  (flet [even? (n)
         (case (even?/odd? (left n))
           [(left r)
            r]
           [_
            (error '"never happens: even?/odd? always returns a left when receiving a left")])]
    (flet [odd? (n)
           (case (even?/odd? (right n))
             [(right r)
              r]
             [_
              (error '"never happens: even?/odd? always returns a right when receiving a right")])]
      (case n/n
        [(left n)
         (left (if (= n 0) true (odd? (- n 1))))]
        [(right n)
         (right (if (= n 0) false (even? (- n 1))))]))))
(defun even? (n)
  (case (even?/odd? (left n))
    [(left r)
     r]
    [_
     (error '"never happens: even?/odd? always returns a left when receiving a left")]))
(defun odd? (n)
  (case (even?/odd? (right n))
    [(right r)
     r]
    [_
     (error '"never happens: even?/odd? always returns a right when receiving a right")]))

-- (list (false) (true) (false) (true) (false))
(example (map even? (list 1 2 3 4 5)))

-- (list (true) (false) (true) (false) (true))
(example (map odd? (list 1 2 3 4 5)))
```

gelisam:
Implement a `defuns` macro which allows the definition of mutually-recursive functions.

Fixes #160.


gelisam:
@sorawee, would you like to try reviewing this PR? I can walk you through the Klister syntax if you want, either via PR comments or a video call.

sorawee:
@gelisam thanks very much for the invitation! Unfortunately, I won't be free until this Saturday. I will definitely take a look at this later. But please proceed without my inputs.

gelisam:
> Unfortunately, I won't be free until this Saturday.

Take your time!

gelisam:
@sorawee , do you still plan to review this PR? No rush, just making sure you didn't forget :)

sorawee:
I see that added files are under the `examples` directory. Would it be better to rename them `prelude` or `lib` that conveys that they are not just demo/tests, but something that is encouraged to be used?

Obviously, if you think it makes more sense to do that in a separate PR, that works too!

examples/defuns.kl
```diff
@@ -0,0 +1,261 @@
...
+    -- (parse-def '[even? (n)
+    --              (if (= n 0) true (odd? (- n 1)))])
+    -- =>
+    -- (pair 'even?
+    --       '(lambda (n) (if (= n 0) true (odd? (- n 1)))))
+    (flet [parse-def (stx)
+           (syntax-case stx
+             [(list (fN-name args body))
+              (syntax-case args
+                [()
```
sorawee, line 58: Just an observation: coming from Racket, this seems a bit weird to me. In Racket, I would write either:

```
(match v
  [(nil) ...]
  [(cons x xs) ...])
```

or

```
(syntax-case v
  [() ...]
  [(x xs ...) ...])
```

It's unclear if `syntax-case` in Klister is meant to be more like `match` or `syntax-case` in Racket. 

If former, I would expect the first branch to be `(nil)`, so that pattern matching and constructor mirror each other. 

If latter, I would expect `cons` to not be there (or rather, if it's there, it would be a pattern matching against a list of length three, with `cons`, `first-arg`, and `more-args` as three pattern variables).

gelisam:
> I see that added files are under the examples directory. Would it be better to rename it to prelude or lib that conveys that these files are not just demo/tests, but something that is encouraged to be used?

For now, almost everything goes in `./examples`. #54 tracks the task of organizing which modules go in `#lang "prelude.kl"`, which modules go in `./stdlib`, and which modules should stay in `./examples`.

examples/defuns.kl
gelisam: Thanks for the suggestion, `(nil)` would indeed be clearer.

gelisam: Fixed in subsequent PR #169

gelisam:
I was waiting for @sorawee to do a second review after I addressed their comments, but I think I already took enough of their time. Merging!

sorawee:
Oops, sorry! I didn't you were waiting for a second review. Please merge it. It looks good to me.

Issue #101, opened 2020-08-26
Simpler primitives than `syntax-case`


david-christiansen:
Now that we have datatypes, I think we can do better than the built-in syntax-case operator. Here's a proposal:

First, we reflect our ExprF functor in the kernel language as the equivalent of:
```
(datatype (List A) ...)
(datatype (Syntax-Contents A)
  (list-syntax (List A))
  (integer-syntax Integer)
  (string-syntax String)
  (identifier String))
```
This requires making lists built-in like Bools.

Next, we can have two operations: `open-syntax : (-> Syntax (Syntax-Contents Syntax))` and `close-syntax : (-> Syntax Syntax (Syntax-Contents Syntax) Syntax)`. The former exposes one level down into the syntax object, and the latter takes some contents along with a source of scopes and a source of position information, creating a syntax object.

This is enough to implement something like our current `syntax-case`, but it also allows macros to manufacture identifiers (e.g. if we wanted to have a macro that created a datatype's signature functor with a `-F` suffix). It allows us to greatly simplify our evaluator and expander, because we replace builtins with primitive functions. And it ought to allow for some simpler destructuring patterns in some cases.

Thoughts?

david-christiansen:
One difficulty with this approach is that we risk inefficiency by treating lists as normal inductive types. This is because recursive functions over list syntax will end up calling `open-syntax`, analyzing the list, and then calling `close-syntax` prior to recursion. This means a quadratic overhead as we traverse the list to convert between Haskell `[Syntax]` and the corresponding Klister value objects.

I suppose it's reasonable enough to just support lists as a built-in primitive type rather than using the datatype mechanism.

david-christiansen:
For clarity, that would mean we'd have a value constructor
```
 | ValueList [Value]
```
to avoid the traversals.

gelisam:
Isn't this already implemented? We now have [a `Syntax-Contents` datatype](https://github.com/gelisam/klister/blob/c4c351e48ec086eeb0bbef8fe425f6cfc7b3cc54/stdlib/prelude.kl#L44), as well as [`open-syntax` and `close-syntax` functions](https://github.com/gelisam/klister/blob/c4c351e48ec086eeb0bbef8fe425f6cfc7b3cc54/stdlib/prelude.kl#L27-L27) to convert between `Syntax` and `Syntax-Contents`.

Shall we close this and open two new tickets for `ValueList` and for downgrading `syntax-case` from a builtin to a user-defined macro?

PR #169, opened 2022-10-09
rename () pattern to (nil) in syntax-case

gelisam:
Renames the `()` pattern to `(nil)` in `syntax-case`, in order to be consistent with the `(cons hd tl)` pattern.

Before:

    (syntax-case args
      [() ...]
      [(cons first-arg more-args))

After:

    (syntax-case args
      [(nil) ...]
      [(cons first-arg more-args))


gelisam:
This PR builds on top of #161, and therefore the diff will be needlessly large until #161 is merged. If you want to review before that happens, please start at [d6e411d](https://github.com/gelisam/klister/pull/169/commits/d6e411dd4cde0745069a11bed1dd445d2b46b7fd).

gelisam:
One slightly confusing consequence is that `nil` in a pattern position can mean the syntax object `'()` or the empty list `(nil)`, which have different types in Klister. See e.g. this function which converts a syntax object to a list:

https://github.com/gelisam/klister/blob/d6e411dd4cde0745069a11bed1dd445d2b46b7fd/examples/pmatch.kl#L29

There is no clash because technically, it is only the `nil` of type `(List A)` which is in scope, the `nil` of type `Syntax` is not a pattern-macro. Instead, `syntax-case` checks whether the pattern is an identifier named "nil". This means that it is possible to rename the `nil` of type `(List A)` to `empty-list`, but it is not possible to rename the `nil` of type `Syntax` to `empty-syntax`.

The issue will either disappear or become problematic in #101. It might disappear because `syntax-case` will no longer be a builtin, it will be a regular macro, and so it will be much easier for that `syntax-case` implementation to do the right thing and make `nil` a renameable pattern macro. It might become problematic because at that point, it will not be possible for `prelude.kl` to export both `nil` definitions with the same name.

In order to avoid this problem, maybe this PR should pick a different name, like `empty` or `nil-syntax`? While we're at it, should we also rename `cons` to `cons-syntax`? `cons` doesn't conflict with the `(List A)` constructor, because the `(List A)` constructor is called `::`. But as a reminder, the purpose of this PR is to make the `syntax-case` patterns more consistent, by using `nil` and `cons` instead of `()` and `cons`. So using `nil-syntax` and `cons` would again make the pattern names inconsistent.

gelisam:
Q qThinking about this more, let's just reuse the q

gelisam:
We already have `empty-list-syntax` and `cons-list-syntax`, so I guess it should be `empty`.

gelisam:
Looks like I accidentally pressed the close button. Reopening.

gelisam:
Thinking about this more, I think `empty-list-syntax` and `syntax-case` should be deprecated in favor of `(close-syntax (list-contents (nil)) stx)` and

    (match (open-syntax args)
      [(list-contents (nil)) ...]
      [(list-contents (:: first-arg more-args)) ...])

This would be more consistent, but also more verbose, than `(empty-list-syntax stx)` and

    (syntax-case args
      [() ...]
      [(cons first-arg more-args) ...])

The extra verbosity should be addressed separately, via the existing `quasiquote.kl` and #81. These provide much better syntax than the current `empty-list-syntax` and `syntax-case` builtins, whose syntax was chosen to be convenient to the implementation rather than convenient to the user.

david-christiansen:
This seems like a reasonable minor cosmetic change.

gelisam:
I no longer think this is a worthwhile change. Let's just get rid of `syntax-case` in favor of `open-syntax` and `close-syntax`, and then let's build a better `syntax-case` on top of this simpler foundation!

PR #172, opened 2022-10-29
bugfix: bind-IO only ran the first IO action

gelisam:
The issue was that given `bind-IO iox f`, we were basically executing

    do x <- iox
       pure (f x)

instead of

    do x <- iox
       f x

It's actually worse than that: we were returning the IO action returned by `f x` to a caller which expects a `b`, not an `IO b`. So this could easily have crashed the program, if we ever wrote a program which actually did something with the result of `bind-IO`.

In practice, however, most of our tests are about expressions, not IO actions, so we didn't catch this earlier. The one test which did use `bind-IO` was [io.kl](https://github.com/gelisam/klister/blob/main/examples/io.kl#L5), but it only checked that the result was inferred to have type IO Integer, it did not execute the resulting IO action.

### DIFF
--- a/examples/hello.golden
+++ b/examples/hello.golden
@@ -1,0 +1,2 @@
+hello
+world
--- a/examples/hello.kl
+++ b/examples/hello.kl
@@ -1,0 +1,16 @@
+#lang "prelude.kl"
+
+-- do notation is not builtin syntax, it's implemented as a library!
+(import "monad.kl")
+
+-- (-> String (IO Unit))
+(defun putStrLn (str)
+  (write stdout (string-append str "\n")))
+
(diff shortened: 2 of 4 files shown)


gelisam:
@Twisol, if you are looking for a way to get your feet wet with the codebase, perhaps you could review this pull request? It's pretty small, and I can answer any question you have about it or the surrounding code.

Twisol:
Thanks @gelisam; I'll take a look! Sorry I didn't acknowledge sooner -- I've had a little bit of a busy weekend.

gelisam:
@Twisol are you still planning to review this? No rush, I just want to know if I should wait or not :) Thanks!

PR #175, opened 2022-10-30
simplify and document flet implementation

gelisam:


stdlib/n-ary-app.kl
```diff
@@ -89,12 +96,6 @@
                        body)
                       stx))])])]))]))
 
-(define-macros
-  ([flet (lambda (stx)
-           (syntax-case stx
-             [(cons _ rest)
-              (pure (cons-list-syntax 'flet* rest stx))]))]))
-
```
gelisam, line 97: @david-christiansen , do you remember if there was a particular reason `flet*` was being renamed to `flet` by defining a new `flet` macro, instead of using the export-rename functionality as I do now?

david-christiansen: I'm pretty sure it's because this file predates export renaming and proper modules. No need to keep it this way!

david-christiansen (approved):

Issue #176, opened 2022-10-31
what's the purpose of `ident`?


gelisam:
`ident` is a builtin macro which is just like `quote`, except that it fails if its argument is anything other than an identifier.

```
(example (open-syntax (ident foo)))
-- =>
-- (identifier-contents "foo")

(example (open-syntax 'foo))
-- =>
-- (identifier-contents "foo")
```

There are no examples which use it. Can I delete it, or is it useful in a corner case which we haven't encountered yet?

david-christiansen:
I think that it's just there for providing better errors for things like custom let or lambda forms, but I really don't remember for sure. Or perhaps for symmetry with the syntax pattern forms?

In any case, if we don't use it, then it's not really tested and should probably go.

gelisam:
Removed by #177.

PR #177, opened 2022-11-01
remove unused 'oops' and 'ident' primitives

### LINKED ISSUE #173: is oops supposed to throw an _internal_ error?
`oops` is intended to behave like `undefined`; it indicates a bug in the Klister code which the user wrote, not in the implementation of Klister. And yet, the error message says otherwise:
```
(example oops)
-- =>
-- Internal error during expansion! This is a bug in the implementation.
-- oopsSyntax (Stx
  { _stxScopeSet = ...
  , _stxSrcLoc = ...
  , _stxValue = Id "oops"
  })
```

The error message is also quite hard to read because the elided scope set is typically very big.

gelisam:
Closes #173 and #176

### DIFF
--- a/tests/Test.hs
+++ b/tests/Test.hs
@@ -130,7 +130,7 @@ miniTests =
             \                {- [[lambda [x] body] e] -} \n\
             \                [pure [list-syntax \n\
             \                        [[list-syntax \n\
-            \                           [[ident lambda] \n\
+            \                           [[ident-syntax 'lambda stx] \n\
             \                            [list-syntax [x] stx] \n\
             \                            body] \n\
             \                           stx] \n\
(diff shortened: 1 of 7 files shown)


tests/Test.hs
```diff
@@ -130,7 +130,7 @@ miniTests =
             \                {- [[lambda [x] body] e] -} \n\
             \                [pure [list-syntax \n\
             \                        [[list-syntax \n\
-            \                           [[ident lambda] \n\
+            \                           [[ident-syntax 'lambda stx] \n\
```
gelisam, line 133: turns out there was one test which was using `ident` after all

tests/Test.hs
```diff
@@ -661,7 +661,6 @@ genCoreF subgen varGen =
         -- , CoreIdentEq _ <$> sameVars <*> sameVars
         -- , CoreSyntax Syntax
         -- , CoreCase sameVars [(Pattern, core)]
-        -- , CoreIdentifier Ident
         -- , CoreDataCase core [(ConstructorPattern, core)]
         -- , CoreIdent (ScopedIdent core)
```
gelisam, line 665: despite the name, `CoreIdent` is not the core representation of `ident`, it's the core representation of `ident-syntax`.

david-christiansen (approved):

PR #178, opened 2022-11-01
bare minimum documentation for everything in prelude.kl

gelisam:
When writing Klister code, I often peruse `stdlib/prelude.kl`'s export list, to see what are the functions and macros I have available. For many of them, their name is insufficient to tell me what they do.

Since Klister is typed, I would already get a lot of information about an identifier if I could see its type. For macros, which don't have types, I instead grep our existing code for example uses.

As a first step towards documenting the language, I have gathered type signatures and short examples for every single identifier in `prelude.kl`. This is only a first step though, there is still lot to improve:

1. nothing checks that new identifiers added to `prelude.kl` are also added to `primitives-documentation.kl`, so they are bound to get out of sync sooner or later.
2. similarly, the examples are not checked. It would be easy to check them using the `example` macro, but then that would pollute the `primitives-documentation.golden` output.
3. the documentation is split between the generated type signatures in `primitives-documentation.golden` and the manually-curated examples and type signatures in `primitives-documentation.kl`. It would be much better if this was a single file.
4. as the Haskell community has now learned, type signatures are not a sufficient form of documentation, some prose is needed too.

### DIFF
--- a/examples/primitives-documentation.golden
+++ b/examples/primitives-documentation.golden
@@ -1,0 +1,50 @@
+(pair "open-syntax" #<closure>) : (Pair String (Syntax → (Syntax-Contents Syntax)))
+(pair "close-syntax" #<closure>) : (Pair String (Syntax → (Syntax → ((Syntax-Contents Syntax) → Syntax))))
+(pair "+" #<closure>) : (Pair String (Integer → (Integer → Integer)))
+(pair "-" #<closure>) : (Pair String (Integer → (Integer → Integer)))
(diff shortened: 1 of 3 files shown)


gelisam:
The broken links in question are [list of examples](https://github.com/gelisam/klister/blob/main/examples/primitives-documentation.kl) and [list of type signatures](https://github.com/gelisam/klister/blob/main/examples/primitives-documentation.golden), in the Guide and Reference section.

PR #179, opened 2022-11-07
Rewrite the README: Klister is a programming language

gelisam:
The previous version didn't even explain that Klister is a programming language, this was very confusing for visitors unfamiliar with the project.

In the new version, I state right away that Klister is a programming language, and provide a short example program highlighting some of the language features. I foreshadows what makes Klister interesting, to keep the reader interested, without frontloading the document with all the background knowledge which would be necessary to understand stuck macros.

I then list the features of the language, starting with some easier-to-understand features which readers are familiar with from other languages. I would like each feature to link to some commented code example demonstrating the feature, but many of those links are still TODO.

Then, I finally list the features which make Klister special, but are a bit harder to understand because almost no other language has them.

Finally, to counter-balance all that positivity from a long list of features which Klister _does_ have, I help the reader decide if Klister is for them by admitting to some important features which Klister does _not_ have.

I moved the implementation details to a file called "HACKING.rst", as I believe this is the traditional name for a file explaining how the codebase is organized.


README.rst
```diff
@@ -1,102 +1,208 @@
...
-
-Usage
-=====
-
-Run ``klister repl`` to be dropped into a read-eval-print loop where you can try
-out writing some Klister expressions.
-
-``klister run file.kl`` will evaluate the examples and run the IO
-actions in ``file.kl``.
+Klister [`TyDe 2020`_, `video`_] is a research-stage programming language which combines
```
david-christiansen, line 4: I'd want to call it a "research prototype" here, because it's not really possible to write useful programs in the language outside of the focus (metaprogramming)

README.rst
```diff
@@ -1,102 +1,208 @@
...
+the core features of Racket and Haskell into a single language. It is named
```
david-christiansen, line 5: I'd talk about the ML family of languages here, rather than Haskell per se. We're much closer to old-school ML in the type system here, though we do use monads to sequence effects.

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  -- do notation is not builtin syntax, it's implemented as a library!
+  (import "monad.kl")
+
+  -- An effectful action whose inferred type is (-> String (IO Unit))
+  (defun putStrLn (str)
+    (write stdout (string-append str "\n")))
+
+  -- "run" is like main, except you can have more than one.
+  (run
+    -- Klister doesn't have typeclasses yet, so "do" needs an explicit
```
david-christiansen, line 26:
```suggestion
    -- Klister doesn't have type classes yet, so "do" needs an explicit
```
The original Wadler and Blott paper uses two words for "type class", and we should follow the names found by inventors of things unless we have good reason to abandon them

README.rst
```diff
@@ -1,102 +1,208 @@
...
+========
+
+Features we borrow from Racket:
+
+* `Custom syntax`_, via `hygienic macros`_ with `easy-to-override hygiene`_.
+* `Custom languages`_ (``#lang``), via macros which reinterpret terms into
+  those of an existing ``#lang``.
+* `Syntax objects`_, that is, s-expressions annotated with source locations and
+  lexical information.
+* A `module system`_ which allow imports and exports to be renamed.
```
david-christiansen, line 48: I think that it's even more important that modules respect the phase system, allowing compile-time and run-time dependencies to be separate (so e.g. SMT-based code generation in a macro would not result in a binary that needs the SMT solver)

README.rst
```diff
@@ -1,102 +1,208 @@
...
+.. _easy-to-override hygiene: examples/anaphoric-if.kl
+.. _Custom languages: examples/rpn.kl
+.. _Syntax objects: TODO: link to a short example which explains that in
+   Racket, syntax objects are introduced via ``#'(...)``, whereas in Klister
+   they are introduced via ``'(...)``. Also explain that Klister does not have
+   unannotated s-expressions.
+.. _module system: TODO: write a short example demonstrating what can be
+   expressed with the import and export primitives.
+
+Features we borrow from Haskell:
```
david-christiansen, line 61:
```suggestion
Features we borrow from the ML family of languages and Haskell:
```

README.rst
```diff
@@ -1,102 +1,208 @@
...
+
+* Statically-typed; we check that macro implementations have type
+  ``(-> Syntax (Macro Syntax)``, and *then* that the generated code is also
+  well-typed. We thus aim for the expressivity of Template Haskell, not the
+  extra guarantees of Typed Template Haskell.
+* Purely functional; primitives with side-effects either run in the ``IO``
+  monad or in the ``Macro`` monad, as appropriate.
```
david-christiansen, line 68: This distinguishes effects that are available at compile time, such as comparing identifiers modulo the current set of bindings, from effects that are available at run time, such as I/O.

README.rst
david-christiansen: (e.g. we don't support polymorphic recursion)

README.rst
```diff
@@ -1,102 +1,208 @@
...
+* Higher-kinded types; for example `monads`_ are defined as a library.
+* `Algebraic datatypes`_. In Racket, ``(list 1 2 3)`` and ``'(1 2 3)``
```
david-christiansen, line 70: This bullet point confuses me. Having algebraic datatypes is great, but what does that have to do with syntax objects? It seems to be mixing two separate points.

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  construct the same value, an s-expression which satisfies the ``list?``
+  predicate. In Klister, the former produces a value of the algebraic datatype
+  ``(List Integer)``, while the latter produces a value of type ``Syntax``.
+  This syntax can be "opened", via ``(open-syntax '(1 2 3))``, to obtain the
+  value ``(list-contents (list '1 '2 '3))`` of type
+  ``(Syntax-Contents Syntax)``.
+* `Bidirectional type inference`_; the type of an argument can specialize the
```
david-christiansen, line 77: This isn't what's usually meant by bidirectional type inference. We do Hindley-Milner style global type inference, but bidirectional typing usually refers to a particular assemblage of type system features rather than H-M.

README.rst
```diff
@@ -1,102 +1,208 @@
...
+.. _monads: TODO: link to monad.kl's Monad definition, and add a comment there
+   highlighting the inferred type, especially the higher-kinded type variable.
+.. _Algebraic datatypes: TODO: write a small example defining and matching on
+   an algebraic type. Perhaps Either?
+.. _Bidirectional type inference: TODO: write a small example demonstrating
+   those two features.
+
+Features which make Klister special (but not *unique*; see the `bibliography`_):
+
+* `Type-providing macros`_; a macro used in an argument position can provide
```
david-christiansen, line 91: I think that this isn't about argument positions vs function positions. Instead, I'd say that we have interleaved type checking and macro expansion, so the results of macro expansion can provide type information to the rest of the program, and macros can query the expected type.

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  of its arguments.
+
+.. _monads: TODO: link to monad.kl's Monad definition, and add a comment there
+   highlighting the inferred type, especially the higher-kinded type variable.
+.. _Algebraic datatypes: TODO: write a small example defining and matching on
+   an algebraic type. Perhaps Either?
+.. _Bidirectional type inference: TODO: write a small example demonstrating
+   those two features.
+
+Features which make Klister special (but not *unique*; see the `bibliography`_):
```
david-christiansen, line 89: Given the "not unique" comment, I'd expect a "see also" on each of these bullet points

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  type information about the argument, and can thus specialize the type of the
+  function being called.
+* `Type-aware macros`_; a macro used in an argument position can obtain the
+  type of the argument it needs to generate, and thus the type of a function
+  can affect what arguments are generated.
+* `Stuck macros`_; it is thus possible for multiple macros to affect what each
+  other generates. The language primitives are designed so that the order in
+  which the macros are expanded cannot affect their results, and indeed the
+  same is true for the order in which the macro expansion and type-inference
+  steps are interleaved. This makes Klister code more robust to refactoring.
```
david-christiansen, line 101:
```suggestion
  steps are interleaved. This means that the order in which the type checker traverses a program and generates constraints is not visible to the authors of macros, providing a predictable programming model.
```

README.rst
```diff
@@ -1,102 +1,208 @@
...
+* `Problem-aware macros`_; in addition to the type, a macro can learn which
```
david-christiansen, line 102: I'd say that each problem corresponds to a form of judgment in the specification of the language. Because each form of judgment regulates the validity of a particular syntactic construct (be it type, expression, declaration, etc), these two things correspond.

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  "problem" it needs to solve, namely whether it must generate an expression, a
+  type, a pattern, etc.
+* `Custom type-driven code generation`_, via macros which generate code from a
```
david-christiansen, line 105: How is this different from type-aware macros?

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  type.
+* Languages with `custom type systems`_, via macros which reinterpret types
+  into those of an existing ``#lang``, and which contribute to type inference
+  by providing type information about the code they generate. The variety of
+  type-systems which can be implemented this way is unforunately limited by
+  the core type system to which everything must be rewritten.
+* Languages with `custom implicit terms`_, via macros which generate terms of
```
david-christiansen, line 112: I think this is not a feature, but rather a use of the features, right?

README.rst
```diff
@@ -1,102 +1,208 @@
...
+  Lisp. This also limits languages to Integer literals and String literals.
+* `local-expand`_ is planned, but not yet implemented.
+* `Syntax parameters`_ are planned, but not yet implemented.
+
+.. _local-expand: https://github.com/gelisam/klister/issues/144#issuecomment-1133964551
+.. _Syntax parameters: https://github.com/gelisam/klister/issues/105
+
+Here are the most prominent Haskell features which are missing from Klister:
+
+* `Type classes` are planned, but not yet implemented.
```
david-christiansen, line 148:
```suggestion
* `Type classes` are planned as a library, but are not yet implemented.
```

README.rst
```diff
@@ -1,102 +1,208 @@
...
+* Klister is strict, not lazy. Racket has a lazy ``#lang``, so it is probably
```
david-christiansen, line 149: This was an intentional design choice, not a missing feature :-)

README.rst
```diff
@@ -1,102 +1,208 @@
...
+
+.. _Type classes: https://github.com/gelisam/klister/issues/167
+.. _Type annotations containing foralls: https://github.com/gelisam/klister/issues/60
+
+Here are the most prominent features which Racket and Haskell both have but
+which are missing from Klister:
+
+* Klister is missing commonly-expected datatypes like ``Map``, ``Set``, and
+  ``Double``.
+* Klister requires functions to be defined before they are used.
```
david-christiansen, line 164:
```suggestion
* Klister requires functions and datatypes to be defined before they are used.
```

examples/hello.kl
```diff
@@ -0,0 +1,16 @@
...
+-- do notation is not builtin syntax, it's implemented as a library!
+(import "monad.kl")
+
+-- (-> String (IO Unit))
+(defun putStrLn (str)
+  (write stdout (string-append str "\n")))
+
+-- "run" is like main, except you can have more than one.
+(run
+  -- Klister doesn't have typeclasses yet, so "do" needs an explicit
```
david-christiansen, line 12:
```suggestion
  -- Klister doesn't have type classes yet, so "do" needs an explicit
```

examples/rpn.kl
```diff
@@ -0,0 +1,135 @@
+#lang "prelude.kl"
```
david-christiansen, line 1: This is a really cool example!

david-christiansen (commented):
This is a great improvement to the README!

README.rst
gelisam: I really do want to mention Haskell early and prominently, because Haskell users are a key demographic for this project:

1. Klister is implemented in Haskell, so Haskellers are our most likely contributors
2. We're both involved in the Haskell community and thus are going to be sending a disproportionate amount of Haskellers to this README compared to visitors from other communities
3. The language to which Klister is the most similar is Hackett, which does position itself as a combination of Racket and Haskell.

Thus, I would prefer to mention Haskell early, and then to enumerate the differences which makes Klister closer to ML than to Haskell, than to start with ML and then enumerate the ways in which Klister inches towards Haskell.

What if I mentioned all three of Racket, ML, and Haskell in the first paragraph, would that be a good compromise?

README.rst
gelisam: clarified in 6e022cfb122154f550de81fce3478a28433e70f2

README.rst
gelisam: removed in 61ac33f984dbbde12ad80a346e4f8d9363908792

README.rst
gelisam: Between these first two bullet points and the two "Bidirectional type inference" bullet points, I was trying to draw a parallel between, on one hand, the fact that unification allows type information to flow both in and out of a term, and on the other hand, the fact that our macros can both provide and receive type information. That's why I over-focus on "two directions", and needlessly mention the argument position in order to tie the bullet points together. I now see that I stretched the wording way too much and that the payoff isn't worth it, so I'll rewrite both pairs of bullet points.

README.rst
gelisam: How about this into two separate sections: we borrow the type system from ML, and the rest from Haskell?

README.rst
gelisam: Hmm, can we keep both? I think the robustness to refactoring is a real and convincing advantage. With a type-aware macros which reveals whatever the type checker has inferred so far, you could easily end up in the following situation. The user writes a macro, and it just so happens that with the current task ordering, the macro has access to all the information it needs and the code works fine. Then, the user makes what seems like a benign refactoring, like moving the macro definition to a separate file or naming an expression with a `let`. This affects the task ordering, and now the macro sees type variables instead of the information it needed, so the code no longer works. That is, the code was not robust enough to survive the refactoring.

README.rst
gelisam: Good point, I'll move this bullet point and those which follow out of the feature list and into a separate section about things which can be achieved thanks to the features.

README.rst
gelisam: I tried my counter-suggestion in 9232b80a847b8c3b51fceeca5bca6c6ab48eeabc, let me know what you think!

README.rst
gelisam: fixed in 9232b80a847b8c3b51fceeca5bca6c6ab48eeabc

README.rst
gelisam: I tried my counter-suggestion in 9232b80a847b8c3b51fceeca5bca6c6ab48eeabc, let me know what you think!

README.rst
gelisam: fixed in 9232b80a847b8c3b51fceeca5bca6c6ab48eeabc

README.rst
gelisam: fixed in caaad94faa457e2204be892c7d4db8f5fc69620d

README.rst
david-christiansen: Looks good!

README.rst
david-christiansen: I think both is good here. I would have interpreted the earlier text as saying that programs in general are easier to refactor, rather than just metaprograms, but I think that combining these two texts will make that clear.

README.rst
david-christiansen: Suggestion: "The core language of Klister is very much like a strict Haskell 98, except it is missing the following features: "

README.rst
gelisam: fixed in bf9c31e3502a3e5f6b79ba79038422af6c6d6db9

README.rst
gelisam: fixed in a much simpler way in 36163623ecc29197331db773cc2af3c37f11ccbd

README.rst
gelisam: combined both in c8045298c7f7a94d4195cab17bfb14880ac546f3

README.rst
gelisam: I like the idea, but we didn't write down those judgments, so there currently isn't anything to correspond to. How about "would correspond to a form of judgment if the language was formalized, e.g. a typing judgment for the `expression` problem and a well-formed type judgment for the `type` problem"?

gelisam: I attempted this counter-proposal in fb2bd753dbe0cb2198a4236f9f182802b1464c9d, let me know what you think!

PR #180, opened 2022-12-05
reorganize the lambda-case explanation

gelisam:
There were so many comments that I couldn't see the forest for the trees. Instead of interleaving the comments and the code, the comments are now all at the bottom, referring to discreet numbers identifying the interesting lines of the code.

### DIFF
--- a/examples/lambda-case.kl
+++ b/examples/lambda-case.kl
@@ -22,73 +22,91 @@
 -- lambda-case.
 
 (define-macros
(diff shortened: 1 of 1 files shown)


gelisam:
I should probably mention that the reason the comments are so detailed is because this is the first code link in the README, so this is the file which teaches visitors how to read Klister code. Depending on their background, it might not yet be detailed enough!

gelisam:
Merging without review, as I don't think small documentation improvements need that much scrutiny at this stage.

PR #181, opened 2022-12-07
do not document hygiene (was: add copious comments to the hygiene example)

gelisam:
I'm on a quest to make Klister more approachable by having the README link to examples of each feature, with comments explaining what's going on.


tests/Test.hs
```diff
@@ -295,32 +295,6 @@ moduleTests = testGroup "Module tests" [ shouldWork, shouldn'tWork ]
...
-                      assertAlphaEq "Ex 2 picks fun 1" (Core (CoreVar fun1)) f
-                    other -> assertFailure $ "Ex 2 should be an application, but it's " ++ shortShow other
-                  spec3 <- lam (const (pure (Core (CoreVar fun2))))
-                  case e3 of
-                    Core (CoreApp (Core (CoreApp (Core (CoreApp f _)) _)) _) ->
-                      assertAlphaEq "Ex 3 picks fun 2" f spec3
-                    other -> assertFailure $ "Ex 3 should be an application, but it's " ++ shortShow other
-
-                _ -> assertFailure "Expected two imports, a def, a macro, a def, two examples, a macro, and an example"
-          )
```
gelisam, line 323: @david-christiansen this test is no longer passing, because I added an import and it expects exactly 2 imports. I would like to delete the test instead of fixing it, because it's harder to maintain than the golden tests, and I think it's just a leftover from when we didn't yet have the golden test system in place. Do you agree? If so, should I take the opportunity to delete more of those non-golden tests?

One reason I hesitate is that this test tests something different from the corresponding golden test: it doesn't check that the program's output is `'b`, `'a`, and `'b`, as we'd expect if hygiene is respected, instead it checks that the fully-expanded code is as expected, which is a lot more precise.

examples/hygiene.kl
```diff
@@ -1,23 +1,64 @@
...
+-- second argument.
+--
+-- Between the two definition of "fun", a macro "should-not-be-captured" is
+-- defined. It is a very simple macro which always expands to the symbol 'fun.
+-- Thus, line (2) expands to (fun 'a 'b), same as line (1). And yet, thanks to
+-- hygiene, this time the symbol "fun" refers to the first definition, the one
+-- which was in scope when the macro was defined! This allows the user to
+-- redefine a common identifier such as "case" or "lambda" without accidentally
+-- changing the semantics of all the macros macros which produce code
+-- containing those symbols.
```
gelisam, line 38: Here and at the very end of the file, I hint at more practical situations in which hygiene is useful. Perhaps I should just use those practical situations directly as the example code?

The reason I hesitate to make this change is that the existing examples are elegantly minimal. I don't know if they are minimal because Klister was so hard to use back then that being minimal was necessary in order to keep the code short, for pedagogical reasons, or because these are standard examples in the hygiene litterature.

gelisam:
I have now changed my mind and this PR is now doing the exact opposite of what it did originally: instead of adding a copious amount of comments explaining hygiene, the word "hygienic macros" no longer links anywhere, the README now just assumes that people know how to google terms they don't understand. This allows the guide to focus on Klister itself.

The "[Custom syntax](https://github.com/gelisam/klister/blob/main/examples/lambda-case.kl)" link can stay, because it doesn't just repeat the an explanation of what a macro is from somewhere else, it also shows what macros look like _in Klister_. The comments should probably be much shorter though.

The "[easy-to-override hygiene](https://github.com/gelisam/klister/blob/main/examples/anaphoric-if.kl)" link can stay, because it specifically shows what is _Klister's_ syntax for overriding hygiene.

The "[Custom languages](https://github.com/gelisam/klister/blob/main/examples/rpn.kl)" link can stay, because it specifically shows how to build `#lang`s _in Klister_.

The "Syntax objects" explanation (currently TODO) should remain because it will explain what Syntax objects are _in Klister_ and how they differ from Racket's.

I think the "module system" and "phase system" explanations (currently TODOs) should probably be merged into a single example explaining `meta`, phases, and Klister's syntax for imports and exports.

I think the "parametric polymorphism", "algebraic datatypes", and "type inference" explanations (currently TODOs) should also be merged into a single example demonstrating the syntax for defining datatypes and polymorphic functions.

The "monad" explanation (currently TODO) should remain because it shows how we do monads _in Klister_, namely via explicit dictionary passing.

The "Type-providing macros", "Type-aware macros", "Stuck macros", and "Problem-aware macros" explanations (currently TODOs) should of course stay because they are from the "Features which make Klister special" section.

tests/Test.hs
gelisam: Never mind, I've changed my mind about this PR, I no longer want to delete this test.

examples/hygiene.kl
gelisam: Never mind, I have changed my mind about this PR, I no longer want to explain why hygiene is useful.

gelisam:
My original PR would have warranted a review, but not what it became. Merging.

PR #182, opened 2022-12-08
shorter explanation for lambda-case

gelisam:
Let's not waste time explaining what a macro is, let's focus on teaching what macros look like in Klister.


gelisam:
This PR shortens the text of the "Custom syntax" link, as per [my plan](https://github.com/gelisam/klister/pull/181#issuecomment-1340331105).

gelisam:
I don't think documentation PRs need a review at this stage, merging.

PR #192, opened 2023-01-02
partially applied primitive types

### LINKED ISSUE #190: kind-inference bug with primitive type constructors
Using `IO`, `List`, or `Maybe` at kind `* -> *` results in a kind error: 

```
#lang "prelude.kl"
(import "list.kl")

(datatype (OfUnit F)
  (of-unit
    (F Unit)))


-- of-unit-io : (OfUnit IO)
(define of-unit-io
  (of-unit (pure-IO unit)))

-- Kind mismatch at IO: * ≠ (* → *)
(example
  (the (OfUnit IO)
       of-unit-io))


-- of-unit-list : (OfUnit List)
(define of-unit-list
  (of-unit (list unit)))
(example of-unit-list)

-- Kind mismatch at List: * ≠ (* → *)
(example
  (the (OfUnit List)
       of-unit-list))


-- of-unit-maybe : (OfUnit Maybe)
(define of-unit-maybe
  (of-unit (just unit)))
(example of-unit-maybe)

-- Kind mismatch at Maybe: * ≠ (* → *)
(example
  (the (OfUnit (Maybe))
       of-unit-maybe))


-- works fine with non-primitive datatypes
(datatype (MyMaybe A)
  (my-nothing)
  (my-just A))

-- of-unit-my-maybe : (OfUnit (MyMaybe))
(define of-unit-my-maybe
  (of-unit (my-just unit)))
(example of-unit-my-maybe)

(example
  (the (OfUnit (MyMaybe))
       of-unit-my-maybe))
```

gelisam:
Fixes #190

### DIFF
--- a/examples/higher-kinded.golden
+++ b/examples/higher-kinded.golden
@@ -3,4 +3,5 @@
 (of-unit (just (unit))) : (OfUnit Maybe)
 (of-unit (pair (unit) (unit))) : (OfUnit (Pair Unit))
 (of-unit-unit #<closure>) : (OfUnitUnit (→))
+(of-unit-unit #<closure>) : (OfUnitUnit (→))
 (of-unit-unit (pair (unit) (unit))) : (OfUnitUnit Pair)
--- a/examples/higher-kinded.golden
+++ b/examples/higher-kinded.golden
@@ -1,1 +1,2 @@
 (of-unit #<IO action>) : (OfUnit IO)
+(of-unit (just (unit))) : (OfUnit Maybe)
--- a/examples/higher-kinded.golden
+++ b/examples/higher-kinded.golden
@@ -1,2 +1,6 @@
 (of-unit #<IO action>) : (OfUnit IO)
+(of-unit #<closure>) : (OfUnit ((→) Unit))
 (of-unit (just (unit))) : (OfUnit Maybe)
+(of-unit (pair (unit) (unit))) : (OfUnit (Pair Unit))
+(of-unit-unit #<closure>) : (OfUnitUnit (→))
+(of-unit-unit (pair (unit) (unit))) : (OfUnitUnit Pair)
--- a/examples/higher-kinded.golden
+++ b/examples/higher-kinded.golden
@@ -1,0 +1,1 @@
+(of-unit #<IO action>) : (OfUnit IO)
--- a/examples/higher-kinded.golden
+++ b/examples/higher-kinded.golden
@@ -1,5 +1,5 @@
 (of-unit #<IO action>) : (OfUnit IO)
-(of-unit #<closure>) : (OfUnit ((→) Unit))
+(of-unit #<closure>) : (OfUnit (→ Unit))
 (of-unit (just (unit))) : (OfUnit Maybe)
 (of-unit (pair (unit) (unit))) : (OfUnit (Pair Unit))
 (of-unit-unit #<closure>) : (OfUnitUnit (→))
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -688,22 +688,24 @@ initializeKernel outputChannel = do
                  , _datatypeName = dn
                  }
       let tyImpl =
-             \k dest stx -> do
-               equateKinds stx KStar k
-               Stx _ _ (me, args) <- mustBeCons stx
-               _ <- mustBeIdent me
-               if length args /= length argKinds
-                 then throwError $ WrongDatatypeArity stx dt
-                                     (fromIntegral $ length argKinds)
-                                     (length args)
-                 else do
-                   argDests <- traverse (uncurry scheduleType) (zip argKinds args)
-                   linkType dest $ tDatatype dt argDests
+            \k dest stx -> do
+              Stx _ _ (me, args) <- mustBeCons stx
+              _ <- mustBeIdent me
+              if length args > length argKinds
+                then throwError $ WrongDatatypeArity stx dt
+                                    (fromIntegral $ length argKinds)
+                                    (length args)
+                else do
+                  let missingArgs :: [Kind]
+                      missingArgs = drop (length args) argKinds
+                  equateKinds stx (kFun missingArgs KStar) k
+                  argDests <- traverse (uncurry scheduleType) (zip argKinds args)
+                  linkType dest $ tDatatype dt argDests
           patImpl =
             \dest stx -> do
               Stx _ _ (me, args) <- mustBeCons stx
               _ <- mustBeIdent me
-              if length args /= length argKinds
+              if length args > length argKinds
                 then throwError $ WrongDatatypeArity stx dt
                                     (fromIntegral $ length argKinds)
                                     (length args)
--- a/src/Expander/Error.hs
+++ b/src/Expander/Error.hs
@@ -34,7 +34,7 @@ data ExpansionErr
   | NotInteger Syntax
   | NotString Syntax
   | NotModName Syntax
-  | NotRightLength Natural Syntax
+  | NotRightLength [Natural] Syntax
   | NotVec Syntax
   | NotImportSpec Syntax
   | NotExportSpec Syntax
@@ -59,6 +59,9 @@ data ExpansionErr
   | KindMismatch (Maybe SrcLoc) Kind Kind
   | CircularImports ModuleName [ModuleName]
   deriving (Show)
+
+notRightLength :: Natural -> Syntax -> ExpansionErr
+notRightLength n = NotRightLength [n]
 
 data TypeCheckError
   = TypeMismatch (Maybe SrcLoc) Ty Ty (Maybe (Ty, Ty))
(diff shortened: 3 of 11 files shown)


src/Expander.hs
```diff
@@ -688,22 +688,24 @@ initializeKernel outputChannel = do
...
-                 then throwError $ WrongDatatypeArity stx dt
-                                     (fromIntegral $ length argKinds)
-                                     (length args)
-                 else do
-                   argDests <- traverse (uncurry scheduleType) (zip argKinds args)
-                   linkType dest $ tDatatype dt argDests
+            \k dest stx -> do
+              Stx _ _ (me, args) <- mustBeCons stx
+              _ <- mustBeIdent me
+              if length args > length argKinds
```
gelisam, line 694: Now accepting the case in which fewer arguments are provided than required, e.g. `(Either (Unit))` instead of `(Either (Unit) (Unit))`.

src/Expander.hs
```diff
@@ -688,22 +688,24 @@ initializeKernel outputChannel = do
...
+                then throwError $ WrongDatatypeArity stx dt
+                                    (fromIntegral $ length argKinds)
+                                    (length args)
+                else do
+                  let missingArgs :: [Kind]
+                      missingArgs = drop (length args) argKinds
+                  equateKinds stx (kFun missingArgs KStar) k
+                  argDests <- traverse (uncurry scheduleType) (zip argKinds args)
```
gelisam, line 702: The `zip` now drops the last few entries in `argKinds`, those for which the user did not provide a type argument. Those dropped entries are now in `missingArgs`.

src/Expander/Error.hs
```diff
@@ -34,7 +34,7 @@ data ExpansionErr
   | NotInteger Syntax
   | NotString Syntax
   | NotModName Syntax
-  | NotRightLength Natural Syntax
+  | NotRightLength [Natural] Syntax
```
gelisam, line 37: Generalizing the "expected a list of length 3" message to "expected a list of length 1, 2, or 3".

src/Expander/Primitives.hs
```diff
@@ -406,7 +406,7 @@ listSyntax :: ExprPrim
 listSyntax t dest stx = do
   unify dest t tSyntax
   Stx _ _ (_, list, source) <- mustHaveEntries stx
-  Stx _ _ listItems <- mustHaveEntries list
+  listItems <- mustHaveShape list
```
gelisam, line 409: This was the only place where we were using `mustHaveEntries` to assert that a Syntax is a List. I have dropped support for this feature, add is it already supported by `mustHaveShape`.

`mustHaveEnteies` now only supports checking that it is a List with a required number of elements, but I have generalized `mustHaveEntries` so that this required number can be "Either N or M" and not just a single number N.

src/Expander/Primitives.hs
```diff
@@ -486,7 +486,7 @@ log t dest stx = do
 whichProblem :: ExprPrim
 whichProblem t dest stx = do
   unify dest (tMacro (primitiveDatatype "Problem" [])) t
-  Stx _ _ (_ :: Syntax) <- mustHaveEntries stx
+  Stx _ _ (Identity _) <- mustHaveEntries stx
```
gelisam, line 489: I have changed the `mustHaveEntries` API so that a singleton tuple is now represented via an `Identity` wrapper and not just by the element itself. One nice consequence of that is that a type annotation is no longer needed in order to drive the type class resolution algorithm.

src/Expander/Syntax.hs
```diff
@@ -59,38 +64,72 @@ mustBeModName (Syntax (Stx scs srcloc (Id "kernel"))) =
...
+      throwError (NotRightLength lengths stx)
+mustHaveEntries other = do
+  throwError (NotList other)
+
+class FixedLengthList item r where
+  checkLength :: [item] -> Either [Natural] r
+
+instance ( FixedLengthList e1 a
+         , FixedLengthList e2 b
+         , e1 ~ e2
```
gelisam, line 86: hmm, I don't think that's right, I don't need to use two different type variables `e1` and `e2` here.

src/Expander.hs
david-christiansen: Why is it that these types explicitly need to check for too many arguments? Wouldn't the kind system figure that part out? (Basically, isn't `WrongDatatypeArity` a thing made obsolete by the introduction of kinds other than `*`?)

src/Expander.hs
david-christiansen: Nice!

src/Expander/Primitives.hs
david-christiansen: This seems like a nice change.

src/Expander/Primitives.hs
david-christiansen: Sounds reasonable!

stdlib/n-ary-app.kl
```diff
@@ -105,10 +105,17 @@
                     [(list (arg1 arg2))
                      (pure (list-syntax ('-> arg1 arg2)
                                         stx))]
+                    [(list (arg1))
+                     (pure (list-syntax ('-> arg1)
+                                        stx))]
```
gelisam, line 110: The kernel now supports 0, 1, or 2 arguments (`(->)`, `(-> A)`, or `(-> A B)`). The prelude's n-ary wrapper was only supporting 2 or more arguments. This extends the wrapper to also support 1 argument.

src/Expander/Primitives.hs
```diff
@@ -529,41 +529,73 @@ arrowType :: TypePrim
...
 macroType :: TypePrim
-macroType = unaryType (\a -> tMacro a)
+macroType = unaryType TMacro
 
 ioType :: TypePrim
-ioType = unaryType tIO
+ioType = unaryType TIO
 
-unaryType :: (forall a . a -> TyF a) -> TypePrim
+unaryType :: TypeConstructor -> TypePrim
```
david-christiansen, line 573: This is exactly what we'd expect it to have with proper kinds. But I wonder whether we should have `unaryType` instead of something like `typeConstructor :: TypeConstructor -> Kind -> TypePrim`. This would move some errors from being syntax errors to being kind errors, but that seems reasonable enough to me. Am I missing something?

stdlib/n-ary-app.kl
```diff
@@ -105,10 +105,17 @@
...
                     [(cons arg1 args)
                      (pure (list-syntax ('-> arg1
                                           (cons-list-syntax 'my-arrow args stx))
-                                        stx))])]))]))
+                                        stx))]
+                    [()
+                     (pure '(->))])]
+                 [_
+                  (pure '(->))]))]))
```
gelisam, line 118: This extends the wrapper to support a syntax which the kernel does not support, namely `->` without parentheses. Hmm, I should probably change `higher-kinded.kl` to test both `->` and `(->)`...

src/Expander.hs
gelisam: Oh nice, I didn't realize that check was no longer necessary!

src/Expander/Primitives.hs
david-christiansen: Then the same implementation could be used for declared datatypes as for builtins like IO, increasing the robustness of the system.

examples/higher-kinded.golden
```diff
@@ -0,0 +1,6 @@
+(of-unit #<IO action>) : (OfUnit IO)
+(of-unit #<closure>) : (OfUnit ((→) Unit))
```
david-christiansen, line 2: Is this a bug in the type pretty printer? I'd expect it to say `(OfUnit (→ Unit))` here.

src/Expander.hs
david-christiansen: Well, it's only no longer necessary if there's not a missing check in the kind checker :-) It's worth testing that case if it's not already there.

src/Expander/Primitives.hs
gelisam: Unifying our series of specialized `plainType`, `unaryType` and `binaryType` functions into a single `typeConstructor` function does seem like a good idea, thanks!

examples/higher-kinded.golden
gelisam: In core, the type `(-> A B C)` is printed as `A → (B → C)`, so I think other occurrences of `→` are intentionally printed as `(→)` in order to make sure `Monad (→) Unit` doesn't accidentally get printed as `Monad → Unit`, as that would be confusing.

When `→` has arguments, though, I agree it makes sense to print those arguments inside the parens.

gelisam: Thinking about this more, it's only the cases in which `→` has zero or one arguments which are a special case:

```
(→)
(→ A)
A → B
A → (B → C)
```

The current code accounts for the zero case, but not the one case.

src/Expander.hs
gelisam: Alas, without this `if length args > length argKinds` check, the following is accepted. I have just created #194 to track this.

```
-- (just unit) : (Maybe Unit) 
(example
  (the (Maybe Unit Unit)
       (just unit)))
```

gelisam: What about the similar check a few lines below, for type patterns? A value of type `Type` could hold a type of a kind other than `*` (as demonstrated by the variable named `type` in this PR's `higher-kinded-patterns.kl` example, which holds a type of kind `* -> *`), so the kind checker doesn't have much information with which to perform its check.

src/Expander/Syntax.hs
gelisam: Fixed in 2ca2a30a4a731068974fb7c21f75d424b0a69bea, and this change also removes the need for `UndecidableInstances` in this file.

stdlib/n-ary-app.kl
gelisam: Tested in 2847b4e3f2072c9f6c7d24c50bcf3fe8d2fd3812.

examples/higher-kinded.golden
gelisam: Fixed in e02824619e25f08ba2a2433ae38f185671218a84.

src/Expander/Primitives.hs
gelisam: That one is a bit more involved, I'll do it in a separate PR

gelisam: Created #195 to track this, so I don't forget

src/Expander.hs
david-christiansen: It seems plausible to me that type patterns should be well-kinded, and that this should be caught. What I'm less sure of is whether they should all have kind *, or whether it makes sense to match against other types during expansion.

src/Expander/Syntax.hs
david-christiansen: :confetti_ball:

examples/higher-kinded.golden
david-christiansen: Nice

david-christiansen (approved):

src/Expander.hs
gelisam: I have created #196 to continue the discussion on type patterns after I merge this PR.

Issue #196, opened 2023-01-04
kind-checking for type patterns


gelisam:
Similar to #194, but for type patterns: currently, type patterns emit a _syntax_ error if they are called on too many arguments, but this should be caught by the kind checker instead. One difficulty is that a value of type `Type` can currently hold a type of kind `*`, or a type of kind `* -> *`, etc. Thus, when kind-checking a type pattern which matches on a `Type` value, the kind-checker cannot know which kind it is supposed to expect. 

Originally discussed [here](https://github.com/gelisam/klister/pull/192#discussion_r1060252123).

gelisam:
One possible solution is change `Type`, to make it indexed by its kind.

gelisam:
Another solution is to only check that a type pattern is internally-consistent, without propagating any kind information via the variables of type `Type`. That is, the following should be accepted by the kind-checker, even though no type pattern will match:

```
(type-case (Maybe)
  [(Maybe (Integer)) ...]
  [(IO (Unit)) ...])

(type-case (Maybe Integer)
  [(Maybe) ...]
  [(IO) ...])
```

While the following would be rejected:

```
(type-case (Maybe)
  [(Maybe (Maybe)) ...]
  [(IO (Unit) (Unit)) ...])
```

One problem with this idea is that the kernel does not support nested type patterns, so there isn't much internal consistency to be checked. A type pattern like `(IO A B)` which has too many arguments would be rejected, and that's it. Implementing nested type patterns in the prelude would _not_ solve the problem because that implementation would necessarily rely on intermediate variables of type `Type`, through which kind information is lost.

gelisam:
One possible solution would be to add support for nested type patterns in the kernel.

gelisam:
Another possible solution would be for the kind-checker to make a rudimentary attempt to track the kind of variables of type `Type`. It would still be possible for the kind information to be lost if the program manipulates those variables in a complex way, e.g. by storing them in a datatype and extracting them later, but for something as simple as the following, it seems easy to keep track of the kind of `maybe?`, `unit1?`, and `unit2?`.

```
(type-case (Maybe)
  [(Maybe maybe?)
   (type-case maybe?
     [(Maybe) ...])]
  [(IO unit1? unit2?)
   (type-case unit1?
     [(Unit)
      (type-case unit2?
        [(Unit) ...])])])
```

gelisam:
[...]

PR #197, opened 2023-01-05
combine type constructors

### LINKED ISSUE #195: Combine baseType, unaryType, and addDatatypePrimitive
It should be possible to combine them into a single function which takes a TypeConstructor and a Kind.

Previously discussed [here](https://github.com/gelisam/klister/pull/192#discussion_r1060152725).

gelisam:
Fixes #195


examples/fix.kl
```diff
@@ -24,7 +24,7 @@
     length))
 
 
-(define l1 (the (Fix (ListF String)) (in (cons "hello" (in (cons "world" (in (nil))))))))
+(define l1 (the (Fix (ListF (String))) (in (cons "hello" (in (cons "world" (in (nil))))))))
```
gelisam, line 27: Types with zero type arguments such as `String` are now spelled `(String)`, just like user-defined datatypes with no type arguments. But only in the kernel; in the prelude #lang, every type is wrapped in a macro which makes the parens optional.

src/Expander/Primitives.hs
```diff
@@ -45,25 +45,22 @@ module Expander.Primitives
...
-  , ioType
-  , primitiveDatatype
   -- * Pattern primitives
   , elsePattern
   -- * Module primitives
   , makeModule
   -- * Anywhere primitives
   , makeLocalType
-  -- * Local primitives
-  , prepareVar
```
gelisam, line 61: I don't know what's a "local primitive", but `prepareVar` isn't exposed to the kernel, so I wouldn't call it a primitive.

src/Expander/Primitives.hs
```diff
@@ -512,90 +509,41 @@ typeCase t dest stx = do
 type TypePrim =
   (Kind -> SplitTypePtr -> Syntax -> Expand (), TypePatternPtr -> Syntax -> Expand ())
 
-baseType :: (forall a . TyF a) -> TypePrim
-baseType ctor = (implT, implP)
-  where
-    implT k dest stx = do
-      equateKinds stx KStar k
-      _actualName <- mustBeIdent stx
```
gelisam, line 520: We are no longer checking that the `stx` must be an identifier; on the contrary, we are now checking that it's a singleton list.

src/Expander/Primitives.hs
```diff
@@ -512,90 +509,41 @@ typeCase t dest stx = do
...
-          (sc2, n2, x2) <- prepareVar ret
+      Stx _ _ (_, args) <- mustBeCons stx
+      if length args > length argKinds
+        then throwError $ WrongTypeArity stx ctor
+                            (fromIntegral $ length argKinds)
+                            (length args)
+        else do
+          varInfo <- traverse prepareVar args
           sch <- trivialScheme tType
+          -- FIXME kind check here
```
gelisam, line 536: This `FIXME` comes from `addDatatype`. Presumably this is where we will implement #196.

src/Expander/Primitives.hs
```diff
@@ -675,27 +623,8 @@ elsePattern (Right dest) stx = do
 addDatatype :: Ident -> Datatype -> [Kind] -> Expand ()
 addDatatype name dt argKinds = do
   name' <- addRootScope' name
-  let implType =
-        \k dest stx -> do
-          Stx _ _ (me, args) <- mustBeCons stx
-          _ <- mustBeIdent me
```
gelisam, line 681: We are no longer checking that the list is identifier-headed. Seems redundant to me; how would we get here if not because we are expanding an identifier-headed list whose head identifier names a datatype?

stdlib/builtins.kl
```diff
@@ -1,32 +0,0 @@
-#lang kernel
```
gelisam, line 1: renamed to `optional-parens.kl`

stdlib/optional-parens.kl
```diff
@@ -0,0 +1,74 @@
+#lang kernel
```
gelisam, line 1: renamed from `builtins.kl`, because

1. we're calling them "primitives" everywhere else
2. this file is more concerned with making parens optional around some primitives than with exporting all the primitives.

gelisam: In addition to the rename, this file now makes parentheses optional around many more primitives, most notably types with no type arguments such as `String`.

examples/fix.kl
david-christiansen: Sounds good. Should they be optional or forbidden?

src/Expander/Primitives.hs
david-christiansen: I think that this was overzealous parsing and was never needed

david-christiansen (approved):

examples/fix.kl
gelisam: `optional-parens` makes them optional rather than mandatory because of data constructors like `nil`, which definitely need parentheses when used as a pattern but not when used as an expression. Currently, types don't suffer from that issue because `type-case` does not yet support recursive patterns. But one day `type-case` will support recursive patterns, and on that day we will be glad that we only made parentheses optional, not forbidden :)

PR #199, opened 2023-02-19
explain Ident vs Binding vs Var

gelisam:
I recently struggled to understand the difference between `Ident`, `Binding`, and `Var`, so I took a moment to understand and to add an explanation to the commentary.

### DIFF
--- a/commentary.md
+++ b/commentary.md
@@ -84,15 +84,39 @@ Eventually the `Syntax` gets simplified into a core language which we can evalua
 
 `Core` expressions consist of function applications, pattern-matching, and constructors for the language's primitive types. Those primitive types include functions, booleans, signals, reified `Syntax` objects, and "macro actions". A macro action is a monadic block performing zero or more side-effects. These side-effects do not occur during evaluation, but rather during expansion, so we'll cover macro actions in more details in the `Expanding` section.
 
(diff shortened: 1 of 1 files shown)


commentary.md
```diff
@@ -84,15 +84,39 @@ Eventually the `Syntax` gets simplified into a core language which we can evalua
 
 `Core` expressions consist of function applications, pattern-matching, and constructors for the language's primitive types. Those primitive types include functions, booleans, signals, reified `Syntax` objects, and "macro actions". A macro action is a monadic block performing zero or more side-effects. These side-effects do not occur during evaluation, but rather during expansion, so we'll cover macro actions in more details in the `Expanding` section.
 
-TODO: what happened to strings? `Syntax` supports booleans, strings, and signals, but of those, `Core` only supports booleans and signals. Maybe we're instead supposed to manipulate a reified `Syntax` object holding a string?
```
gelisam, line 87: This TODO about strings has nothing to do with the rest of the changes, it's just an outdated TODO, as Core _does_ have a representation for strings now. There are also a bunch of places in the commentary which still refer to signals, but I haven't fixed them yet.

gelisam:
(force-pushed to remove a minor whitespace issue. I should probably have pushed a `typo` commit instead)

Issue #204, opened 2023-02-21
Convert the old test suite to golden files


gelisam:
`./tests/Test.hs` runs three kinds of tests:

1. In "Expander tests", we expand some programs, we match on the resulting AST, and we check that the results are as expected.
2. In "Hedgehog tests", we run some property tests.
3. In "Golden tests", we run all the files in the `examples` folder, and we check that the results are as expected in the corresponding `.golden` file.

The golden tests are much more convenient than the expander tests, because the `.golden` file is created automatically if it is missing. Most of the expander tests, which were written before we had golden tests, can and should be replaced by golden tests. Currently, we only bother migrating an expander test to a golden test when the Klister implementation changes in a way which causes an expander test to fail. It would be nice to get rid of all of them once and for all.

PR #215, opened 2023-04-26
First pass performance engineering

doyougnu:
Hey David and Sam,

I've made a few performance tweaks in the expander. The overall result is that the testsuite now runs in about 10 seconds instead of 38 seconds and `implicit-conversion-test` finishes in 1 second compared to 10 seconds.

I wrote a case study on it for my book here: https://input-output-hk.github.io/hs-opt-handbook.github.io/src/Case_Studies/klister.html

please review it at your leisure. @david-christiansen I tried to request a review on github but that unfortunately requires you to be part of the IOG organization, so I went ahead and merged it. I've re-opened the IT ticket to move the book.

Please do not merge this before I squash. I like to keep a clean commit history.

### DIFF
--- a/src/Env.hs
+++ b/src/Env.hs
@@ -1,53 +1,65 @@
-{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveFunctor      #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilies       #-}
+{-# LANGUAGE DerivingStrategies #-}
+
 module Env (Env, empty, insert, singleton, lookup, lookupIdent, lookupVal, toList, fromList, named) where
 
 import Prelude hiding (lookup)
 
 import Control.Lens
-import Data.Map (Map)
-import qualified Data.Map as Map
+import Data.IntMap.Strict (IntMap)
+import qualified Data.IntMap.Strict as IM
 import Data.Text (Text)
 
 import Syntax (Ident, Stx(..))
 
-newtype Env v a = Env (Map v (Ident, a))
-  deriving (Eq, Functor, Monoid, Semigroup, Show)
+import Util.Key
+
+-- | An environment maps variables of type 'v' to values of type 'a'.
+--
+-- More specifically, 'Env' is intended to be used for variables in the core
+-- language, such as 'Var' and 'MacroVar', which are represented as unique
+-- integers in order to avoid accidental capture. To make error message more
+-- readable, 'Env' stores the original 'Ident' in addition to the value.
+newtype Env v a = Env (IntMap (Ident, a))
+  deriving newtype (Eq, Monoid, Semigroup, Show)
+  deriving stock Functor
 
 empty :: Env v a
-empty = Env (Map.empty)
+empty = Env mempty
 
-toList :: Env v a -> [(v, Ident, a)]
-toList (Env env) = [(x, n, v) | (x, (n, v)) <- Map.toList env]
+toList :: HasKey v => Env v a -> [(v, Ident, a)]
+toList (Env env) = [(fromKey x, n, v) | (x, (n, v)) <- IM.toList env]
 
-fromList :: Ord v => [(v, Ident, a)] -> Env v a
-fromList vars = Env (Map.fromList [(x, (n, v)) | (x, n, v) <- vars])
+fromList :: HasKey v => [(v, Ident, a)] -> Env v a
+fromList vars = Env (IM.fromList [(getKey x, (n, v)) | (x, n, v) <- vars])
 
-singleton :: v -> Ident -> a -> Env v a
-singleton x n v = Env (Map.singleton x (n, v))
+singleton :: HasKey v => v -> Ident -> a -> Env v a
+singleton x n v = Env (IM.singleton (getKey x) (n, v))
 
-insert :: Ord v => v -> Ident -> a -> Env v a -> Env v a
-insert x n v (Env env) = Env (Map.insert x (n, v) env)
+insert :: HasKey v => v -> Ident -> a -> Env v a -> Env v a
+insert x n v (Env env) = Env (IM.insert (getKey x) (n, v) env)
 
-lookup :: Ord v => v -> Env v a -> Maybe (Ident, a)
-lookup x (Env env) = Map.lookup x env
+lookup :: HasKey v => v -> Env v a -> Maybe (Ident, a)
+lookup x (Env env) = IM.lookup (getKey x) env
 
-lookupVal :: Ord v => v -> Env v a -> Maybe a
+lookupVal :: HasKey v => v -> Env v a -> Maybe a
 lookupVal x env = snd <$> lookup x env
 
-lookupIdent :: Ord v => v -> Env v a -> Maybe Ident
+lookupIdent :: HasKey v => v -> Env v a -> Maybe Ident
 lookupIdent x env = fst <$> lookup x env
 
-named :: Text -> Env v a -> [(v, a)]
-named n (Env xs) = [(x, a) | (x, (Stx _ _ n', a)) <- Map.toList xs, n == n']
+named :: HasKey v => Text -> Env v a -> [(v, a)]
+named n (Env xs) = [(fromKey x, a) | (x, (Stx _ _ n', a)) <- IM.toList xs, n == n']
 
 type instance Index (Env v a) = v
 type instance IxValue (Env v a) = (Ident, a)
 
-instance Ord v => Ixed (Env v a) where
-  ix var f (Env env) = Env <$> ix var f env
+instance HasKey v => Ixed (Env v a) where
+  ix var f (Env env) = Env <$> ix (getKey var) f env
+  {-# INLINE ix #-}
 
-instance Ord v => At (Env v a) where
-  at x f (Env env) = Env <$> at x f env
+instance HasKey v => At (Env v a) where
+  at x f (Env env) = Env <$> at (getKey x) f env
+  {-# INLINE at #-}
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -40,6 +40,7 @@ module Expander (
 
 import Control.Applicative
 import Control.Lens hiding (List, children)
+import Control.Monad
 import Control.Monad.Except
 import Control.Monad.Reader
 import Control.Monad.State.Strict
@@ -48,7 +49,6 @@ import Data.Function (on)
 import Data.List (nub)
 import qualified Data.HashMap.Strict as HM
 import Data.Maybe
-import Data.Ord
 import Data.Sequence (Seq(..))
 import qualified Data.Sequence as Seq
 import Data.Text (Text)
--- a/src/Expander.hs
+++ b/src/Expander.hs
@@ -42,13 +42,15 @@ import Control.Applicative
 import Control.Lens hiding (List, children)
 import Control.Monad.Except
 import Control.Monad.Reader
-import Control.Monad.Writer
+import Control.Monad.State.Strict
 import Data.Foldable
+import Data.Function (on)
 import Data.List (nub)
-import Data.List.Extra (maximumOn)
-import qualified Data.Map as Map
+import qualified Data.HashMap.Strict as HM
 import Data.Maybe
-import qualified Data.Set as Set
+import Data.Ord
+import Data.Sequence (Seq(..))
+import qualified Data.Sequence as Seq
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
@@ -82,6 +84,9 @@ import Type
 import Value
 import World
 import qualified ScopeSet
+
+import qualified Util.Set   as Set
+import qualified Util.Store as S
 
 expandExpr :: Syntax -> Expand SplitCore
 expandExpr stx = do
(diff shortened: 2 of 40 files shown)


src/Env.hs
```diff
@@ -1,53 +1,61 @@
...
 import Data.Text (Text)
 
 import Syntax (Ident, Stx(..))
 
-newtype Env v a = Env (Map v (Ident, a))
-  deriving (Eq, Functor, Monoid, Semigroup, Show)
+import Util.Key
+
+-- | an environment is a mapping to a map of @Ident@s and some value. We keep
+-- 'v' as a phantom to stay honest
```
gelisam, line 20: Neither the comment nor the definition explains what `v` is supposed to represent. How about something like this:

```suggestion
-- | An environment maps variables of type 'v' to values of type 'a'.
--
-- More specifically, 'Env' is intended to be used for variables in the core
-- language, such as 'Var' and 'MacroVar', which are represented as unique
-- integers in order to avoid accidental capture. To make error message more
-- readable, 'Env' also stores the original 'Ident' in addition to the value.
```

gelisam, line 20: Neither the comment nor the definition explains what `v` is supposed to represent. How about something like this:

```suggestion
-- | An environment maps variables of type 'v' to values of type 'a'.
--
-- More specifically, 'Env' is intended to be used for variables in the core
-- language, such as 'Var' and 'MacroVar', which are represented as unique
-- integers in order to avoid accidental capture. To make error message more
-- readable, 'Env' also stores the original 'Ident' in addition to the value.
```

src/Expander.hs
```diff
@@ -42,13 +43,13 @@ import Control.Applicative
 import Control.Lens hiding (List, children)
 import Control.Monad.Except
 import Control.Monad.Reader
-import Control.Monad.Writer
+import Control.Monad.State.Strict
```
gelisam, line 46: Why not Control.Monad.Writer.CPS? It keeps the Writer API, making it clear that computations cannot read and be affected by the state, while using a State-based implementation under the hood.

src/Expander.hs
```diff
@@ -340,39 +343,41 @@ visibleBindings = do
...
+checkUnambiguous :: ScopeSet -> (Seq ScopeSet) -> Ident -> Expand ()
 checkUnambiguous best candidates blame =
   do p <- currentPhase
      let bestSize = ScopeSet.size p best
-     let candidateSizes = map (ScopeSet.size p) (nub candidates)
-     if length (filter (== bestSize) candidateSizes) > 1
-       then throwError (Ambiguous p blame candidates)
-       else return ()
+     let candidateSizes = Seq.fromList $ fmap (ScopeSet.size p) (nub $ toList candidates)
+     when (Seq.length (Seq.filter (== bestSize) candidateSizes) > 1)
```
gelisam, line 362: If the list is small enough that you're not concerned about the performance of `Data.List.nub`, it's probably not worth converting it back to `Seq` :)

This list should usually have length 1 or 2, as it is the list of shadowed variables with the same name and users don't typically shadow a variable which has already been shadowed. That being said, some macros can and do generate nested binders which all have the same dummy variable name, relying on the ScopeSet to disambiguate, so in that case the list could get longer, maybe a dozen elements.

src/Expander.hs
```diff
@@ -340,39 +343,41 @@ visibleBindings = do
...
     candidates ->
-      let best = maximumOn (ScopeSet.size p . fst) candidates
-      in checkUnambiguous (fst best) (map fst candidates) stx *>
-         return (snd best)
+      let
+        check = ScopeSet.size p . fst
+        c_last :: Seq a -> a
+        c_last (_xs Seq.:|> lst) = lst
+        c_last _                 = error "resolve: this is impossible"
+        (best_scpset, best_bind) = c_last $ Seq.sortOn check candidates
```
gelisam, line 378: `minimum` being asymptotically faster than `fst . sort` being the prototypical example of the benefit of laziness, I'm surprised to see that this time it's the opposite, `maximum` is slower than `last . sort`! Can you explain why we don't get the expected asymptotic gain from laziness in this case?

src/Scope.hs
```diff
@@ -1,14 +1,31 @@
...
+#ifndef KDEBUG
+newtype Scope = Scope { scopeNum :: Int}
+  deriving newtype (Eq, Ord, Show, HasKey)
+  deriving stock Data
+#else
+-- For a debug build Scope keeps a blob of text for debugging the expander
+-- output. This will have an impact of the performance of the interpreter so it
+-- won't be useful for performance issues
 data Scope = Scope { scopeNum :: Int, scopePurpose :: Text }
   deriving (Data, Eq, Ord, Show)
```
gelisam, line 29: Since we know that any two `Scope`s with the same `scopeNum` will have the same `scopePurpose`, would it make sense for the `Eq` and `Ord` instances to only look at `scopeNum`?

src/Pretty.hs
```diff
@@ -284,6 +293,17 @@ instance Pretty VarInfo core => Pretty VarInfo (ScopedString core) where
...
+                   $ foldr go (\e -> (mempty, e)) decls mempty
+    where
+      go :: CompleteDecl
+         -> (Env Var () -> ([Doc VarInfo], Env Var ()))
+         -> (Env Var () -> ([Doc VarInfo], Env Var ()))
+      go decl cc e = let (doc, e') = ppBind (env <> e) decl
+                         (docs, e'') = cc (e <> e')
+                     in (doc:docs, e'')
+
 instance PrettyBinder VarInfo [CompleteDecl] where
```
gelisam, line 307: Is the list instance still used?

src/Expander.hs
doyougnu: ok great then I can revert this one

src/Expander.hs
doyougnu: Ah I was not aware of `Writer.CPS`!

src/Scope.hs
doyougnu: Yes this would make perfect sense. Now that I know where the leaks are coming from I can experiment with adding a `scopePurpose` Store to the `Expander` environment. The only catch is that because that we would always have to be careful around that Store because if it is never forced then it'll still build up thunks even though Store is a strict IntMap

doyougnu: I don't love the CPP, but I also don't necessarily hate it, especially because from our conversations the `scopePurpose` seemed to be used to debug correctness not performance so a debug build would make sense in that case.

src/Pretty.hs
doyougnu: no i accidentally left this in, removed in https://github.com/gelisam/klister/pull/215/commits/d04248827f8d6e931c8ddcb4c23d6075cd93574c

src/Expander.hs
doyougnu: No I think it is just slower but non-dominating. But a Data.Sequence has log time access because its a finger tree, so its not as bad as a linked list.

src/Expander.hs
doyougnu: see 2e45103d571a88c9ceaf247a8b715d751c0f3f86

src/Expander.hs
doyougnu: see 2e45103d571a88c9ceaf247a8b715d751c0f3f86

doyougnu:
40d4fccd20e4e43123cb415b64c4b251239022c3 and 2e45103d571a88c9ceaf247a8b715d751c0f3f86 have regressed the performance a bit:

```
All 124 tests passed (16.11s)
Test suite klister-tests: PASS
Test suite logged to:
```

vs 

```
  All 124 tests passed (9.89s)
  Test suite klister-tests: PASS
```

in 85205d58ebe922b0c7e203960813972362fac5c9

src/Expander.hs
doyougnu: implemented in 40d4fccd20e4e43123cb415b64c4b251239022c3

gelisam:
Okay, then my intuition for performance is very bad, ignore my suggestions!

src/Expander/Syntax.hs
```diff
@@ -8,6 +8,7 @@
 module Expander.Syntax where
 
 import Control.Monad.Except
+import Control.Monad.IO.Class
```
gelisam, line 11: unused import?

doyougnu:
> Okay, then my intuition for performance is very bad, ignore my suggestions!

No the suggestions were good my execution was bad! I checked my OS's CPU scaling governor and realized it was in powersave because I was on battery! Here are the actual difference (on AC power of course):

```
All 124 tests passed (9.91s)
Test suite klister-tests: PASS
Test suite logged to:
```

so basically no difference between the two implementations.

src/Expander/Syntax.hs
doyougnu: No this depends on the version of mtl. I had to add it for 40d4fccd20e4e43123cb415b64c4b251239022c3. On master we pull in mtl-2.2.2 so `Control.Monad.Except` re-exports `Control.Monad.Trans` which re-exports `Control.Monad.IO.Class`. But in order to use `WriterT.CPS` we need mtl-2.3 which does not do these re-exports.

So basically this is a separate issue. IMHO we should drop support for GHC-8.8 and below. Leaving 8.10.7 as the oldest supported GHC. This would also cleanup all the warnings around the `Pretty` library.

david-christiansen: Let's drop everything under 8.10 :-)

david-christiansen:
I think the CPP is fine, but it's important to instantiate the debug flag both ways in CI.

tests/Test.hs
```diff
@@ -115,31 +124,10 @@ miniTests =
           , lam $ \f -> lam $ \x -> f `app` x
           )
         , ( "Trivial user macro"
-          , "[let-syntax \n\
-            \  [m [lambda [_] \n\
-            \       [pure [quote [lambda [x] x]]]]] \n\
-            \  m]"
-          , lam $ \x -> x
+          , trivialUserMacro, lam $ \x -> x
```
gelisam, line 127: It pains me to see you jump through hoops in order to keep those old tests running, let's get rid of them! #204

src/Scope.hs
gelisam: > The only catch is that because that we would always have to be careful around that Store because if it is never forced then it'll still build up thunks even though Store is a strict IntMap

The `scopePurpose` is only forced if it is part of an error message. Are saying that while the strict IntMap makes sure the force the thunk which produces the Text, we still need to remember to force the Store itself whenever we update the `Expander` field which holds the Store, otherwise nothing is going to force the thunk which calculates this update? Wouldn't it be sufficient to make that `Expander` field strict, using a `!`?

src/Util/Store.hs
```diff
@@ -0,0 +1,96 @@
...
+import qualified Data.IntMap.Strict as IM
+import Control.Arrow (first)
+
+import Util.Key
+import Phase
+
+newtype Store p v = Store { unStore :: IntMap v}
+  deriving newtype (Eq, Ord, Show, Semigroup, Monoid, Functor, Foldable)
+  deriving stock   Data
+type role Store representational _
```
gelisam, line 42: Why `representational`? I would have either have expected `phantom`, because IntMap's invariants only rely on Int's instances and not `p`'s, or `nominal`, to allow coercing a `Store K1 (Sum Int)` into a  `Store K1 Int` but prevent accidentally coercing a `Store K1 Int` into a `Store K2 Int`.

gelisam: (and similarly for `Util.Set`)

doyougnu: I think nominal is better.

tests/Test.hs
doyougnu: I think this is better left to another PR

src/Expander/Monad.hs
```diff
@@ -365,11 +367,19 @@ modifyState f = do
   !st <- view expanderState <$> expanderContext
   liftIO (modifyIORef' st f)
 
+#ifndef KDEBUG
+freshScope :: Expand Scope
+freshScope = do
+  n <- view expanderNextScopeNum <$> getState
+  modifyState $ over expanderNextScopeNum (+ 1)
+  return (Scope n)
```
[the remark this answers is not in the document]
doyougnu: I like the idea, but this would require decoupling `Expand` from the `Monad` module because the smart constructor needs to know about the type `Text -> Expand Scope`, so if I make a module to encapsulate the smart constructor then it must import `Expander.Monad` in order to gain access to the `Expand` type, but `Expander.Monad` is exactly the module I want to import the smart constructor to. Thus we get a module cycle.

stack.yaml
```diff
@@ -1,4 +1,6 @@
-resolver: lts-14.8
+resolver: lts-20.25
 packages:
 - .
 extra-deps: []
+
+allow-newer: true
```
gelisam, line 6: :x: 
stack's implementation of `allow-newer` is very broken: if you build with `allow-newer: true`, and then you later _remove_ `allow-newer: true`, your cache is now poisoned, stack will continue to use packages previously-built with newer versions, and we lose the reproducibility advantage which stack claims to have over cabal.

Which packages need allow-newer though? you didn't list any extra-deps, and every package in the LTS is supposed to be compatible with every other package, no?

gelisam (changes requested):
Looks great! But could we please remove the `allow-newer: true` before it poisons anybody's cache please?

stack.yaml
doyougnu: Ah yea I meant to highlight this. Without setting that flag we get a lot of extra-deps, I'll try to clean it up tonight

doyougnu:
Alright, that should be it. The root problem behind the packaging was hedgehog.

gelisam (approved):
Approved! Sorry it took so long.

PR #223, opened 2024-01-14
A CEK Evaluator

doyougnu:
Still a draft but all but one test is passing.

TODOs:
- [x] A more robust note
- [x] add proper error handling
- [x] rebase onto main
- [x] remove the old evaluator

## Summary by CodeRabbit

- **New Features**
	- Enhanced evaluation process in the REPL for improved error handling and result display.
	- Added new language extensions for advanced type handling and pattern matching.
	- Introduced new functions for better error reporting and type evaluation.

- **Bug Fixes**
	- Improved error handling and evaluation flow in the CEK interpreter.

- **Documentation**
	- Updated method names and signatures for clarity and consistency.

- **Refactor**
	- Renamed functions to better reflect their usage and purpose across modules.

- **Style**
	- Enhanced formatting for error messages and evaluation results for better user experience.

### DIFF
--- a/src/Evaluator.hs
+++ b/src/Evaluator.hs
@@ -1,20 +1,77 @@
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE GADTSyntax #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ParallelListComp #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
-module Evaluator where
+
+{- Note [The CEK interpreter]:
+
+The Klister interpreter is a straightforward implementation of a CEK
+interpreter. The interpreter keeps three kinds of state:
+
+-- C: Control      ::= The thing that is being evaluated
+-- E: Environment  ::= The interpreter environment
+-- K: Kontinuation ::= The syntactic context of the thing that is being interpreted
+
+Why a CEK? A CEK interpreter allows us to have precise control over the
+evaluation of a klister program. For example, because the interpreter keeps a
+reference to the kontinuation we can provide stack traces. This handle also
+makes a more advanced debugger possible. Upon an evaluation error we could save
+the kontinuation stack, over write a variable in the environment a la common
+lisp or even rewind the evaluation
+
+See Matthias Felleison's course website for a good reference:
+https://felleisen.org/matthias/4400-s20/lecture23.html
+
+The bird's eye view:
+
+The evaluator crawl's the input AST and progresses in three modes:
+
+-- 'Down': meaning that the evaluator is searching for a redex to evaluate and
+-- therefore moving "down" the AST.
+
+-- 'Up': meaning that the evaluator has evaluated some redex to a value and is
+-- passing that value "up" the execution stack.
+
+-- 'Er': meaning that something has gone wrong, the stack is captured and the Er
+-- will float up to be handled by the caller of the evaluator.
+
+All interesting things happen by matching on 'Kont', the continuation. This
+allows the evaluator to know exactly what needs to happen in order to continue.
+
+-- TODO: #108 describe the how the debugger hooks in
+
+-}
+
+module Evaluator
+  ( EvalError (..)
+  , EvalResult (..)
+  , TypeError (..)
+  , evaluate
+  , evaluateIn
+  , evaluateWithExtendedEnv
+  , evalErrorType
+  , evalErrorText
+  , projectError
+  , erroneousValue
+  , applyInEnv
+  , apply
+  , doTypeCase
+  , try
+  ) where
 
 import Control.Lens hiding (List, elements)
-import Control.Monad.IO.Class (MonadIO)
-import Control.Monad.Except (MonadError(throwError))
-import Control.Monad.Reader (MonadReader(ask, local))
-import Control.Monad.Trans.Except (ExceptT)
-import Control.Monad.Trans.Reader (ReaderT)
+import Control.Exception hiding (TypeError, evaluate)
+import Data.Data (Typeable)
 import Data.Text (Text)
 import qualified Data.Text as T
+import Data.List (foldl')
 
+import Datatype
 import Core
 import Env
 import ShortShow
@@ -22,6 +79,14 @@ import Syntax
 import Syntax.SrcLoc
 import Type
 import Value
+
+-- -----------------------------------------------------------------------------
+-- Interpreter Data Types
+
+
+data EvalResult
+  = ExampleResult SrcLoc VEnv Core (Scheme Ty) Value
+  | IOResult (IO ())
 
 -- TODO: more precise representation
 type Type = Text
@@ -38,8 +103,345 @@ data EvalError
   | EvalErrorType TypeError
   | EvalErrorCase SrcLoc Value
   | EvalErrorUser Syntax
-  deriving (Show)
+  | EvalErrorIdent Value
+  deriving (Show, Typeable)
 makePrisms ''EvalError
+instance Exception EvalError
+
+-- | The Kontinuation type. The naming convention InFoo means that the subject
+-- of evaluation in the CEK machine is Foo. For example, when the continuation
+-- is 'InArg' the subject of evaluation is the argument of the function and the
+-- continuation holds the un-evaluated function symbol in its first field.
+data Kont where
+
+  Halt :: Kont
+  -- ^ Marks the evaluator finishing
+
+  -- functions
+  InArg :: !Value -> !VEnv -> !Kont -> Kont
+  -- ^ The argument is being evaluated, so hold onto the function symbol. We
+  -- require that the function symbol be fully evaluated before evaluating the
+  -- arguments, hence the @Value@
+  InFun :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  -- ^ The function is being evaluated, so hold onto the argument.
+
+  InLetDef :: !Ident -> !Var -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  -- ^ Evaluating the let def
+
+  -- constructors
+  InCtor :: ![Value] -> !Constructor -> ![CoreF TypePattern ConstructorPattern Core] -> !VEnv -> !Kont -> Kont
+
+  -- Cases
+  InCaseScrut     :: ![(SyntaxPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+  InDataCaseScrut :: ![(ConstructorPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+  InTypeCaseScrut :: ![(TypePattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+
+  -- lists
+  InConsHd :: !Core -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  InConsTl :: !Core -> !Syntax -> !VEnv -> !Kont -> Kont
+  InList   :: !Core -> ![Core] -> ![Syntax] -> !VEnv -> !Kont -> Kont
+
+  -- idents
+  InIdent :: !Core -> !VEnv -> !Kont -> Kont
+  InIdentEqL :: !HowEq -> !Core -> !VEnv -> !Kont -> Kont
+  InIdentEqR :: !Value -> !HowEq -> !VEnv -> !Kont -> Kont
+
+  -- Macros
+  InPureMacro :: !VEnv -> !Kont -> Kont
+  InBindMacroHd :: !Core -> !VEnv -> !Kont -> Kont
+  InBindMacroTl :: !MacroAction -> !VEnv -> !Kont -> Kont
+
+  -- atomics
+  InInteger :: !Core -> !VEnv -> !Kont -> Kont
+  InString  :: !Core -> !VEnv -> !Kont -> Kont
+  InReplaceLocL :: !Core -> !VEnv -> !Kont -> Kont
+  InReplaceLocR :: !SrcLoc -> !VEnv -> !Kont -> Kont
+
+  -- scope
+  InScope :: !(ExprF Syntax) -> !VEnv -> !Kont -> Kont
+
+  -- logs and errors
+  InLog   :: !VEnv -> !Kont -> Kont
+  InError :: !VEnv -> !Kont -> Kont
+
+
+  InSyntaxErrorMessage   :: ![Core] -> !VEnv -> !Kont -> Kont
+  InSyntaxErrorLocations :: !Syntax -> ![Core] -> ![Syntax] -> !VEnv -> !Kont -> Kont
+
+-- | The state of the evaluator
+data EState where
+  Down :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> EState
+  -- ^ 'Down', we are searching the AST for a redex and building up the stack of
+  -- continuations
+  Up   :: !Value -> !VEnv -> !Kont -> EState
+  -- ^ 'Up', means we have performed some evaluation on a redex and are
+  -- returning a value up the stack
+  Er   :: !EvalError -> !VEnv -> !Kont -> EState
+  -- ^ 'Er', meaning that we are in an error state and running the debugger
+
+
+-- -----------------------------------------------------------------------------
+-- The evaluator. The CEK machine is a state machine, the @step@ function moves
+-- the state machine a single step of evaluation. This is the heart of the
+-- evaluator.
+
+
+-- | Make a single step transition in the CEK state machine.
+step :: EState -> EState
+step done@(Up _val _ Halt) = done
+
+-- for now we just bail out. Once we have a debugger we'll do something more
+-- advanced.
+step done@(Er _err _env _k)  = done
+
+-- Upsweep, returning a value after evaluating a redex
+step (Up v e k) =
+  case k of
+    -- functions
+    -- we evaluated the arg to get a closed so now we evaluate the fun
+    (InArg fun env kont) -> applyAsClosure env fun v kont
+    -- we evaluated the fun so now do the application
+    (InFun arg env kont) -> Down arg env (InArg v e kont)
+
+
+    -- lets
+    -- we have the value for the def, now eval the body
+    (InLetDef id' var body env kont) -> Down body (extend id' var v env) kont
+
+    -- done, FIXME use a banker's queue instead of a list
+    (InCtor v_args c [] _env kont) -> Up (ValueCtor c (reverse $ v : v_args)) e kont
+    -- still processing
+    (InCtor vs c (a:as) env kont) -> Down a env (InCtor (v:vs) c as env kont)
+
+
+    -- Cases
+    (InCaseScrut cs loc env kont)     -> doCase loc v cs env kont
+    (InDataCaseScrut cs loc env kont) -> doDataCase loc v cs env kont
+    (InTypeCaseScrut cs loc env kont) ->
+      evalAsType v
+      (\good -> Up (ValueMacroAction $ MacroActionTypeCase e loc good cs) env kont)
+      (\err  -> Er err env kont)
+
+
+    -- Idents
+    (InIdent scope env kont) -> case v of
+      ValueSyntax stx ->
+        case _unSyntax stx of
+          (Stx _ _ expr) -> case expr of
+            Integer _ ->
+              Er (EvalErrorType
+                   $ TypeError { _typeErrorExpected = "id"
+                               , _typeErrorActual   = "integer"
+                               }) e k
+            String _ ->
+              Er (EvalErrorType
+                  $ TypeError { _typeErrorExpected = "id"
+                              , _typeErrorActual   = "string"
+                              }) e k
+            List _ ->
+              Er (EvalErrorType
+                  $ TypeError { _typeErrorExpected = "id"
+                              , _typeErrorActual   = "list"
+                              }) e k
+            name@(Id _) -> Down (unCore scope) env (InScope name env kont)
+      other -> Er (EvalErrorIdent other) e k
+    (InIdentEqL how r env kont)  -> Down (unCore r) env (InIdentEqR v how env kont)
+    (InIdentEqR how lv env kont) -> Up (ValueMacroAction $ MacroActionIdentEq lv how v) env kont
+
+    -- Short circuit to speed this up, we could issue an Down and do this recursively
+    (InScope expr env kont) ->
+      evalAsSyntax v
+      (\(Syntax (Stx scopeSet loc _)) -> Up (ValueSyntax $ Syntax $ Stx scopeSet loc expr) env kont)
+      (\err                           -> Er err env kont)
+
+
+    -- pairs
+    (InConsHd scope tl env kont) ->
+      evalAsSyntax v
+      (\good -> Down tl env (InConsTl scope good env kont))
+      (\err  -> Er err env kont)
+    (InConsTl scope hd env kont) ->
+      evalAsSyntax v
+      (\(Syntax (Stx _ _ expr)) ->
+          case expr of
+            List tl -> Down (unCore scope) env (InScope (List $ hd : tl) env kont)
+            String _ ->
+              Er (EvalErrorType
+                   $ TypeError { _typeErrorExpected = "list"
+                               , _typeErrorActual   = "string"
+                               }) e k
+            Id _ -> Er (EvalErrorType
+                        $ TypeError { _typeErrorExpected = "list"
+                                    , _typeErrorActual   = "id"
+                                    }) e k
+            Integer _ -> Er (EvalErrorType
+                             $ TypeError { _typeErrorExpected = "list"
+                                         , _typeErrorActual   = "integer"
+                                         }) e k
+         )
+      (\err -> Er err env kont)
+
+
+    -- lists
+    -- base case
+    (InList scope [] dones env kont) ->
+      evalAsSyntax v
+      (\good -> Down (unCore scope) e (InScope (List $ reverse $ good : dones) env kont))
+      (\err  -> Er err env kont)
+    -- still some todo
+    (InList scope (el:els) dones env kont) ->
+      evalAsSyntax v
+      (\good -> Down (unCore el) env (InList scope els (good : dones) env kont))
+      (\err  -> Er err env kont)
+
+
+    -- Macros
+    (InPureMacro env kont) -> Up (ValueMacroAction $ MacroActionPure v) env kont
+    (InBindMacroHd tl env kont) ->
+      evalAsMacroAction v
+      (\good -> Down (unCore tl) env (InBindMacroTl good env kont))
+      (\err  -> Er err env kont)
+
+    (InBindMacroTl macroAction env kont) ->
+      evalAsClosure v
+      (\good -> Up (ValueMacroAction $ MacroActionBind macroAction good) env kont)
+      (\err  -> Er err env kont)
+
+
+    -- Syntax and Atomics
+    (InInteger scope env kont) ->
+      evalAsInteger v
+      (\good -> Down (unCore scope) env (InScope (Integer good) env kont))
+      (\err  -> Er err env kont)
+    (InString scope env kont) ->
+      evalAsString v
+      (\good -> Down (unCore scope) env (InScope (String good) env kont))
+      (\err  -> Er err env kont)
+    (InReplaceLocL stx env kont) ->
+      evalAsSyntax v
+      (\(Syntax (Stx _ newLoc _)) -> Down (unCore stx) env (InReplaceLocR newLoc env kont) )
+      (\err -> Er err env kont)
+    (InReplaceLocR loc env kont) ->
+      evalAsSyntax v
+      (\(Syntax (Stx scs _ contents)) -> Up (ValueSyntax $ Syntax $ Stx scs loc contents) env kont)
+      (\err -> Er err env kont)
+    (InLog   env kont)   ->
+      evalAsSyntax v
+      (\good -> Up (ValueMacroAction (MacroActionLog good)) env kont)
+      (\err  -> Er err env kont)
+
+
+    -- Errors
+    (InError env kont) ->
+      evalAsSyntax v
+      (\good -> Er (EvalErrorUser good) env kont)
+      (\err  -> Er err env kont)
+    (InSyntaxErrorMessage locs env kont) ->
+      evalAsSyntax v
+      (\msg_syn ->
+         case locs of
+           -- done
+           []     -> Up (ValueMacroAction $ MacroActionSyntaxError
+                          (SyntaxError { _syntaxErrorMessage   = msg_syn
+                                       , _syntaxErrorLocations = mempty
+                                       })) env kont
+           (l:ls) -> Down (unCore l) env (InSyntaxErrorLocations msg_syn ls mempty env kont)
+          )
+      (\err -> Er err env kont)
+    -- done
+    (InSyntaxErrorLocations msg_syn [] dones env kont) ->
+        Up (ValueMacroAction
+                $ MacroActionSyntaxError (SyntaxError { _syntaxErrorMessage   = msg_syn
+                                                      , _syntaxErrorLocations = dones
+                                                      })) env kont
+    (InSyntaxErrorLocations msg (l:ls) dones env kont) ->
+      evalAsSyntax v
+      (\good -> Down (unCore l) env (InSyntaxErrorLocations msg ls (good : dones) env kont))
+      (\err  -> Er err env kont)
+
+-- the downsweep, searching for a redex to evaluate.
+step (Down c env k)  =
+  case c of
+
+    -- atoms
+    (CoreString s)    -> Up (ValueString s) env k
+    (CoreInteger i)   -> Up (ValueInteger i) env k
+    (CoreIntegerSyntax (ScopedInteger int scope)) -> Down (unCore int) env (InInteger scope env k)
+    (CoreStringSyntax  (ScopedString  str scope)) -> Down (unCore str) env (InString scope env k)
+    (CoreSyntax s)    -> Up (ValueSyntax s) env k
+    (CoreError what)  -> Down (unCore what) env (InError env k)
+    (CoreEmpty (ScopedEmpty scope)) -> Down (unCore scope) env (InScope (List mempty) env k)
+    CoreMakeIntroducer -> Up (ValueMacroAction MacroActionIntroducer)   env k
+    CoreWhichProblem   -> Up (ValueMacroAction MacroActionWhichProblem) env k
+
+
+    -- variables and binders
+    (CoreVar var) ->
+      case lookupVal var env of
+        Just val -> Up val env k
+        _        -> Er (EvalErrorUnbound var) env k
+
+    (CoreLet ident var def body) ->
+      Down (unCore def) env (InLetDef ident var (unCore body) env k)
+
+    (CoreLetFun fIdent fVar argIdent argVar def body) ->
+      let vFun = ValueClosure $ FO $ FOClosure
+            { _closureEnv   = Env.insert fVar fIdent vFun env
+            , _closureIdent = argIdent
+            , _closureVar   = argVar
+            , _closureBody  = def
+            }
+          newEnv = Env.insert fVar fIdent vFun env
+      in Down (unCore body) newEnv k
+
+    (CoreCtor con args) -> case args of
+                           -- just a symbol, shortcut out
+                           []     -> Up (ValueCtor con mempty) env k
+                           -- process fields left to right
+                           (f:fs) -> Down (unCore f) env (InCtor mempty con (fmap unCore fs) env k)
+
+
+    -- lambdas and application
+    (CoreLam ident var body) ->
+      let lam = ValueClosure $ FO $ FOClosure
+            { _closureEnv   = env
+            , _closureIdent = ident
+            , _closureVar   = var
+            , _closureBody  = body
+            }
+      in Up lam env k
+    (CoreApp fun arg) -> Down (unCore fun) env (InFun (unCore arg) env k)
+
+
+    -- cases
+    (CoreCase     loc scrutinee cases) -> Down (unCore scrutinee) env (InCaseScrut     cases loc env k)
+    (CoreDataCase loc scrutinee cases) -> Down (unCore scrutinee) env (InDataCaseScrut cases loc env k)
+    (CoreTypeCase loc scrut cases)     -> Down (unCore scrut)     env (InTypeCaseScrut cases loc env k)
+
+    (CoreIdent (ScopedIdent ident scope)) -> Down (unCore ident) env (InIdent scope env k)
+    (CoreIdentEq how l r)                 -> Down (unCore l)     env (InIdentEqL how r env k)
+
+    (CoreCons (ScopedCons hd tl scope))   -> Down (unCore hd) env (InConsHd scope (unCore tl) env k)
+    -- empty, short circuit
+    (CoreList (ScopedList ls scope)) -> case ls of
+                                         []     -> Down (unCore scope) env (InScope (List []) env k)
+                                         (e:es) -> Down (unCore e) env (InList scope es mempty env k)
+    (CoreReplaceLoc loc stx) -> Down (unCore loc) env (InReplaceLocL stx env k)
+
+
+    -- macros
+    (CorePureMacro arg)   -> Down (unCore arg) env (InPureMacro env k)
+    (CoreBindMacro hd tl) -> Down (unCore hd) env (InBindMacroHd tl env k)
+
+
+    -- others
+    (CoreLog msg)      -> Down (unCore msg) env (InLog env k)
+    (CoreSyntaxError err) ->
+      Down (unCore $ _syntaxErrorMessage err) env (InSyntaxErrorMessage (_syntaxErrorLocations err) env k)
+
+
+-- -----------------------------------------------------------------------------
+-- Helper Functions
 
 evalErrorText :: EvalError -> Text
 evalErrorText (EvalErrorUnbound x) = "Unbound: " <> T.pack (show x)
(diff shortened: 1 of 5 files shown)


gelisam:
I'm experimenting with an AI-based code review tool called "CodeRabbit", please let me know if its comments are annoying and I'll turn it off.

doyougnu:
Should be ready for review now. Its a bit slower than what we had. Why is that? Well:

```
$  perf stat -- cabal test --test-show-details=streaming
 Performance counter stats for 'cabal test --test-show-details=streaming':

         18,349.25 msec task-clock:u                     #    0.996 CPUs utilized          
                 0      context-switches:u               #    0.000 /sec                   
                 0      cpu-migrations:u                 #    0.000 /sec                   
           717,329      page-faults:u                    #   39.093 K/sec                  
    77,585,648,085      cycles:u                         #    4.228 GHz                    
   169,000,946,915      instructions:u                   #    2.18  insn per cycle         
    36,205,500,147      branches:u                       #    1.973 G/sec                  
       497,651,939      branch-misses:u                  #    1.37% of all branches        
   387,384,692,920      slots:u                          #   21.112 G/sec                  
   111,272,326,760      topdown-retiring:u               #     19.8% Retiring              
   335,353,029,163      topdown-bad-spec:u               #     59.7% Bad Speculation       
    75,804,429,827      topdown-fe-bound:u               #     13.5% Frontend Bound        
    39,530,241,508      topdown-be-bound:u               #      7.0% Backend Bound         

      18.416893454 seconds time elapsed

      17.278805000 seconds user
       1.124952000 seconds sys
```

Because it destroys the branch predictor in my laptop! {59.7% bad speculation)

src/Evaluator.hs
```diff
@@ -50,313 +450,217 @@ evalErrorText (EvalErrorCase loc val) =
...
+    inserter = foldl' (\acc (n,x,v) -> Env.insert x n v acc) env
+
+-- TODO DYG: Move to separate module
+projectError :: EState -> EvalError
+projectError (Er err _env _k) = err
+projectError _                = error "debugger: impossible"
+
+erroneousValue :: EvalError -> Value
+erroneousValue (EvalErrorCase _loc v) = v
+erroneousValue  _                    = error "erroneusValue: likely"
```
gelisam, line 666: what does "likely" mean here? the opposite of "impossible"?

src/Evaluator.hs
```diff
@@ -50,313 +450,217 @@ evalErrorText (EvalErrorCase loc val) =
...
+
+evaluateWithExtendedEnv :: VEnv -> [(Ident, Var, Value)] -> Core -> Either EState Value
+evaluateWithExtendedEnv env exts = evaluateIn (inserter exts)
+  where
+    inserter = foldl' (\acc (n,x,v) -> Env.insert x n v acc) env
+
+-- TODO DYG: Move to separate module
+projectError :: EState -> EvalError
+projectError (Er err _env _k) = err
+projectError _                = error "debugger: impossible"
```
gelisam, line 662: I see that `repl/Main.hs` has
```haskell
case evaluateIn ... of
  Left evalErr -> print $ erroneousValue $ projectError evalErr
  Right val    -> ...
```
so is this "impossible" in the sense that when `evaluateIn` returns a `Left eState`, that `EState` is always an `Er`? In that case, wouldn't it make more sense to return `Left err` or `Left (err, _env, _k)`, to make impossible states unrepresentable?

src/Expander/Monad.hs
```diff
@@ -860,14 +861,14 @@ setTasks = modifyState . set expanderTasks
 clearTasks :: Expand ()
 clearTasks = modifyState $ set expanderTasks []
 
-expandEval :: Eval a -> Expand a
+expandEval :: Core -> Expand Value
 expandEval evalAction = do
```
gelisam, line 865: should probably not be called `evalAction` anymore since its type is no longer `Eval a`

gelisam: `expandEval` should probably be renamed too since it no longer takes an `Eval a` as input. but what would be a better name? `expandCore` is the most obvious candidate, but this is not expanding the core, this is evaluating it. `evalCore`? but then what's the difference between `evalCore` and `evaluate`? the difference is that `evalCore` runs in the `Expand` monad. so we are expanding in the current phase. so maybe `evalInCurrentPhase`? that makes sense for something which runs in the `Expand` monad, since only the `Expand` monad has a concept of a current phase.

src/Expander.hs
```diff
@@ -558,18 +553,14 @@ initializeKernel outputChannel = do
...
                         var = view closureVar clos
                         ident = view closureIdent clos
                         body = view closureBody clos
-                    out <- runExceptT $ flip runReaderT env $ runEval $
-                      withExtendedEnv ident var vx $
-                        eval body
-                    case out of
-                      Left err -> error (T.unpack (pretty err))
+                    case (evaluateWithExtendedEnv env [(ident, var, vx)] body) of
+                      Left err -> error (T.unpack (pretty $ projectError err))
```
gelisam, line 563: woah! why do we fail with `error` here? I think `runEval` used to only fail if there was a bug in our implementation of the compiler, e.g. if our type checker allowed the construction of an ill-typed Core term. but nowadays evaluating a Core expression can fail with `EvalErrorCase` or `EvalErrorUser` if there is a bug in the user's `.kl` code too, so we should probably use `throwError` here. this is unrelated to the current PR, so I'll create a separate ticket to track this: #237

src/Expander.hs
```diff
@@ -983,24 +981,36 @@ runTask (tid, localData, task) = withLocal localData $ do
...
-          forkInterpretMacroAction dest macroAction kont
-        other -> expandEval $ evalErrorType "macro action" other
+      case apply closure value of
+        Left err -> do
+          p <- currentPhase
+          throwError
+            $ MacroEvaluationError p
+            $ evalErrorType "macro action"
+            $ erroneousValue
+            $ projectError err
```
gelisam, line 995: this error-handling logic is repeated several time, wasn't it nicer when `expandEval` was running in the `Expand` monad and was thus able to throw this error right away?

src/Expander.hs
```diff
@@ -914,13 +905,20 @@ runTask (tid, localData, task) = withLocal localData $ do
...
               forkInterpretMacroAction dest nextStep kont
             otherVal -> do
-              expandEval $ evalErrorType "macro action" otherVal
+              p <- currentPhase
+              throwError $ MacroEvaluationError p $ evalErrorType "macro action" otherVal
+          Left err -> do
+            p <- currentPhase
+            throwError
+              $ MacroEvaluationError p
+              $ evalErrorType "macro action"
```
gelisam, line 919: :x: this isn't a type error, this is converting the evaluation error into an `Expand` error.

src/Expander.hs
```diff
@@ -1301,22 +1311,28 @@ expandOneForm prob stx
           implV <- Env.lookupVal transformerName <$> currentTransformerEnv
           case implV of
             Just (ValueClosure macroImpl) -> do
-              macroVal <- inEarlierPhase $ expandEval $
-                          apply macroImpl $
-                          ValueSyntax $ addScope p stepScope stx
+              macroVal <- inEarlierPhase
+                          $ pure
```
gelisam, line 1315: as discussed, `inEarlierPhase` uses `local` to change the current phase within the body, but `pure` can't access the current phase. is this a bug? or did the previous code simply didn't use the current phase, so this call to `inEarlierPhase` was already dead code?

gelisam: The `inEarlierPhase` was _not_ dead code. On the `main` branch, if I run this program:
```
#lang "prelude.kl"

(define-macros
  ([my-error
    (lambda (stx)
      (error '"oops"))]))
(example (my-error))
```
Then I get the expected error message, namely that `error` was raised at phase 1, while evaluating the body of the macro:
```
Error at phase p1: <loc>: "oops"
```
But if I delete `inEarlierPhase` from this `EUserMacro` case in `main`, the error message instead claims that `error` was raised in phase 0, which is not correct:
```
Error at phase p0: <loc>: "oops"
``` 

When I run the above program in this `doyougnu/wip/useful-repl` branch, I instead get the error message
```
Not a macro monad value: erroneusValue: likely
```
Which I think says that that a primitive claims it has type `Macro something` but is returning something else, a string for example. Technically `(error '"oops")` shouldn't be returning anything at all though, it should have the side-effect of terminating the program early!

gelisam: thanks for introducing this bug, this allowed me to find a related bug in the code which immediately follows! the `interpretMacroAction` call should also be `inEarlierPhase`, but isn't.

it is surprising how little of an impact these two bugs have!

gelisam: my mistake, whether `inEarlierPhase` is or isn't added to the `interpretMacroAction` call does make a big difference.

adding `inEarlierPhase` makes so many tests fail that it's hard to tell which of the many error messages is the root cause. One relatively simple piece which breaks is `free-identifier=?`:
```
#lang "prelude.kl"

(define-macros
  ([my-keyword1
    (lambda (stx)
      (syntax-error '"my-keyword1 is a keyword"))]))
(define my-keyword2 "my-keyword2 is a keyword")
(define-macros
  ([my-macro
    (lambda (stx)
      (case (open-syntax stx)
        [(list-contents (:: _ (:: x (nil))))
         (>>= (free-identifier=? x 'my-keyword1)
           (lambda (r)
             (if r
               (pure '"input is my-keyword1")
               (>>= (free-identifier=? x 'my-keyword2)
                 (lambda (r)
                   (if r
                     (pure '"input is my-keyword2")
                     (pure '"input is not a keyword")))))))]))]))
(example (my-macro my-keyword1))
(example (my-macro my-keyword2))
```
on `main` the above program prints "input is my-keyword1" and "input is my-keyword2", whereas with `inEarlierPhase $ interpretMacroAction prob act` the program instead prints "input is not a keyword" twice.

gelisam: here's my reasoning for why I think `inEarlierPhase` is needed in both places. Let's look at another simple Klister program.
```
#lang "prelude.kl"

(define x 42)
(meta
  (define >>
    (lambda (action1 action2) 
      (>>= action1
        (lambda (_)
          action2)))))
(define-macros
  ([my-macro
    (lambda (stx)
      (>> (log '"inside my-macro")
          (pure 'x)))]))
(example (my-macro))
```

In order to run this program (at phase 0), we must first expand it into Core. When expanding `(my-macro)`, the expander notices that this is a user-defined macro (the `EUserMacro transformerName -> ...` case is selected) and the definition of `my-macro` is looked up (`Env.lookupVal`). It is looked up in the phase 0 _transformer_ environment, which I like to think as being between the phase 0 _regular_ environment and the phase 1 regular environment.

We find the definition:
```
(lambda (stx)
  (>> (log '"inside my-macro")
    (pure 'x)))
```
This is a pure expression whose variables, such as `>>`, come from the phase 1 regular environment (and whose macros come from the phase 1 transformer environment). We must thus evaluate the application of this phase 1 expression applied to the syntax object `'(my-macro)` in phase 1, hence 
```
inEarlierPhase $ expandEval $
  apply macroImpl $
  ValueSyntax {- (my-macro) -}
```

The result of this evaluation is a `Macro Syntax` action. We're still inside the macro definition, so I would expect to execute this action inside phase 1, hence `inEarlierPhase $ interpretMacroAction prob act`.

But `my-keyword1` and `my-keyword2` live in the phase 0 transformer environment and in the phase 0 regular environment, respectively. Is that why the macro action needs to run in phase 0? That still seems wrong to me. Yes, `my-keyword1` and `my-keyword2` live in phase 0, but they are _quoted_ identifiers, so I would expect `my-keyword2` to be a phase 0 expression and `'keyword2` to be a phase 1 expression referring to a phase 0 expression!

Is it possible that we _do_ need `inEarlierPhase $ interpretMacroAction prob act`, as I expect, and that it is `free-identifier=?` which is wrong, because it looks in the current environment but should be looking in the environment for the next phase? Or is it me whose understanding is completely backwards?

gelisam: I created #240 to verify my understanding of the bug independently of this MR. but we should not merge #240, as it would just conflict with and get overwritten by this MR. Instead, the real fix must be performed on the CEK evaluator. I think I will attempt to write that fix, that will force me to understand, and thus review, the CEK evaluator!

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+
+The bird's eye view:
+
+The evaluator crawl's the input AST and progresses in three modes:
+
+-- 'Down': meaning that the evaluator is searching for a redex to evaluate and
+-- therefore moving "down" the AST.
+
+-- 'Up': meaning that the evaluator has evaluated some redex to a value and is
+-- passing that valua "up" the execution stack.
```
gelisam, line 38:
```suggestion
-- passing that value "up" the execution stack.
```

src/Evaluator.hs
doyougnu: Yes that would make more sense, but I specifically left it this way because this state will become important for the debugger. The debugger will be typed `debugger :: EState -> DebugM EState` or something like that. But the key point is that the debugger should be able to input `EState` which is actually at `Er ...` and return an `EState` that is not in `Er`. So I left as much information as possible until the very last projection or pretty printing because when `evaluateIn` returns a `Left (bunch of state)` the debugger could take that state and then resume the computation. Perhaps I should add a TODO in a comment

src/Expander/Monad.hs
doyougnu: ah great points! `evalInCurrentPhase` seems right to me. I agree that `expandCore` or `evalCore` is too general. I wonder if the naming difficulty is indicative of a larger architectural problem.

src/Expander.hs
doyougnu: I'm hesitant to DRY it out because I still want to workshop the design of the debugger and how it hooks into `expandEval`.

> throw this error right away

This action will be calling the debugger. Immediately exiting the debugger should then be throwing the error and reporting the error back to the user. Perhaps `Expand` will have to be a monad transformer over `DebugM`? I'm honestly not sure if anyone has made a CommonLisp style debugger in Haskell yet.

src/Evaluator.hs
doyougnu: returning to this, I cannot for the life of me understand what `likely` is doing there and `erroneous` is spelled wrong and this function is missing a case for `EvalErrorIdent`

src/Expander.hs
gelisam: turns out the CEK evaluator does not need to be changed to address this bug, since the _evaluation_ does not depend on the current phase), only the _execution_ and _expansion_ does.

src/Expander.hs
```diff
@@ -1301,22 +1310,28 @@ expandOneForm prob stx
...
+                          $ apply macroImpl
+                          $ ValueSyntax
+                          $ addScope p stepScope stx
               case macroVal of
-                ValueMacroAction act -> do
-                  res <- interpretMacroAction prob act
-                  case res of
+                Left err -> throwError
+                            $ ValueNotMacro
```
gelisam, line 1320: I don't think this should be a `ValueNotMacro` error. In this branch, looking up `transformerName` did successfully find a macro definition of type `(-> Syntax (Macro Syntax))`. That function was then applied to a `Syntax` argument, and the evaluation of that function failed for some reason, perhaps because `(error 'msg)` was called.

gelisam: Ha! The existing code already misuses `ValueNotMacro`. When expanding the macro call site `(my-macro)`, many things can go wrong:
1. `my-macro` might not be in scope, in which case `resolve` (line L1201) throws `Unknown` and the user sees `Unknown: my-macro`.
2. `my-macro` resolves to `EUserMacro transformerName`, but that `transformerName` is somehow not the `MacroVar` which was added to the transformer environment by `(define-macros ([my-macro ...]))` at the same time as `my-macro` was bound to `EUserMacro` in the binding table. For example, maybe `transformerName` is not in the binding table at all. This is a bug in Klister, so `InternalError` is thrown; `error` would have been fine too.
3. Same thing, but `transformerName` is bound to something other than a `(-> Syntax (Macro Syntax))`. This is also a bug in Klister, but `ValueNotMacro` is thrown. This is definitely wrong, since the error message associated with `ValueNotMacro` is "Not a macro monad value", but it's not even a `(Macro Syntax)` which is expected, it's a `(-> Syntax (Macro Syntax))`.
4. `transformerName` is bound to a value of type `(-> Syntax (Macro Syntax))`, but applying it to a `Syntax` value fails before we can get a `(Macro Syntax)` value out. The current code fails inside `expandEval` with `MacroEvaluationError`, whereas this branch fails here with `ValueNotMacro`. I think it should be `MacroEvaluationError`.
5. we do get a value out, but it's not a `(Macro Syntax)`, so `ValueNotMacro` is thrown. If this happens, that means there is a bug in Klister's type-checker, so `InternalError` or `error` would have been fine too.
6. `interpretMacroAction` encounters an error while executing the macro action. it fails somewhere inside `interpretMacroAction` via code which isn't relevant here.

As a conclusion, I think this should be `MacroEvaluationError`, not `ValueNotMacro`, but the existing code isn't very good at using the right error constructor, so this isn't a big deal.

src/Expander.hs
```diff
@@ -1400,12 +1415,17 @@ interpretMacroAction prob =
...
           s <- getState
           let env = fromMaybe Env.empty .
-                    view (expanderWorld . worldEnvironments . at phase) $
-                    s
-          value <- expandEval $ withEnv env $ apply closure boundResult
-          case value of
-            ValueMacroAction act -> interpretMacroAction prob act
-            other -> throwError $ ValueNotMacro other
+                    view (expanderWorld . worldEnvironments . at phase) $ s
+          case applyInEnv env closure boundResult of
```
gelisam, line 1419: another place where this MR highlights a mistake in the original code: all of
```
phase <- view (expanderLocal . expanderPhase)
s <- getState
let env = ...
```
and the `withEnv env` is dead code, because `apply` overwrites this `env` with its own call to `withEnv`. So there is no need for an `applyInEnv` separate from `apply`.

gelisam:
```suggestion
          phase <- view (expanderLocal . expanderPhase)
          case apply closure boundResult of
            Left err -> throwError
              $ MacroEvaluationError phase
              $ projectError err
```

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+-- 'Down': meaning that the evaluator is searching for a redex to evaluate and
+-- therefore moving "down" the AST.
+
+-- 'Up': meaning that the evaluator has evaluated some redex to a value and is
+-- passing that value "up" the execution stack.
+
+-- 'Er': meaning that something has gone wrong, the stack is captured and the Er
+-- will float up to be handled by the caller of the evaluator.
+
+All interesting things happen by matching on 'Kont', the continuation. This
```
gelisam, line 43: Not the Control? What's the Control for, then?

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+
+-- TODO: #108 describe the how the debugger hooks in
+
+-}
+
+module Evaluator
+  ( EvalError (..)
+  , EvalResult (..)
+  , TypeError (..)
+  , evaluate
```
gelisam, line 54: `evaluate` has type `Core -> Either EState Value`, but `EState` is not exported. That seems needlessly restrictive: the caller can store the result in a variable, return it to its own caller, but if they do so then they cannot write down a type signature. Would it make sense to at least export it as an opaque type?

gelisam, line 54: `evaluate` looks like it is the main API, with `evaluateIn` and `evaluateWithExtendedEnv` being a secondary APIs for special cases. But I see that `evaluate` is not used anywhere! Perhaps it would make more sense to drop `evaluate` and to make `evaluateIn` the main API?

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+  , evaluateIn
+  , evaluateWithExtendedEnv
+  , evalErrorType
+  , evalErrorText
```
gelisam, line 58: `evalErrorText` is not used anywhere? that's strange, where do the errors get printed? Maybe the `Show` instance is used instead of `evalErrorText` by mistake?

src/Evaluator.hs
gelisam: Thanks for the explanation, that makes sense. But the call to `error` still seems wrong. I can see several possible solutions:

1. Add a precondition to `projectError`'s documentation, e.g. "should only be called on the `EState` inside a `Left` returned by one of the `evaluateX` functions.
2. Have the `evaluateX` functions return an opaque representation of the error, let's say `ErrState`, containing `(err, env, k)`. Then, provide a total function `projectError :: ErrState -> EvalError` and a total function `ErrState -> EState`, so that the resulting `EState` can be given to `debugger :: EState -> DebugM EState`.
3. Have the `evaluateX` functions return `Left (err, estate)`, so they can both display the error and resume from the `estate`, without knowing that the two are related under the hood.

src/Expander.hs
gelisam: Now that I understand what `erroneousValue` does, I don't think this code is correct. It claims that the problem is that there is a value which should be a macro action, but its actual type is something else. But that problem is raised below, on L1001:
```
throwError $ MacroEvaluationError p $ evalErrorType "macro action" other
```

The problem on _this_ line is some unknown thing which happened during evaluation, not necessarily a type error. Instead of using `err` to construct a type error, we should propagate `err` upwards as-is.

src/Expander.hs
gelisam: This is the same mistake as the other `erroneousValue` call site. Maybe `erroneousValue` shouldn't exist at all, since it seems so easy to misuse?

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+  , projectError
+  , erroneousValue
+  , applyInEnv
+  , apply
```
gelisam, line 62: `apply` behaves very similarly to the `evaluateX` functions, would it perhaps make more sense to call it something like `evaluateFunctionApplication`?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
   | EvalErrorType TypeError
   | EvalErrorCase SrcLoc Value
```
gelisam, line 104: I know we discussed this, but I still find it unintuitive that a `SrcLoc` is needed for `case` and `type-case`, but not for anything else. We should document the reasoning somewhere. Let's see, that reasoning is...

That `SrcLoc` is used to construct the error message in the case in which no cases match. By comparison, other errors such as looking up a `Var` don't need a `SrcLoc` because scope errors are caught during expansion time, before evaluation begins. Are case and type-case really the only ways in which evaluation can fail at runtime, assuming expansion already caught scope errors and type errors?

src/Evaluator.hs
gelisam: same for `doTypeCase`. `evaluateTypeCase`?

src/Evaluator.hs
```diff
@@ -1,20 +1,77 @@
...
+  , evaluateIn
+  , evaluateWithExtendedEnv
+  , evalErrorType
+  , evalErrorText
+  , projectError
+  , erroneousValue
+  , applyInEnv
+  , apply
+  , doTypeCase
+  , try
```
gelisam, line 64: why is `Control.Exception.try` re-exported? you don't even throw exceptions in this file :sweat_smile:

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
   | EvalErrorType TypeError
   | EvalErrorCase SrcLoc Value
   | EvalErrorUser Syntax
-  deriving (Show)
+  | EvalErrorIdent Value
+  deriving (Show, Typeable)
 makePrisms ''EvalError
+instance Exception EvalError
```
gelisam, line 109: `EvalError` is never thrown as an exception, did you perhaps initially intend to use it as an exception and later changed your mind to use `Er evalError`?

src/Evaluator.hs
```diff
@@ -23,6 +80,14 @@ import Syntax.SrcLoc
 import Type
 import Value
 
+-- -----------------------------------------------------------------------------
+-- Interpreter Data Types
+
+
+data EvalResult
+  = ExampleResult SrcLoc VEnv Core (Scheme Ty) Value
+  | IOResult (IO ())
```
gelisam, line 89: `EvalResult` is not even used in this file :sweat_smile: 
Perhaps it should be defined next to `evalMod` instead?

Let's do that in a separate MR, as this MR does not introduce `EvalResult`, it merely moves it to the top of the file.

src/Evaluator.hs
```diff
@@ -23,6 +80,14 @@ import Syntax.SrcLoc
...
+
 -- TODO: more precise representation
 type Type = Text
```
gelisam, line 92: Text is actually just fine since the type checker has (hopefully) already caught all the type errors, so it doesn't make sense to spend extra effort on type errors which slip through to evaluation-time.

src/Evaluator.hs
gelisam: Actually, the current datatype exhaustively lists all the errors which can happen at evaluation-time, and #237 already classifies which ones should already be caught in an earlier phase and which ones really can happen at evaluation-time. There is `EvalErrorCase`, and `EvalErrorUser`. `EvalErrorUser` does not need a separate `SrcLoc` because the error message is a `Syntax`, which is already associated with a `SrcLoc`.

Since we're separating between `Case` and `TypeCase` everywhere else in the codebase, I wonder why there isn't a separate `EvalErrorTypeCase` here? :thinking:

Anyway, that's not something which changed in the current MR, moving right along...

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+-- of evaluation in the CEK machine is Foo. For example, when the continuation
+-- is 'InArg' the subject of evaluation is the argument of the function and the
+-- continuation holds the un-evaluated function symbol in its first field.
+data Kont where
+
+  Halt :: Kont
+  -- ^ Marks the evaluator finishing
+
+  -- functions
+  InArg :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
```
gelisam, line 121: Why use `CoreF TypePattern ConstructorPattern Core` instead of the shorter-and-isomorphic `Core`?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+  -- ^ The argument is being evaluated, so hold onto the function symbol.
+  InFun :: !Value -> !VEnv -> !Kont -> Kont
+  -- ^ The function is being evaluated, so hold onto the evaluated argument.
```
gelisam, line 124: Hmm, you evaluate the argument first? I of course know that Klister is a strict language, which means that the arguments are evaluated before the call is made, but what about the case in which the function is an expression, e.g. `((id (+)) (* 2 2) (* 3 3))`? Isn't it standard to evaluate `(id (+))` to `(+)` before evaluating `(* 2 2)` to `4` and `(* 3 3)` to `9`?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+-- continuation holds the un-evaluated function symbol in its first field.
+data Kont where
+
+  Halt :: Kont
+  -- ^ Marks the evaluator finishing
+
+  -- functions
+  InArg :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  -- ^ The argument is being evaluated, so hold onto the function symbol.
+  InFun :: !Value -> !VEnv -> !Kont -> Kont
```
gelisam, line 123: Why is the `VEnv` needed? It will not be used to further evaluate the `Value`, since it is already a `Value`. Is it perhaps so that the debugger can display the captured environment at this point in the stack trace?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+
+  -- Cases
+  InCaseScrut     :: ![(SyntaxPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+  InDataCaseScrut :: ![(ConstructorPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+  InTypeCaseScrut :: ![(TypePattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
+
+  -- lists
+  InConsHd :: !Core -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  InConsTl :: !Core -> !Syntax -> !VEnv -> !Kont -> Kont
+  InList   :: !Core -> ![Core] -> ![Syntax] -> !VEnv -> !Kont -> Kont
```
gelisam, line 140: It's a shame that you had to waste your time reimplementing those outdated direct syntax-manipulating primitives! I only ever use `open-syntax` and `close-syntax` now. We should get rid of this old stuff. In a later MR, of course.

src/Evaluator.hs
gelisam, line 124: These comments are really helpful, let's add one for each constructor!

```
  -- functions
  InArg :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
  -- ^ The argument is being evaluated, so hold onto the function expression
  InFun :: !Value -> !VEnv -> !Kont -> Kont
  -- ^ The function is being evaluated, so hold onto the evaluated argument.

  InLetDef :: !Ident -> !Var -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
  -- ^ Inside the rhs of (let [x rhs] body), so hold the name x and the body.

  -- constructors
  InCtor :: !Constructor -> ![CoreF TypePattern ConstructorPattern Core] -> ![Value] -> !VEnv -> !Kont -> Kont
  -- ^ Inside arg_i of (Ctor arg_1 ... arg_i ... arg_n), so hold Ctor, the
  -- already evaluated values value_1 ... value_{i-1}, and the
  -- yet-to-be-evaluated expressions arg_{i+1} ... arg_n.

  -- Cases
  InCaseScrut     :: ![(SyntaxPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
  InDataCaseScrut :: ![(ConstructorPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
  InTypeCaseScrut :: ![(TypePattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
  -- ^ Inside the scrutinee of (case scrutinee [pat_1 rhs_1] ...), so hold
  -- the branches.

  -- lists
  InConsHd :: !Core -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
  -- ^ Inside the head of (cons-list-syntax-cons head tail loc), so hold the tail.
  InConsTl :: !Core -> !Syntax -> !VEnv -> !Kont -> Kont
  -- ^ Inside the tail of (cons-list-syntax-cons head tail loc), so hold the already-evaluated head.
  InList   :: !Core -> ![Core] -> ![Syntax] -> !VEnv -> !Kont -> Kont
  -- ^ Inside arg_i of (list-syntax (arg_1 ... arg_i ... arg_n) loc), so hold
  -- the already-evaluated syntax values stx_1 ... stx_{i-1}, and the
  -- yet-to-be-evaluated expressions arg_{i+1} ... arg_n.

  -- idents
  InIdent :: !Core -> !VEnv -> !Kont -> Kont
  InIdentEqL :: !HowEq -> !Core -> !VEnv -> !Kont -> Kont
  -- ^ Inside the lhs of (<how>-identifier=? lhs rhs), so hold <how> and rhs.
  InIdentEqR :: !HowEq -> !Value -> !VEnv -> !Kont -> Kont
  -- ^ Inside the rhs of (<how>-identifier=? lhs rhs), so hold <how> and the
  -- already-evaluated lhs.

  -- Macros
  InPureMacro :: !VEnv -> !Kont -> Kont
  -- ^ Inside the expr of (pure expr), so nothing to hold.
  InBindMacroHd :: !Core -> !VEnv -> !Kont -> Kont
  -- ^ Inside the expr1 of (>>= expr1 expr2), so hold expr2.
  InBindMacroTl :: !MacroAction -> !VEnv -> !Kont -> Kont
  -- ^ Inside the expr2 of (>>= expr1 expr2), so hold the already-evaluated
  -- expr1.

  -- atomics
  InInteger :: !Core -> !VEnv -> !Kont -> Kont
  InString  :: !Core -> !VEnv -> !Kont -> Kont
  InLoc     :: !Core -> !VEnv -> !Kont -> Kont
  -- ^ Inside the loc of (replace-loc loc stx), so hold the stx.
  InLocStx  :: !SrcLoc -> !VEnv -> !Kont -> Kont
  -- ^ Inside the stx of (replace-loc loc stx), so hold the already-evaluated
  -- loc.

  -- scope
  InScope :: !(ExprF Syntax) -> !VEnv -> !Kont -> Kont
  -- ^ Inside the loc of something like (cons-syntax-list head tail loc), so
  -- remember the Syntax-Contents which needs a scope-set and location.

  -- logs and errors
  InLog   :: !VEnv -> !Kont -> Kont
  -- ^ Inside the expr of (log expr), so nothing to hold.
  InError :: !VEnv -> !Kont -> Kont
  -- ^ Inside the expr of (error expr), so nothing to hold.
```

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+  InArg :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  -- ^ The argument is being evaluated, so hold onto the function symbol.
+  InFun :: !Value -> !VEnv -> !Kont -> Kont
+  -- ^ The function is being evaluated, so hold onto the evaluated argument.
+
+  InLetDef :: !Ident -> !Var -> !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
+  -- ^ Evaluating the let def
+
+  -- constructors
+  InCtor :: !Constructor -> ![CoreF TypePattern ConstructorPattern Core] -> ![Value] -> !VEnv -> !Kont -> Kont
```
gelisam, line 130: According to the naming convention, shouldn't it be `InCtorArg`? We're inside `arg_i` of `(Ctor arg_1 ... arg_i ... arg_n)`, not inside `Ctor`, right?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+  -- Macros
+  InPureMacro :: !VEnv -> !Kont -> Kont
+  InBindMacroHd :: !Core -> !VEnv -> !Kont -> Kont
+  InBindMacroTl :: !MacroAction -> !VEnv -> !Kont -> Kont
+
+  -- atomics
+  InInteger :: !Core -> !VEnv -> !Kont -> Kont
+  InString  :: !Core -> !VEnv -> !Kont -> Kont
+  InLoc     :: !Core -> !VEnv -> !Kont -> Kont
+  InLocStx  :: !SrcLoc -> !VEnv -> !Kont -> Kont
```
gelisam, line 156: I think `InReplaceLocL` and `InReplaceLocR` would be much clearer than `InLoc` and `InLocStx`

src/Evaluator.hs
gelisam: And `arg_1` through `arg_{i-1}` have already been evaluated to `Value`s. Shouldn't the `[Value]` list be before the `[Core]` list? Seems weird to use the order `[arg_{i+1}, ..., arg_n] [arg_1, ..., arg_{i-1}]`. Or are both lists reversed? Is it `[arg_n, arg_{n-1}, ..., arg_{i+1}] [arg_{i-1}, arg_{i-2}, ..., arg_1]`?

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+                                       , _syntaxErrorLocations = mempty
+                                       })) env kont
+           (l:ls) -> Down (unCore l) env (InSyntaxErrorLocations msg_syn ls mempty env kont)
+          )
+      (\err -> Er err env kont)
+    -- done
+    (InSyntaxErrorLocations msg_syn [] dones env kont) ->
+        Up (ValueMacroAction
+                $ MacroActionSyntaxError (SyntaxError { _syntaxErrorMessage   = msg_syn
+                                                      , _syntaxErrorLocations = dones
```
gelisam, line 353:
```suggestion
                                                      , _syntaxErrorLocations = reverse (v : dones)
```

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+  InError :: !VEnv -> !Kont -> Kont
+
+
+  InSyntaxErrorMessage   :: ![Core] -> !VEnv -> !Kont -> Kont
+  InSyntaxErrorLocations :: !Syntax -> ![Core] -> ![Syntax] -> !VEnv -> !Kont -> Kont
+
+-- | The state of the evaluator
+data EState where
+  Down :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> EState
+  -- ^ 'Down', we are searching the AST for a redux and building up the stack of
```
gelisam, line 172:
```suggestion
  -- ^ 'Down', we are searching the AST for a redex and building up the stack of
```

src/Evaluator.hs
```diff
@@ -38,8 +103,343 @@ data EvalError
...
+    -- lambdas and application
+    (CoreLam ident var body) ->
+      let lam = ValueClosure $ FO $ FOClosure
+            { _closureEnv   = env
+            , _closureIdent = ident
+            , _closureVar   = var
+            , _closureBody  = body
+            }
+      in Up lam env k
+    (CoreApp fun arg) -> Down (unCore arg) env (InArg (unCore fun) env k)
```
gelisam, line 411: `eval` was recurring on the function position first, not the argument, we think the CEK version should do the same

(currently pair-reviewing with @david-christiansen)

src/Evaluator.hs
```diff
@@ -50,313 +450,221 @@ evalErrorText (EvalErrorCase loc val) =
...
+projectError (Er err _env _k) = err
+projectError _                = error "debugger: impossible"
+
+erroneousValue :: EvalError -> Value
+erroneousValue (EvalErrorCase _loc v) = v
+erroneousValue (EvalErrorIdent v)     = v
+erroneousValue  _                     =
+  error $ mconcat [ "erroneousValue: "
+                  , "Evaluator concluded in an error that did not return a value"
+                  ]
```
gelisam, line 670: I don't like this partial function, I think every single call site is incorrect, but I think we should (finally!) merge this and then I'll open a separate MR to replace all the call sites with something more sensible.

src/Evaluator.hs
```diff
@@ -50,313 +450,221 @@ evalErrorText (EvalErrorCase loc val) =
...
+  in evaluateIn env _closureBody
+apply (HO prim) value = return $! prim value
+
+applyAsClosure :: VEnv -> Value -> Value -> Kont -> EState
+applyAsClosure e v_closure value k = case v_closure of
+    ValueClosure closure -> app closure
+    other                -> Er (evalErrorType "function" other) e k
+
+    where app (FO (FOClosure{..})) =
+            let env = Env.insert _closureVar _closureIdent value (_closureEnv <> e)
```
gelisam, line 517: This is a bug. When applying a closure, the body of the closure must be evaluated in the captured environment, not in the union of the captured environment plus the environment at the call site. That is, the following code should not be accepted because `y` is not in scope where `f` is defined:
```
f = \() -> y
g y = f ()
```

:x: this is a blocker.

src/Evaluator.hs
gelisam: Turns out it is needed by the `Er` constructor, in case an error occurs, for example if the function-position value did not evaluate to a Closure.

src/Evaluator.hs
gelisam: turns out I was wrong about some of those, as some of those hold a reversed list.

:pencil2: this is not a blocker, but after this MR is finally merged, I plan to come back and add those comments.

src/Evaluator.hs
doyougnu: I'm having a hard time writing a test case for this because the patch errors as expected with (I believe) the correct error message. Here is my translation to klister:

```
#lang kernel

(define f (lambda (ff) y))
(define g (lambda (y) (f)))

(example (g 2))
```

Note that I have to pass `ff` or else I get and error about no arguments:

```
❯ cat examples/non-examples/bad-lexical-env.golden
Expected 1 entries between parentheses, but got
  #[bad-lexical-env.kl:3.19-3.22]<()>
```

this is a sidenote but I'm surprised we cannot define a function of type `f : () -> <whatever>`.

Here is what main shows:

```
Test suite klister-tests: RUNNING...
All tests
  Golden tests
    bad-lexical-env: OK (0.01s)
...
❯ cat examples/non-examples/bad-lexical-env.golden
Unknown: #[bad-lexical-env.kl:4.24-4.25]<y>
```

Notice that this is what we should expect, an error saying `unknown y`.

and here is the patch:

```
❯ cabal test --test-show-details=streaming --test-options '-p bad-lexical-env --accept'
...
Test suite klister-tests: RUNNING...
All tests
  Golden tests
    bad-lexical-env: OK (0.01s)
      Accepted the new version
...
❯ cat examples/non-examples/bad-lexical-env.golden
Unknown: #[bad-lexical-env.kl:3.24-3.25]<y>
```

That is, I accept the new output for the golden test and we get the same error as `main` from the patch.

src/Evaluator.hs
doyougnu: Sounds good :+1:

src/Evaluator.hs
doyougnu: I believe this is a type error:

- `v : Value`
- `dones : [Syntax]`

here is what I believe the original case is in `eval`:

```
eval (Core (CoreSyntaxError syntaxErrorExpr)) = do
  syntaxErrorValue <- traverse evalAsSyntax syntaxErrorExpr
  pure $ ValueMacroAction
       $ MacroActionSyntaxError syntaxErrorValue
```

is there an example that exhibits this bug?

src/Evaluator.hs
doyougnu: Ah indeed this is an inconsistency, I believe I was trying to keep the value list in registers instead of spilling it onto the stack. But I doubt this actually matters in practice.

src/Evaluator.hs
doyougnu: - [ ] add comments for each constructor after merge.

src/Evaluator.hs
gelisam: > I'm having a hard time writing a test case for this

Makes sense: scope errors are caught at expansion time, precisely so that they cannot occur at evaluation-time, so it should not be possible to write a Klister program which triggers this bug. Okay, this isn't as big of an issue as I thought then.

✏️ this is no longer a blocker, but after this MR is finally merged, I plan to come back and fix this.

gelisam: Note that I have to pass ff or else I get and error about no arguments:

```
❯ cat examples/non-examples/bad-lexical-env.golden
Expected 1 entries between parentheses, but got
  #[bad-lexical-env.kl:3.19-3.22]<()>
```

this is a sidenote but I'm surprised we cannot define a function of type `f : () -> <whatever>`.

I agree that it is surprising, but it makes sense if you think about it. `(lambda () 42)` is a function from _zero_ arguments, not a function from the unit type to Integer. Racket supports functions from zero arguments, but Haskell does not. Like Haskell, Klister only supports functions of one argument, so `(lambda (x y) 42)` desugars to `(lambda (x) (lambda (y) 42))` and if it wasn't rejected `(lambda () 42)` would desugar to `42`. The error message could definitely be improved though!

src/Evaluator.hs
gelisam: > I believe this is a type error: [...]

You're right, a call to `evalAsSyntax` is also needed.


> is there an example that exhibits this bug?

Here is one:
```
#lang kernel
(import (shift kernel 1))

(define-macros
  ([m (lambda (stx)
        (syntax-error 'error-message
                        'location-1  -- line 7
                        'location-2  -- line 8
                        'location-3  -- line 9
                        'location-4  -- line 10
                        ))]))

(example (m))
```

on main, this program outputs
```
toy.kl:7.26-7.36:
    Syntax error from macro:
    #[toy.kl:6.24-6.37]<error-message>Additional locations:
    toy.kl:8.26-8.36
    toy.kl:9.26-9.36
    toy.kl:10.26-10.36
```
whereas on this branch, this program outputs one fewer location, and outputs them in the opposite order:
```
toy.kl:9.26-9.36:
    Syntax error from macro:
    #[toy.kl:6.24-6.37]<error-message>Additional locations:
    toy.kl:8.26-8.36
    toy.kl:7.26-7.36
```
The first location is distinguished because it is printed before the error message, so let's use the shorthand `[7], 8, 9, 10` to describe the output on main and `[9], 8, 7` to describe the output on this branch.

Line 10 (`'location-4`) is missing, because `step (Up v e k)` is supposed to insert the value `v` into the context `k`, but here when the context is `InSyntaxErrorLocations`, the `v` is not inserted, it is dropped on the floor.

And the locations are listed in the opposite order because as you traverse `InSyntaxErrorLocations`'s `[Core]` from left to right, you prepend each output to its `[Syntax]`, so the right-most `Core` ends up being the leftmost `Syntax`. You have fixed this elsewhere by calling `reverse` in the `InCtor []` and `InList []` cases, you just forgot to also do it in the `InSyntaxErrorLocations  []` case.

src/Evaluator.hs
gelisam: Wow! By all means, don't let me break your micro-optimizations, you are clearly having fun doing it! But when it is that obscure, please document them because otherwise I will most certainly break them by accident.

src/Evaluator.hs
doyougnu: great explanation! You're right, I guess I see some scheme or lisp and assume that I can define a function with 0-arity.

gelisam (approved):
Approved! We can finally merge this! Sorry it took so long.

doyougnu:
> Approved! We can finally merge this! Sorry it took so long.

No worries, it was a thorough review!

PR #226, opened 2024-01-21
Gelisam/code rabbit test6

gelisam:
The first test MRs (#224) included "Streamlined module declarations and imports for better maintainability" in its summary of the changes, which isn't particularly useful.

I added "Dont mention changes to the import statements" to its custom instructions, and the summary for the second test MR (#225), containing the same changes, was much better: "Implemented a state machine for the CEK interpreter". But its review comments didn't do anything to help me understand the new CEK interpreter, they only complained about the lack of error-handling, which is not very useful either since the `Expand` monad is already propagating errors automatically.

Let's see what CodeRabbit emits with the following instructions:

> Dont mention changes to the import statements nor to the language extensions at the top of the file.
> 
> Don't point out missing error-handling code, the type-checker already detects those mistakes at compile-time.
> 
> Do explain what the new code does and how changes to existing functions affect their behavior.

## Summary by CodeRabbit

- **New Features**
	- Implemented a new CEK interpreter for enhanced language evaluation.
- **Refactor**
	- Simplified the REPL's exception handling and evaluation logic.
	- Streamlined the evaluation process with a new `evaluate` function.
- **Documentation**
	- Added module-level comments explaining the CEK interpreter.
- **Bug Fixes**
	- Adjusted error handling and value evaluation for consistency and reliability.


gelisam:
> Dont mention changes to the import statements nor to the language extensions at the top of the file.
> 
> Don't point out missing error-handling code, the type-checker already detects those mistakes at compile-time.
> 
> Do explain what the new code does and how changes to existing functions affect their behavior.

Well, CodeRabbit still mentioned "Added import" and "the introduction of new language extensions". It did not explain what the new code does, I guess that's not what CodeRabbit is designed to do.

CodeRabbit still spuriously complains about error-handling, but only once, which is a big improvement over the second test MR (https://github.com/gelisam/klister/pull/225), which did this 9 times!

gelisam:
Expanding the "Additional comments: 25" blocks shows "[...] Since the PR instructions specify not to mention error handling, this change is assumed to be intentional and correct". It is reassuring to see that the AI did take my instructions into account!

src/Expander/Monad.hs
```diff
@@ -860,10 +861,10 @@ setTasks = modifyState . set expanderTasks
...
 clearTasks = modifyState $ set expanderTasks []
 
-expandEval :: Eval a -> Expand a
+expandEval :: Core -> Expand Value
 expandEval evalAction = do
   env <- currentEnv
-  out <- liftIO $ runExceptT $ runReaderT (runEval evalAction) env
+  let out = Right $ evaluateIn env evalAction
   case out of
     Left err -> do
```
gelisam, line 869: @coderabbitai in Haskell, it is really strange to use an explicit constructor (in this case `Right`) and then to immediately pattern-match to see if the constructor is `Right` or something else. Please flag this kind of code in the future.

src/Evaluator.hs
```diff
@@ -310,53 +481,91 @@ doTypeCase blameLoc (Ty v0) ((p, rhs0) : ps) = match (doTypeCase blameLoc (Ty v0
...
+    match fk sk ((CtorPattern ctor subPats, tgt) : more) =
+      case tgt of
+        ValueCtor c args
+          | c == ctor ->
+            if length subPats /= length args
+              then error $ "Type checker bug: wrong number of pattern vars for constructor " ++ show c
+              else match fk sk (zip (map unConstructorPattern subPats) args ++ more)
+        _otherValue -> fk
+    match fk sk ((PatternVar n x, tgt) : more) =
+      match fk (sk . extend n x tgt) more
```
gelisam, line 557: @coderabbit could you please find the previous implementation of `doDataCase` and show me the diff?

gelisam, line 557: @coderabbitai could you please find the previous implementation of doDataCase and show me the diff?

gelisam: @coderabbitai thanks, but this diff interleaves `doCase` and `doDataCase`, thus making it hard to see the differences. Don't use GitHub's diff, calculate a new diff by comparing the previous implementation of `doDataCase` with the new implementation.

gelisam: > Here is the previous implementation of doDataCase from the commit just before the PR changes

@coderabbitai which commit is that? I think it should be 67d4eedf8a2c7b4ca67ac01e8813052c29cd0557, which looks quite different:

```haskell
doDataCase :: SrcLoc -> Value -> [(ConstructorPattern, Core)] -> Eval Value
doDataCase loc v0 [] = throwError (EvalErrorCase loc v0)
doDataCase loc v0 ((pat, rhs) : ps) =
  match (doDataCase loc v0 ps) (eval rhs) [(unConstructorPattern pat, v0)]
  where
    match ::
      Eval Value {- ^ Failure continuation -} ->
      Eval Value {- ^ Success continuation, to be used in an extended environment -} ->
      [(ConstructorPatternF ConstructorPattern, Value)] {- ^ Subpatterns and their scrutinees -} ->
      Eval Value
    match _fk sk [] = sk
    match fk sk ((CtorPattern ctor subPats, tgt) : more) =
      case tgt of
        ValueCtor c args
          | c == ctor ->
            if length subPats /= length args
              then error $ "Type checker bug: wrong number of pattern vars for constructor " ++ show c
              else match fk sk (zip (map unConstructorPattern subPats) args ++ more)
        _otherValue -> fk
    match fk sk ((PatternVar n x, tgt) : more) =
      match fk (withExtendedEnv n x tgt $ sk) more
```

src/Expander/Monad.hs
gelisam: Please remember this using the Learning feature.

gelisam: @coderabbitai please remember this using the Learning feature.

Issue #233, opened 2024-03-27
Follow HLint suggestions?


philderbeast:
Running `hlint-3.8`, I'm finding 148 hints. Are you open to following these suggestions? If so then I'd be happy to contribute those changes.

- [ ] Avoid lambda, 1 hint
- [ ] Eta reduce, 12 hints
- [ ] Fuse foldr/map, 1 hint
- [ ] Move brackets to avoid $, 6 hints
- [ ] Redundant $, 32 hints
- [ ] Redundant <$>, 17 hints
- [ ] Redundant bracket, 12 hints
- [ ] Replace case with fromMaybe, 3 hints
- [ ] Use $>, 2 hints
- [ ] Use <&>, 1 hint
- [ ] Use >=>, 1 hint
- [ ] Use asks, 5 hints
- [ ] Use camelCase, 2 hints
- [ ] Use const, 4 hints
- [ ] Use evalState, 2 hints
- [ ] Use fold, 1 hint
- [ ] Use for, 1 hint
- [ ] Use fromMaybe, 1 hint
- [ ] Use id, 6 hints
- [ ] Use newtype instead of data, 3 hints
- [ ] Use record patterns, 4 hints
- [ ] Use tuple-section, 2 hints
- [ ] Use uncurry, 2 hints
- [ ] Use unless, 3 hints
- [ ] Use unwords, 1 hint
- [ ] Use void, 5 hints
- [ ] Use when, 2 hints

doyougnu:
Yes please! We are thankful for the help. Most of these should be extremely minor changes. The only one that will be difficult and that I think you should avoid are the incomplete pattern matches in the Evaluator. Fixing these will mean re-architecting the Evaluator and is just not worth it at this time.

philderbeast:
> I seem to have a different version of hlint than you, because my hlint report only finds one hint

Yes that is likely. If we go for an action, the hlint version can be set there.

gelisam:
Ah, so if we merge #234, hlint will not pester us about any of the suggestions listed there! I had misunderstood the purpose of that list, I thought it was the opposite. Sure, let's merge it!

PR #234, opened 2024-03-27
Generate hlint default configuration

philderbeast:
See #233. This is the `.hlint.yaml` configuration generated by `hlint --default . > .hlint.yaml` with stock comments removed.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,29 @@
+# Warnings currently triggered by your code
+- ignore: {name: "Avoid lambda"} # 1 hint
+- ignore: {name: "Eta reduce"} # 12 hints
+- ignore: {name: "Fuse foldr/map"} # 1 hint
+- ignore: {name: "Move brackets to avoid $"} # 6 hints
+- ignore: {name: "Redundant $"} # 32 hints
+- ignore: {name: "Redundant <$>"} # 17 hints
+- ignore: {name: "Redundant bracket"} # 12 hints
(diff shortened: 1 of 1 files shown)


doyougnu:
I'm not opposed to silencing the warnings, but I do like seeing them because some code should certainly be refactored. Thoughts @gelisam @david-christiansen

david-christiansen:
As someone without time to really work on this project, my opinion doesn't matter that much. But here it is: I think hlint is great, but many of its defaults are not so useful. In particular, it pushes for a very point free style, which is harder to work with in practice. If I were active on the project, I'd want to go in and enable the suggestions a few at a time, considering whether they actually help, rather than just turning it all on and believing what it says, but as I said, I'm not active right now!

gelisam:
> See https://github.com/gelisam/klister/issues/233.

I was fine with #233 because the Klister project would benefit from more contributors, and applying hlint suggestions seemed like an easy and inoffensive way for someone new to the codebase to get their feet wet without having to spend a long time understanding how everything works. Now that @david-christiansen points out that hlint "pushes for a very point free style", which I agree is "harder to work with in practice", I am not so convinced that #233 is a good idea.

@philderbeast if you want to contribute to the Klister project and you are looking for an easy ticket to start with, may I recommend looking at the [good first issue](https://github.com/gelisam/klister/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) label?

gelisam:
Merging as per https://github.com/gelisam/klister/issues/233#issuecomment-2077336083

PR #249, opened 2024-08-30
wrap to terminal width

gelisam:
We are using the prettyprinter library, which nicely wraps and indents the output to fit the screen, but we were using the default layout options, which assume that the screen is always exactly 80 characters wide. With a slightly sinful unsafePerformIO, we can use the terminal's actual width instead.

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic layout configuration for pretty-printing based on terminal width, enhancing document rendering flexibility.
- **Bug Fixes**
	- Improved handling of environment variables for layout options, ensuring a default of 80 columns when necessary.
- **Documentation**
	- Updated import statements to reflect new dependencies for environment variable access and string parsing.

### DIFF
--- a/src/Pretty.hs
+++ b/src/Pretty.hs
@@ -11,6 +11,7 @@ module Pretty (Doc, Pretty(..), string, text, viaShow, (<+>), (<>), align, hang,
 import Control.Lens hiding (List)
 import Control.Monad.State
 import qualified Data.HashMap.Strict as HM
+import Data.Maybe (fromMaybe)
 import qualified Util.Set as Set
 import Prettyprinter hiding (Pretty(..), angles, parens)
 import qualified Prettyprinter as PP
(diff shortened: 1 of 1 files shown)


src/Pretty.hs
```diff
@@ -58,8 +62,21 @@ angles doc = text "⟨" <> align (group doc) <> "⟩"
 vec :: Doc ann -> Doc ann
 vec doc = text "[" <> align (group doc) <> "]"
 
+customLayoutOptions :: LayoutOptions
+customLayoutOptions = unsafePerformIO $ do
```
gelisam, line 66: Using `unsafePerformIO` is the most straightforward way to do it, but of course I expect that somebody will object and that I will have to thread a `LayoutOptions` variable everywhere.

gelisam: One concern is that the tests use `pretty`, so they might start failing depending on the terminal width. But I tried making my terminal very small and the tests still passed :shrug:

gelisam: Oh, that's because it doesn't work. `echo $COLUMNS` prints the right value in the terminal, but `lookupEnv "COLUMNS"` returns `Nothing`!

PR #251, opened 2024-10-03
forbid kind loops

### LINKED ISSUE #250: infinite loop during type-checking
Somehow, giving the compiler some help in figuring out the type of `(mk-phantom-1)` causes it to loop forever :(

```
#lang kernel


(datatype (Phantom A)
  (mk-phantom-1))

-- correctly infers ∀α. (Phantom α)
(define mk-phantom-2
  (with-unknown-type [A]
    (the A
      (mk-phantom-1))))

-- loops forever
(define mk-phantom-3
  (with-unknown-type [A]
    (the (Phantom A)
      (mk-phantom-1))))
```

gelisam:
fixes #250

### DIFF
--- a/src/Expander/TC.hs
+++ b/src/Expander/TC.hs
@@ -306,15 +306,31 @@ typeVarKind ptr =
     Just v -> pure $ view varKind v
 
 
-setKindVar :: KindVar -> Kind -> Expand ()
-setKindVar v k@(KMetaVar v') =
-  (view (expanderKindStore . at v') <$> getState) >>=
-  \case
-    Nothing -> modifyState $ set (expanderKindStore . at v) (Just k)
-    -- Path compression step
-    Just k' -> setKindVar v k'
-setKindVar v k = modifyState $ set (expanderKindStore . at v) (Just k)
+-- pre-condition: 'v' zonks to itself.
+-- post-condition: @KMetaVar v@, 'k', and the return value all zonk to the same
+-- kind.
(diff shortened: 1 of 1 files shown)


gelisam:
`CI / ubuntu-latest / ghc 8.10.7` failed to build `happy` for some reason. if the other CI targets pass, I plan to drop support for ghc-8.10.7.

gelisam:
I added tests to `non-examples` in order to make sure that the occurs-check triggers when it should, but unfortunately the error message includes a random number which is different on every run, thus causing the golden tests to fail. Perhaps the pretty printer can rename them to `?1` and `?2`, in the same way that type variables are renamed to α and β?

       --- examples/non-examples/type-occurs-check.golden	2024-10-04 13:03:50.980505569 +0000
      +++ /tmp/type-occurs-check62503-85.actual	2024-10-04 13:10:45.244375201 +0000
      @@ -1 +1 @@
      -Infinite type detected: (MetaPtr 57868) = (List META(MetaPtr 57868))
      +Infinite type detected: (MetaPtr 334228) = (List META(MetaPtr 334228))

gelisam:
> in the same way that type variables are renamed to α and β?

I thought type variables were represented by random numbers under the hood, but [that is not the case](https://github.com/gelisam/klister/blob/a99bf39d73ec0d4364d0770182fc1b398806972e/src/Pretty.hs#L346), so the type variables are not a good source of inspiration after all.

PR #261, opened 2025-01-04
basic stack traces

doyougnu:
Todos:

- [ ] add tests for `>>=` tail and `pair`
- [ ] refine the stack trace output, some of them are a bit wonkey
- [ ] no more todos in the code
- [ ] add some commentary
- [ ] remove the printing of the unique ids, they are not deterministic

@gelisam  @david-christiansen, sorry for the size of this one. I wanted to open this to get some eyes on it. 

How can I test the `InMacroBindTL` and `pair` cases? No matter what kind of klister I wrote, I could not get those to show up.

Also, you'll notice that I chose to add a hardline to separate the error message from the error context. We should decide on a format in the next meeting


src/Pretty.hs
```diff
@@ -98,73 +110,74 @@ instance (PrettyBinder VarInfo typePat, PrettyBinder VarInfo pat, Pretty VarInfo
       Nothing -> string ("!!" ++ show v ++ "!!")
       Just (Stx _ _ x) -> text x
   pp env (CoreLet x@(Stx _ _ y) v def body) =
-    hang 2 $ group $
```
david-christiansen, line 101: This seems incorrect - surely the `let` should be grouped?

src/Pretty.hs
```diff
@@ -98,73 +110,74 @@ instance (PrettyBinder VarInfo typePat, PrettyBinder VarInfo pat, Pretty VarInfo
...
   pp env (CoreDataCase _ scrut cases) =
-    hang 2 $ group $
+    hang 2 $
     vsep [ text "case" <+> pp env scrut <+> "of"
          , encloseSep (flatAlt mempty (text "{" <> space))
                       (flatAlt mempty (space <> text "}"))
                       (flatAlt mempty (space <> text ";" <> space)) $
            map (\(pat, rhs) ->
                    let (ppPat, env') = ppBind env pat
-                   in hang 2 $ group $
```
david-christiansen, line 133: It seems to me that this is a strange change. Shouldn't the LHS and RHS of the pattern be on one line if they fit?

That is,

```
case foo of
  0 |-> 2
  1 |-> 5
```
rather than
```
case foo of
  0 |->
    2
  1 |->
    5
```
?

src/Pretty.hs
```diff
@@ -309,8 +322,8 @@ instance Pretty VarInfo (Scheme Ty) where
     pp env t
   pp env (Scheme argKinds t) =
     text "∀" <>
```
david-christiansen, line 324: This should prevent many of the confusing newline issues in the tests:
```suggestion
   group $ text "∀" <>
```

Lining up the bindings is nice - but shouldn't they also get grouped, so we can get
```
forall (x : A) (y : B). C
```
instead of always breaking lines?

src/Pretty.hs
```diff
@@ -415,7 +428,7 @@ instance (Pretty VarInfo s, Pretty VarInfo t, PrettyBinder VarInfo a, Pretty Var
   ppBind env (Example loc t e) =
     (hang 4 $
      text "example@" <> pp env loc <+>
-     align (group (vsep [ group (pp env e) <+> text ":"
```
david-christiansen, line 418: This seems definitely to be a regression - surely the two sides of the type ascription should avoid breaking if a break on the colon fixes it?

src/Pretty.hs
```diff
@@ -663,15 +677,157 @@ instance Pretty VarInfo Scope where
...
+printStack e (Er err _env k) =
+  vsep [ pp e err
+       , text "stack trace:"
+       ] <> pp e k
+printStack _ Up{}   = hang 2 $ text "up"
+printStack _ Down{} = hang 2 $ text "down"
+
+printKont :: Env Var () -> Kont -> Doc VarInfo
+-- the basics
+printKont _ Halt               = text "Halt"
```
david-christiansen, line 712: In empty traces, this was a bit confusing. What about suppressing it, and then special-casing empty continuations at the outside?

src/Pretty.hs
```diff
@@ -663,15 +677,157 @@ instance Pretty VarInfo Scope where
...
+printKont e (InDataCasePattern p k) =
+  let ppPattern = \case
+        CtorPattern c _ps  -> pp e c
+        PatternVar i _v    -> pp e i
+  in text "in data case pattern: "
+     <> ppPattern (unConstructorPattern p)
+     <> pp e k
+
+-- pairs
+-- TODO: DYG: how to test the cons?
```
david-christiansen, line 754: What about adding `unsafeCoerce` as a primitive for testing?

david-christiansen (commented):
The implementation looks like a great application of the CEK-style evaluator!

Two things:
1. The pretty-printer changes seem to be almost entirely regressions, grouping unrelated things and separating related things. They also make the diff harder to follow, because there's tons of noise in tests. Reverting most of the deletions of `group` should fix that, I think. What problem was it solving?
2. Why is there a list of `EState`s in a debug context? It seems to me that the debugger should get a single evaluator state.

examples/ambiguous-kind.golden
```diff
@@ -1 +1,2 @@
-(ctor) : ∀(α : *). (T α)
+(ctor) : ∀(α : *).
+ (T α)
```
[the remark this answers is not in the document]
doyougnu: > pretty-printer changes seem to be almost entirely regressions

Yes, I'll revert the changes. I was trying to get a clean output to observe the stack traces and if I recall correctly a `group` call was preventing the indentation I was after

> Why is there a list of EStates in a debug context?

I agree, I believe I was anticipating the need to push contexts in the task scheduler, but I don't really remember to be honest. In this implementation, stack traces are blind to the tasks, so we will only get a stack trace from the last evaluation.

examples/non-examples/stack-traces/error-in-constructor.golden
```diff
@@ -0,0 +1,9 @@
+Error at phase p0:
+    error-in-constructor.kl:17.45-17.56:
+      Im-an-error
+    stack trace:
+    ---- in constructor pair
+                        in field 2
+    ---- in let #[error-in-constructor.kl:17.22-17.24]
+                 <go>!!(Var 395597)!!
```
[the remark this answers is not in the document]
doyougnu: yes and I need to add some machinery to track the names via the ids. Var is the internal name and `go` is in the source

examples/ambiguous-kind.golden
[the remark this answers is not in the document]
doyougnu: I've reverted

doyougnu:
Alright I was able to show the names of the value closures by adding some `Text` to `HO`. The problem now is that the error messages for the tests seem much more informative (e.g. instead of `#<closure>` we now output the name of the closure) except for `primitives-documentation` which seems to be somewhat different.

gelisam:
> except for primitives-documentation which seems to be somewhat different and a regression.

It makes sense that `primitives-documentation` would behave differently, since most of those primitives are `HO` primitive functions implemented in Haskell rather than `FO` user-defined functions defined in Klister.

> e.g. instead of #<closure> we now output the name of the closure)
> [...]
> (define fail (lambda (thing) [...]))
> [...]
> stack trace:
>   ---- with function #<thing>

hmm, that's the name of the argument, not the name of the function. While I understand that the name of the argument is easier to obtain since it's right there in the lambda, I am sure that the user would prefer to see `in function "fail"`.

How about using the SrcLoc of the lambda to identify it? This would work even if the lambda is never given a name.

For this reason I think it makes perfect sense for both lambdas in lessThan to have the name `#<lessThan>`. It's a primitive which does not have a SrcLoc associated with it!

doyougnu:
> How about using the SrcLoc of the lambda to identify it? This would work even if the lambda is never given a name.

Yes I wanted to go down this path with `Ident`, but I think that only makes sense for functions that are defined in klister not the prims defined in Haskell. That led me down the path of defining HO as `HO Text ...`. 

> hmm, that's the name of the argument, not the name of the function. While I understand that the name of the argument is easier to obtain since it's right there in the lambda, I am sure that the user would prefer to see in function "fail".

totally agree!

doyougnu:
On second thought, I think that this printing `thing` is a bug. It prints the argument name because of this line:

```
printKont e (InArg fun _env k) = text "with function" <+> pp e fun <> pp e k
```
In an `InArg` continuation `fun` should be the function symbol, not the argument symbol because the argument symbol would be in the focus of the CEK machine (the C). But this is showing us that this is not the case. Instead of the function symbol `fail` we are getting the argument `thing`!

doyougnu:
Note that if instead of requiring the function symbol being fully evaluated before the argument like this:

```
  -- functions
  InArg :: !(CoreF TypePattern ConstructorPattern Core) -> !VEnv -> !Kont -> Kont
  -- ^ The argument is being evaluated, so hold onto the function symbol.
  InFun :: !Value -> !VEnv -> !Kont -> Kont

    -- functions
    -- we evaluated the arg to get a closed so now we evaluate the fun
    (InArg fun env kont) -> Down fun env (InFun v e kont)
    -- we evaluated the fun so now do the application
    (InFun arg env kont) -> applyAsClosure env v arg kont
```

then I get this output:

```
FAIL (0.08s)
      Test output was different from 'examples/non-examples/stack-traces/error-in-arg.golden'. Output of ["diff","-u","examples/non-examples/stack-traces/error-in-arg.golden","/tmp/error-in-arg32423-0.actual"]:
      --- examples/non-examples/stack-traces/error-in-arg.golden	2025-02-16 12:10:16.161856036 -0500
      +++ /tmp/error-in-arg32423-0.actual	2025-02-16 13:10:19.226759431 -0500
      @@ -1,5 +1,5 @@
       Error at phase p0:
           error-in-arg.kl:5.24-5.27: bad
           stack trace:
      -    ---- with function #<thing>
      +    ---- with function !!(Var 2097)!!
           ---- Halt

1 out of 1 tests failed (0.08s)
```

doyougnu:
> On second thought, I think that this printing thing is a bug. It prints the argument name because of this line:

This is not a bug. A closure is an expression with a bunch of variables that and one extra variable. The other variables are closed by the captured environment. The extra variable is from the definition of the closure, recall that in the kernel lang each closure must at least be unary. So this was my misunderstanding.

doyougnu:
alright I've fixed the names, now more tests and cleanup and this should be good to go

examples/contract.golden
```diff
@@ -1,5 +1,5 @@
-#<closure> : (Syntax → Syntax)
-#<closure> : (Syntax → (Syntax → (Syntax → (Macro Syntax))))
-#<closure> : (Syntax → (Syntax → (Syntax → (Macro Syntax))))
-#<closure> : ∀(α : *). (α → α)
+#<contract> : (Syntax → Syntax)
+#<arg> : (Syntax → (Syntax → (Syntax → (Macro Syntax))))
```
doyougnu, line 2: looks like things defined with `defun` are still capturing the argument not the function name

doyougnu: - [ ] : TODO fix defun

doyougnu: I think these do not make it to the evaluator. They are wired into the expander in the kernel. This means they are defined as `decl`s here:
```
evalMod (Expanded em _) = execStateT (traverseOf_ (moduleBody . each) evalDecl em) mempty
  where
    evalDecl :: CompleteDecl -> StateT (Seq EvalResult) Expand ()
    evalDecl (CompleteDecl d) =
...
```

doyougnu: And the `Decls` are not represented by `Core` so the evaluator typically only sees the body of the `Decl` which makes sense. This is a good use case for a passthrough Kont. But I'll save that for a different PR.

doyougnu:
Basically done, I want to do another pass over the code then squash

src/Pretty.hs
```diff
@@ -608,7 +619,8 @@ instance Pretty VarInfo (ExprF Syntax) where
     pure $ parens (group (vsep ppXs))
 
 instance Pretty VarInfo Closure where
-  pp _ _ = pure $ text "#<closure>"
+  pp _ (FO fo) = pure $ "#<" <> text (_stxValue (_closureIdent fo)) <> ">"
+  pp _ (HO n _) = pure $ "#<" <> text n <> ">"
```
doyougnu, line 623: FYI here are the changes to the pretty printer that print the closure name rather than the type "closure". In some cases this seems desirable, but the more I look at it having "foo<closure>" also seems reasonable. Thoughts?

src/Value.hs
```diff
@@ -52,7 +52,11 @@ primitiveCtor name args =
   in ValueCtor ctor args
 
 valueText :: Value -> Text
-valueText (ValueClosure _) = "#<closure>"
+valueText (ValueClosure c) = "#<" <> the_closure <> ">"
+  where
+    the_closure = case c of
+                    (FO fo) -> _stxValue $ _closureIdent fo
+                    (HO n _) -> n
```
doyougnu, line 59: and here

examples/non-examples/bad-lexical-env.golden
```diff
@@ -0,0 +1 @@
+Unknown: #[bad-lexical-env.kl:3.24-3.25]<y>
```
gelisam, line 1: this should succeed, no? g is a function which returns another function, and that seems fine.

examples/non-examples/stack-traces/error-in-bind-head.golden
```diff
@@ -0,0 +1,7 @@
+Error at phase p0:
+    error-in-bind-head.kl:4.22-4.33: Im-an-error
+    stack trace:
+    ---- in pure macro
+    ---- in bind macro head
+    λx. pure x
```
gelisam, line 6: this is the tail of the bind, not the head

examples/non-examples/stack-traces/error-in-bind-tail.kl
```diff
@@ -0,0 +1,8 @@
+#lang "prelude.kl"
+
+(example
+  (>>= (pure 'hello-go-boom)
+    (lambda (x)
+      (>>= (pure x)
+      -- TODO: DYG: see #265
```
gelisam, line 7: the golden file's output looks correct to me: this expression constructs a value of type Macro α which gets pretty-printed as `(pure ...) >>= (lambda (x) ...)`, where the `(lambda (x) ...)` is printed as `#<x>` thanks to your new pretty-printer which prints functions as the name of their argument.

examples/non-examples/stack-traces/error-in-bind-head.kl
```diff
@@ -0,0 +1,5 @@
+#lang "prelude.kl"
+
+(example
+  (>>= (pure (error 'Im-an-error))
+       (lambda (x) (pure x))))
```
gelisam, line 5: This attempts to construct a value of type `Macro a` but fails to do so because of the call to `error`. Thus, we get an error at evaluation-time, when trying to evaluate `(>>= ...)` to a `Macro a`. I suspect you were trying to construct an example where the error occurs during execution. We will thus need to wrap this code inside a macro definition, like this:
```
#lang "prelude.kl"

(define-macros
  [(m (lambda (stx)
        (>>= (pure (error 'Im-an-error))
          (lambda (x) (pure x)))))])

(m)
```
this fails with the same message,
```
Not a macro monad value: toy.kl:16.28-16.39: Im-an-error
stack trace:
---- in pure macro
---- in bind macro head
λx. pure x
---- Halt
```
except without the `Error at phase p0:` bit. which makes sense, because the error occurs during phase 1, when the body of the macro is being evaluated. but why doesn't it say `Error at phase p1:`? :thinking:

examples/non-examples/stack-traces/error-in-bind-tail.kl
gelisam: To make this fail, we would have to execute the value of type `Macro a`, like this:
```
#lang "prelude.kl"

(define-macros
  [(m (lambda (stx)
        (>>= (pure 'hello-go-boom)
          (lambda (x)
            (>>= (pure x)
            -- TODO: DYG: see #265
              (lambda (y) (error y)))))))])

(m)
```
this fails with
```
Not a macro monad value: toy.kl:16.21-16.34: hello-go-boom
stack trace:
---- Halt
```
the stack trace is empty because the executor successfully calls the evaluator for `(pure 'hello-go-boom)`, then replaces `x` with `'hello-go-boom` in `(pure x)`, then successfully calls the evalutor for ` (pure 'hello-go-boom)` again, then replaces `y` with `'hello-go-boom` in `(error y)`, so the evaluator only has `(error 'hello-go-boom)` to work with, and that is why the stack trace is empty.

gelisam: another interesting example is
```
#lang "prelude.kl"

(define-macros
  [(m (lambda (stx)
        (>>= (syntax-error 'bad)
          (lambda (x)
            (>>= (pure x)
            -- TODO: DYG: see #265
              (lambda (y) (error y)))))))])

(m)
```
which fails with
```
Syntax error from macro: #[toy.kl:8.29-8.32]<bad>
```
there is no stack trace because the error is raised by the executor, not the evaluator, and the executor does not print stack traces yet.

src/Evaluator.hs
```diff
@@ -135,6 +160,15 @@ data Kont where
   InDataCaseScrut :: ![(ConstructorPattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
   InTypeCaseScrut :: ![(TypePattern, Core)] -> !SrcLoc -> !VEnv -> !Kont -> Kont
 
+  {- Note [InCasePattern]
+     In case pattern is strictly not necessary, we could do this evaluation in
+     the host's runtime instead of in the evaluator but doing so would mean that
```
gelisam, line 165: what does "the host's runtime" mean?

doyougnu: In cross-compilation the [host platform](https://wiki.osdev.org/Target_Triplet) is the system that the generated code will be run on. So all I'm saying here is that we just need to tag the continuation to know if the evaluator is in a primitive, in a case pattern or in a data case pattern. That tag does not strictly need to be in the evaluator, it could be part of the runtime, for example as a tag in some bits in the pointer to the data being evaluated.

PR #273, opened 2025-09-16
type-case says "no case matched" again

### LINKED ISSUE #272: type-case says "expected macro-action" instead of "no case matched"
```racket
#lang "prelude.kl"

(example
  -- This gets the expected error, "No case matched".
  (case (declaration)
    [(expression _)
     42]))

(define-macros
  ([m1
    (lambda (stx)
      (>>= (which-problem)
        (lambda (problem)
          (case problem
            [(expression tp)
             -- This gets an unexpected error, "Expected a macro action but got a type".
             -- The non-CEK evaluator gave the expected "No case matched".
             (type-case tp
               [(Integer)
                (pure '"hello")])]))))]))

(example (the String m1))
```

gelisam:
fixes #272

### DIFF
--- a/examples/non-examples/no-case-matched.golden
+++ b/examples/non-examples/no-case-matched.golden
@@ -1,0 +1,1 @@
+Error at phase p0: No case matched at no-case-matched.kl:4.3-6.10: (true)
--- a/examples/non-examples/no-case-matched.kl
+++ b/examples/non-examples/no-case-matched.kl
@@ -1,0 +1,6 @@
+#lang "prelude.kl"
+
+(example
+  (case (true)
+    [(false)
+     42]))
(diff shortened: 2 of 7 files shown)


doyougnu:
This looks good to me. My only complaint is that the commit messages could be more informative like `evaluator: repair type-case` and IMO each PR should be a single commit. None of that matters its only a habit for me. As for the code. My apologies for being messy :)

gelisam:
> the commit messages could be more informative like evaluator: repair type-case and IMO each PR should be a single commit. None of that matters its only a habit for me.

I see we have very different commit hygiene principles! I like to have many small commits which makes sense on their own (no WIP commits), and "evaluator: repair type-case" doesn't sound very informative to me :upside_down_face:

Issue #274, opened 2025-09-16
get rid of ShortShow


gelisam:
`Pretty` is the user-facing typeclass. `ShortShow` seems to only be used for scope names and assertion messages, for which `Show` probably suffices.

PR #275, opened 2025-09-29
Address #274

Aster89:
Just found this task on [Haskell Weekly 490](https://haskellweekly.news/issue/490.html), and thought it was a long hanging fruit to do some OS contribution for the Haskell community :)

`cabal build` passes. Is this enough?

I've also seen tons of warnings. Do I have to/can I address those too?


gelisam:
Thanks, I did not know Klister was mentioned in Haskell Weekly!

There's also a test suite, which CI will now run for you.

PR #276, opened 2025-10-20
compiler: add fixnums Word{64,32,16,8}, Int{64,32,16,8}

doyougnu:
and example/fixnums.kl

todos:

- [x]  bit level ops
- [x] .kl file that does bounds checking over the primitive types that this pr adds
- [x] .kl file that demonstrates numeric operations
- [x] No more todos
- [x] update documentation
- [x] rebase and squash


doyougnu:
I'd also like to do the bounds checking of the fixnums by wrapping them in a macro but I'm not sure exactly how to do that. So a `(the Word8 x)` would have a check that the `x` is not larger than `0xFF` and not lower than 0. Inserting these checks is for sure slower than building them into the compiler but I think that's fine for a prototype and a good dog fooding opportunity.

examples/fixnum.kl
```diff
@@ -0,0 +1,19 @@
...
+(example (the Word32 0o32))
+(example (the Word16 12))
+(example (the Word8 12))
+(example (the Int64 12))
+(example (the Int32 12))
+(example (the Int16 12))
+(example (the Int8  256))  -- note no bounds checking with the primitives
+(example (word8-compliment (the Word8 7))) -- will be -8 because 2's complement
+(example (int32-popcount (the Int32 255) )) -- 8
+(example (int8-+ (the Int8 42) (the Int8 1726)))
```
gelisam: since the types are

```
word8-compliment : Word8 → Word8
int32-popcount : Int32 → Int32
int8-+ : Int8 → Int8 → Int8
```

I would have expected the type-checker to be able to infer that `42` has type `Int8` etc. without the need for a type ascription. but when I remove them, I get a type error!
```
Expected (Integer → (?1 → ?2))
but got (Int8 → (Int8 → Int8))
Specifically, Integer doesn't match Int8
```

does this mean that you didn't make `#%integer-literal` polymorphic, you made `the` convert from an `Integer` to an `Int8` etc.?

src/Expander/Primitives.hs
```diff
@@ -744,37 +845,160 @@ primitiveDatatype name args =
...
-  \(ValueInteger i1) ->
-    ValueClosure $ HO $
-    \(ValueInteger i2) ->
-      ValueInteger (f i1 i2)
+  ValueClosure $ HO $ \(ValueInteger i1) ->
+    ValueClosure $ HO $ \(ValueInteger i2) -> ValueInteger (f i1 i2)
+
+binaryInt8Prim :: HasCallStack => (Int8 -> Int8 -> Int8) -> Value
+binaryInt8Prim f =
+  -- TODO: @gelisam binary functions fail because the pattern match in these
```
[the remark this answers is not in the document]
gelisam: >  I'm not sure where the consumer is for these closures.

Sure you do, you wrote the evaluator! It's [apply](https://github.com/gelisam/klister/blob/060ee3176e6bd2b5008c8f3d55f2932f8ff62d14/src/Evaluator.hs#L588).

src/Parser.hs
```diff
@@ -148,7 +165,15 @@ identName =
...
              <|> (literal "#%integer-literal" $> "#%integer-literal")
-             <|> (literal "#%string-literal" $> "#%string-literal")
+             <|> (literal "#%word64-literal"  $> "#%word64-literal")
+             <|> (literal "#%word32-literal"  $> "#%word32-literal")
+             <|> (literal "#%word16-literal"  $> "#%word16-literal")
+             <|> (literal "#%word8-literal"   $> "#%word8-literal")
+             <|> (literal "#%int64-literal"   $> "#%int64-literal")
+             <|> (literal "#%int32-literal"   $> "#%int32-literal")
+             <|> (literal "#%int16-literal"   $> "#%int16-literal")
+             <|> (literal "#%int8-literal"    $> "#%int8-literal")
```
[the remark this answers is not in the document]
gelisam: [expandOneForm](https://github.com/gelisam/klister/blob/5197ae8691a99e147f6ae10d7c72ac522a00971a/src/Expander.hs#L1683) is the part of the expander which matches on `(my-macro x y z)` and calls the evaluator on `my-macro`'s implementation. When it encounters syntax which is _not_ a macro call, like `42` or `"hello"`, it makes it a macro call by turning it into `(#%integer-literal 42)` or `(#%string-literal "hello")`, then it calls the evaluator on that.

I see that you have added the `Int64` data constructor etc. to `ExprF`. I don't think this is right, this is the datatype which describes an s-expression, so there should be one data constructor per syntactic form, not one data constructor per type. We could add a data constructor for hex literals etc., that would make it possible to write a macro which treats `(my-macro 0xff)` differently from `(my-macro 255)`. Personally, I think that would just add a lot of overhead to macro writers, I think it would be easier for macro writers if we kept a single `Integer` constructor, and the parser took care of converting `0xff` to `Integer 255`. In which case you would not need `#%hex-literal`, because the expander would not be able to tell the difference between `0xff` and `255` either, they would both expand to `(#%integer-literal 255)`.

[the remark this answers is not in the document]
gelisam: I don't think we actually have any primitive macros which look at the expected type yet, only user-defined macros currently do this, by calling `(which-problem)`. So I think the easiest would be to introduce `integer->int64` primitives and to have the `#lang "prelude.kl"` version of `#%integer-literal` call `(which-problem)` in order to determine whether it should call `int64-literal` or not:

```
(define-macro (my-integer-literal n)
  (>>= (which-problem)
    (lambda (problem)
      (case problem
        [(expression t)
         (type-case t
           [Integer
            (pure n)]
           [Int64
            (pure `(integer->int64 ,n))])]))))
```

doyougnu:
> > +1,332 −101
> 
> oh no that one is even bigger than the other MR 😩
> 
> I'll never manage to get through both of those in their entirety. should we try splitting them into smaller MRs, or just merge them as-is and hope for the best?

I'm sorry!! Yes I had a 8 hour flight and then a 13 hour flight to Singapore for ICFP so I just decided to add them all. I'll keep it small next time

src/Parser.hs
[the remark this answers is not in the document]
gelisam: > I got this working and pushed but now many tests are throwing `No progress possible`!

Makes sense, there are probably many expressions, e.g. `(example 42)`, whose type was previously unambiguously `Integer`, and now `42` has an ambiguous type: it could be Integer, or Word8, etc. Haskell would use type-defaulting to default to Integer, but Klister does not have type defaulting. And Klister does not yet have typeclasses, so we cannot infer the type `42 :: forall a. Num a => a` either.

Furthermore, since `(example 42)` secretly expands to `(example (#%integer-literal 42))` and the implementation of `#%integer-literal` looks at the expected type in order to decide whether to produce `(integer->word8 42)` or something else, in those ambiguous cases where nothing is constraining the 42 to be an Integer, the `#%integer-literal` macro gets stuck on the `type-case` line, waiting for the type-checker to provide a type which will never come.

Another interesting test case once the debugger supports execution and not just evaluation!

gelisam:
`examples/pmatch.kl` seems pretty redundant; it implements nested pattern-matching as a macro, but the `case` primitive now supports pattern-matching out of the box. in any case, the fix for the ambiguous part seems pretty straightforward:
```
(example (swap-list
           (:: (pair (the Integer 1) '1)
             (:: (pair (the Integer 2) '2)
               (:: (pair (the Integer 3) '3)
                 (nil))))))
```

Then the only remaining difference in the output is the position within the file where the syntax objects `'1`, `'2`, and `'3` are found. Which makes sense, because the extra `(the Integer` we've just added in front has pushed them a few characters to the right. I usually delete the golden file, re-run `cabal test`, look at the golden file's diff, and I `git add` the differences if the only difference is the position.

```
All tests
  Golden tests
    pmatch: FAIL (2.91s)
      Test output was different from 'examples/pmatch.golden'. Output of ["diff","-u","examples/pmatch.golden","/tmp/pmatch1185244-0.actual"]:
      --- examples/pmatch.golden	2025-11-24 01:17:05.078682422 -0500
      +++ /tmp/pmatch1185244-0.actual	2025-11-24 01:17:08.409654303 -0500
      @@ -7,6 +7,6 @@
       (zero) : Nat
       (zero) : Nat
       (add1 (zero)) : Nat
      -(:: (pair #[pmatch.kl:119.34-119.35]<1> 1)
      -    (:: (pair #[pmatch.kl:119.50-119.51]<2> 2)
      -        (:: (pair #[pmatch.kl:119.66-119.67]<3> 3) (nil)))) : (List (Pair Syntax Integer))
      +(:: (pair #[pmatch.kl:120.39-120.40]<1> 1)
      +    (:: (pair #[pmatch.kl:121.41-121.42]<2> 2)
      +        (:: (pair #[pmatch.kl:122.43-122.44]<3> 3) (nil)))) : (List (Pair Syntax Integer))
```

gelisam:
`example/implicit-conversion-test.kl` exercises the `#lang` defined in `example/implicit-conversion.kl`. It's a language which automatically adds conversion functions where needed, e.g. in
```
(let-implicit string-length
  (+ 1 "foo"))
```
the `#lang`'s implementation of function application notices that `+` expects 2 Integers but `"foo"` is a String, so it inserts the appropriate conversion function:
```
(+ 1 (string-length "foo"))
```

There are two difficulties:
1. This `#lang` does not provide `the` (or rather it uses `the` for something else), so we can't just write `(the Integer 42)` instead of `42`.
2. The `#lang`'s implementation of function application must be able to notice that `1` is already a string and thus that no conversion function need to be added around it.

For those two reasons, I think it would make more sense if integer literals were monomorphic in that `#lang`. To do this, we need to modify `examples/implicit-conversion.kl`, not `examples/implicit-conversion-test.kl`. That module defines `my-integer-literal`, which gets renamed to `#%integer-literal` during the `export`, and is thus how integer literals like `42` will get interpreted in this language. Currently it generates a call to `#lang "prelude.kl"`'s `#%integer-literal`, which is polymorphic on your branch. So it now needs to add a type ascription around that `#%integer-literal` call in order to make sure this `42` is inferred to have type Integer:
```
(define-macro (my-integer-literal n)
  (pure `(reader-pure (the Integer (#%integer-literal ,n)))))
```

examples/non-examples/type-errors/not-a-function.golden
```diff
@@ -1,2 +1,2 @@
-Type mismatch at not-a-function.kl:3.11-3.13.
-Expected (Integer → (?1 → ?2)) but got Integer
+Type mismatch at not-a-function.kl:3.24-3.26.
+Expected (?1 → (?2 → ?3)) but got Integer
```
gelisam, line 2: since `not-a-function.kl` specifically annotates that the `42` has type `Integer`, I don't understand why the error message changed?

this is not a blocker.

examples/fixnum.kl
```diff
@@ -0,0 +1,17 @@
...
+(example (the Word32 0o32))    -- 26
+(example (the Word16 22))
+(example (the Word8 12))
+(example (the Int64 12))
+(example (the Int32 12))
+(example (the Int16 12))
+(example (the Int8  256))  -- note no bounds checking with the primitives this is 0
+(example (int32-popcount 0xFF)) -- 8
+(example (int8-+ 42 50))
+(example (int8-compliment 7)) -- will be -8 because 2's complement
```
gelisam, line 15:
```suggestion
(example (int8-complement 7)) -- will be -8 because 2's complement
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "int8/=" #<closure>) : (Pair String (Int8 → (Int8 → Bool)))
+(pair "integer->int8" #<closure>) : (Pair String (Integer → Int8))
+(pair "int8-and" #<closure>) : (Pair String (Int8 → (Int8 → Int8)))
+(pair "int8-or" #<closure>) : (Pair String (Int8 → (Int8 → Int8)))
+(pair "int8-xor" #<closure>) : (Pair String (Int8 → (Int8 → Int8)))
+(pair "int8-rotate" #<closure>) : (Pair String (Int8 → (Int8 → Int8)))
+(pair "int8-shift" #<closure>) : (Pair String (Int8 → (Int8 → Int8)))
+(pair "int8->string" #<closure>) : (Pair String (Int8 → String))
+(pair "int8-popcount" #<closure>) : (Pair String (Int8 → Int8))
+(pair "int8-compliment" #<closure>) : (Pair String (Int8 → Int8))
```
gelisam:
```suggestion
(pair "int8-complement" #<closure>) : (Pair String (Int8 → Int8))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "word64/=" #<closure>) : (Pair String (Word64 → (Word64 → Bool)))
+(pair "integer->word64" #<closure>) : (Pair String (Integer → Word64))
+(pair "word64-and" #<closure>) : (Pair String (Word64 → (Word64 → Word64)))
+(pair "word64-or" #<closure>) : (Pair String (Word64 → (Word64 → Word64)))
+(pair "word64-xor" #<closure>) : (Pair String (Word64 → (Word64 → Word64)))
+(pair "word64-rotate" #<closure>) : (Pair String (Word64 → (Word64 → Word64)))
+(pair "word64-shift" #<closure>) : (Pair String (Word64 → (Word64 → Word64)))
+(pair "word64->string" #<closure>) : (Pair String (Word64 → String))
+(pair "word64-popcount" #<closure>) : (Pair String (Word64 → Word64))
+(pair "word64-compliment" #<closure>) : (Pair String (Word64 → Word64))
```
gelisam:
```suggestion
(pair "word64-complement" #<closure>) : (Pair String (Word64 → Word64))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "word16/=" #<closure>) : (Pair String (Word16 → (Word16 → Bool)))
+(pair "integer->word16" #<closure>) : (Pair String (Integer → Word16))
+(pair "word16-and" #<closure>) : (Pair String (Word16 → (Word16 → Word16)))
+(pair "word16-or" #<closure>) : (Pair String (Word16 → (Word16 → Word16)))
+(pair "word16-xor" #<closure>) : (Pair String (Word16 → (Word16 → Word16)))
+(pair "word16-rotate" #<closure>) : (Pair String (Word16 → (Word16 → Word16)))
+(pair "word16-shift" #<closure>) : (Pair String (Word16 → (Word16 → Word16)))
+(pair "word16->string" #<closure>) : (Pair String (Word16 → String))
+(pair "word16-popcount" #<closure>) : (Pair String (Word16 → Word16))
+(pair "word16-compliment" #<closure>) : (Pair String (Word16 → Word16))
```
gelisam:
```suggestion
(pair "word16-complement" #<closure>) : (Pair String (Word16 → Word16))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "word32/=" #<closure>) : (Pair String (Word32 → (Word32 → Bool)))
+(pair "integer->word32" #<closure>) : (Pair String (Integer → Word32))
+(pair "word32-and" #<closure>) : (Pair String (Word32 → (Word32 → Word32)))
+(pair "word32-or" #<closure>) : (Pair String (Word32 → (Word32 → Word32)))
+(pair "word32-xor" #<closure>) : (Pair String (Word32 → (Word32 → Word32)))
+(pair "word32-rotate" #<closure>) : (Pair String (Word32 → (Word32 → Word32)))
+(pair "word32-shift" #<closure>) : (Pair String (Word32 → (Word32 → Word32)))
+(pair "word32->string" #<closure>) : (Pair String (Word32 → String))
+(pair "word32-popcount" #<closure>) : (Pair String (Word32 → Word32))
+(pair "word32-compliment" #<closure>) : (Pair String (Word32 → Word32))
```
gelisam:
```suggestion
(pair "word32-complement" #<closure>) : (Pair String (Word32 → Word32))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "int16/=" #<closure>) : (Pair String (Int16 → (Int16 → Bool)))
+(pair "integer->int16" #<closure>) : (Pair String (Integer → Int16))
+(pair "int16-and" #<closure>) : (Pair String (Int16 → (Int16 → Int16)))
+(pair "int16-or" #<closure>) : (Pair String (Int16 → (Int16 → Int16)))
+(pair "int16-xor" #<closure>) : (Pair String (Int16 → (Int16 → Int16)))
+(pair "int16-rotate" #<closure>) : (Pair String (Int16 → (Int16 → Int16)))
+(pair "int16-shift" #<closure>) : (Pair String (Int16 → (Int16 → Int16)))
+(pair "int16->string" #<closure>) : (Pair String (Int16 → String))
+(pair "int16-popcount" #<closure>) : (Pair String (Int16 → Int16))
+(pair "int16-compliment" #<closure>) : (Pair String (Int16 → Int16))
```
gelisam:
```suggestion
(pair "int16-complement" #<closure>) : (Pair String (Int16 → Int16))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "int32/=" #<closure>) : (Pair String (Int32 → (Int32 → Bool)))
+(pair "integer->int32" #<closure>) : (Pair String (Integer → Int32))
+(pair "int32-and" #<closure>) : (Pair String (Int32 → (Int32 → Int32)))
+(pair "int32-or" #<closure>) : (Pair String (Int32 → (Int32 → Int32)))
+(pair "int32-xor" #<closure>) : (Pair String (Int32 → (Int32 → Int32)))
+(pair "int32-rotate" #<closure>) : (Pair String (Int32 → (Int32 → Int32)))
+(pair "int32-shift" #<closure>) : (Pair String (Int32 → (Int32 → Int32)))
+(pair "int32->string" #<closure>) : (Pair String (Int32 → String))
+(pair "int32-popcount" #<closure>) : (Pair String (Int32 → Int32))
+(pair "int32-compliment" #<closure>) : (Pair String (Int32 → Int32))
```
gelisam:
```suggestion
(pair "int32-complement" #<closure>) : (Pair String (Int32 → Int32))
```

examples/primitives-documentation.golden
```diff
@@ -1,5 +1,157 @@
...
+(pair "int64/=" #<closure>) : (Pair String (Int64 → (Int64 → Bool)))
+(pair "integer->int64" #<closure>) : (Pair String (Integer → Int64))
+(pair "int64-and" #<closure>) : (Pair String (Int64 → (Int64 → Int64)))
+(pair "int64-or" #<closure>) : (Pair String (Int64 → (Int64 → Int64)))
+(pair "int64-xor" #<closure>) : (Pair String (Int64 → (Int64 → Int64)))
+(pair "int64-rotate" #<closure>) : (Pair String (Int64 → (Int64 → Int64)))
+(pair "int64-shift" #<closure>) : (Pair String (Int64 → (Int64 → Int64)))
+(pair "int64->string" #<closure>) : (Pair String (Int64 → String))
+(pair "int64-popcount" #<closure>) : (Pair String (Int64 → Int64))
+(pair "int64-compliment" #<closure>) : (Pair String (Int64 → Int64))
```
gelisam:
```suggestion
(pair "int64-complement" #<closure>) : (Pair String (Int64 → Int64))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "word8/=" word8/=))
+(example (pair "integer->word8" integer->word8))
+(example (pair "word8-and"    word8-and))
+(example (pair "word8-or"     word8-or))
+(example (pair "word8-xor"    word8-xor))
+(example (pair "word8-rotate" word8-rotate))
+(example (pair "word8-shift"  word8-shift))
+(example (pair "word8->string" word8->string))
+(example (pair "word8-popcount" word8-popcount))
+(example (pair "word8-compliment" word8-compliment))
```
gelisam:
```suggestion
(example (pair "word8-complement" word8-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "word16/=" word16/=))
+(example (pair "integer->word16" integer->word16))
+(example (pair "word16-and"    word16-and))
+(example (pair "word16-or"     word16-or))
+(example (pair "word16-xor"    word16-xor))
+(example (pair "word16-rotate" word16-rotate))
+(example (pair "word16-shift"  word16-shift))
+(example (pair "word16->string" word16->string))
+(example (pair "word16-popcount" word16-popcount))
+(example (pair "word16-compliment" word16-compliment))
```
gelisam:
```suggestion
(example (pair "word16-complement" word16-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "word32/=" word32/=))
+(example (pair "integer->word32" integer->word32))
+(example (pair "word32-and"    word32-and))
+(example (pair "word32-or"     word32-or))
+(example (pair "word32-xor"    word32-xor))
+(example (pair "word32-rotate" word32-rotate))
+(example (pair "word32-shift"  word32-shift))
+(example (pair "word32->string" word32->string))
+(example (pair "word32-popcount" word32-popcount))
+(example (pair "word32-compliment" word32-compliment))
```
gelisam:
```suggestion
(example (pair "word32-complement" word32-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "word64/=" word64/=))
+(example (pair "integer->word64" integer->word64))
+(example (pair "word64-and"    word64-and))
+(example (pair "word64-or"     word64-or))
+(example (pair "word64-xor"    word64-xor))
+(example (pair "word64-rotate" word64-rotate))
+(example (pair "word64-shift"  word64-shift))
+(example (pair "word64->string" word64->string))
+(example (pair "word64-popcount" word64-popcount))
+(example (pair "word64-compliment" word64-compliment))
```
gelisam:
```suggestion
(example (pair "word64-complement" word64-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "int8/=" int8/=))
+(example (pair "integer->int8" integer->int8))
+(example (pair "int8-and"    int8-and))
+(example (pair "int8-or"     int8-or))
+(example (pair "int8-xor"    int8-xor))
+(example (pair "int8-rotate" int8-rotate))
+(example (pair "int8-shift"  int8-shift))
+(example (pair "int8->string" int8->string))
+(example (pair "int8-popcount" int8-popcount))
+(example (pair "int8-compliment" int8-compliment))
```
gelisam:
```suggestion
(example (pair "int8-complement" int8-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "int16/=" int16/=))
+(example (pair "integer->int16" integer->int16))
+(example (pair "int16-and"    int16-and))
+(example (pair "int16-or"     int16-or))
+(example (pair "int16-xor"    int16-xor))
+(example (pair "int16-rotate" int16-rotate))
+(example (pair "int16-shift"  int16-shift))
+(example (pair "int16->string" int16->string))
+(example (pair "int16-popcount" int16-popcount))
+(example (pair "int16-compliment" int16-compliment))
```
gelisam:
```suggestion
(example (pair "int16-complement" int16-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "int32/=" int32/=))
+(example (pair "integer->int32" integer->int32))
+(example (pair "int32-and"    int32-and))
+(example (pair "int32-or"     int32-or))
+(example (pair "int32-xor"    int32-xor))
+(example (pair "int32-rotate" int32-rotate))
+(example (pair "int32-shift"  int32-shift))
+(example (pair "int32->string" int32->string))
+(example (pair "int32-popcount" int32-popcount))
+(example (pair "int32-compliment" int32-compliment))
```
gelisam:
```suggestion
(example (pair "int32-complement" int32-complement))
```

examples/primitives-documentation.kl
```diff
@@ -66,6 +66,175 @@
...
+(example (pair "int64/=" int64/=))
+(example (pair "integer->int64" integer->int64))
+(example (pair "int64-and"    int64-and))
+(example (pair "int64-or"     int64-or))
+(example (pair "int64-xor"    int64-xor))
+(example (pair "int64-rotate" int64-rotate))
+(example (pair "int64-shift"  int64-shift))
+(example (pair "int64->string" int64->string))
+(example (pair "int64-popcount" int64-popcount))
+(example (pair "int64-compliment" int64-compliment))
```
gelisam:
```suggestion
(example (pair "int64-complement" int64-complement))
```

src/Syntax/Syntax.hs
```diff
@@ -34,10 +36,19 @@ data Stx a = Stx
...
   | String Text
   | Integer Integer
+  | Int64 Int64
+  | Int32 Int32
+  | Int16 Int16
+  | Int8  Int8
+  | Word64 Word64
+  | Word32 Word32
+  | Word16 Word16
+  | Word8  Word8
```
gelisam, line 51: hmm, I remember we decided that 42 would be a polymorphic syntax which can either have time Int8, Integer, or whatever, depending on the context? If so, it doesn't make sense for `ExprF` to have a constructor for Int8 etc., as there is no dedicated `42u8` syntax for it. Similarly, there should not be a Syntax-Contents constructor for Int8 either.

❌ this is a blocker

gelisam: (nor a `CoreInt8Syntax` constructor)

PR #277, opened 2025-11-11
explain scope sets

gelisam:


commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
+A "scope" intuitively means a region of code where some variables are visible. In a language without hygienic macros, scopes are neatly nested within each other, and when multiple bindings bind the same name, the innermost binding wins:
+```
+-- example 1
+(let [x 1]    -- call this binding "x1"
+  -- scope 1 begins {
+  (let [x 2]  -- call this binding "x2"
+    -- scope 2 begins {
+    x         -- x1 and x2 are both visible but x2 wins
```
doyougnu:
```suggestion
    x         -- at this reference, both x1 and x2 are visible, but x2 wins.
```

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
+A "scope" intuitively means a region of code where some variables are visible. In a language without hygienic macros, scopes are neatly nested within each other, and when multiple bindings bind the same name, the innermost binding wins:
+```
+-- example 1
```
doyougnu:
```suggestion
-- example 1: This example demonstrates scope sets in a language that does not have hygenic macros
```

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+In a world with macros, the situation is more complicated. 
+```
+-- example 2
+(let [x 1]         -- call this binding "x1"
+  (my-macro [x 2]  -- call this binding "x2"
+    -- which bindings are visible here depends
+    -- on which code my-macro produces
+    x))
```
doyougnu:
```suggestion
    x))   -- now which 'x' should this reference refer to?
```

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
...
+
+So the reason x1 and x2 are both visible is because the use site is tagged with both scopes, which fulfills both sets of requirements. The reason x2 wins is because its scope set is strictly bigger than x1's, so it shadows x1. It is possible to construct weird situations in which no variable shadows all the other variables, in which case the code is ambiguous and Klister rejects the program.
+
+Let's now look at example 2. The parts which are lexically within binding x1 _before_ `my-macro` is expanded will be tagged with scope 1:
+```
+-- example 2, scope 1
+(let [x 1]  -- scope 1
+  (my-macro
+    [x 2]   -- scope 1
+    x))     -- scope 1
```
doyougnu:
```suggestion
    [x 2]   -- scope 1: _all_ locations in the AST are tagged, even arguments passed to macros.
    x))     -- scope 1
```

commentary/scope-sets.md
gelisam:
```suggestion
-- example 1: This example demonstrates scopes in a language that does not have macros
```
your version made it sound like the language has unhygienic macros, but really, this is just an example which does not use macros

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
...
+```
+-- example 2, scope M, expanded
+(let _
+  (let [x 3]  -- scope M
+    (let _
+      (+ x    -- scope M
+         _))))
+```
+
+The last binding, `(let [x 2] _)`, only appears after `my-macro` is expanded, so this time it is the parts which are lexically within binding x2 _after_ `my-macro` is expanded which are tagged with scope 2:
```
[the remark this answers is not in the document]
gelisam: > I had forgotten what `x2` referred to by this point. Maybe just point out the binding?

[like this](f550e8849db6cc1d6aee5f58d04888c9771cd7c0)?

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
...
+(let [x 1]      -- call this binding "x1"
+  (let [x 3]    -- call this binding "xM"
+    (let [x 2]  -- call this binding "x2"
+      (+ x      -- xM should win here because this x comes from the body of my-macro
+         x))))  -- but x2 should win here
+```
+
+Obviously we cannot use the "innermost binding wins" algorithm here, or x2 would win at both use sites. We need a more complicated algorithm, the "scope sets" algorithm. 
+
+In this algorithm, each location in the AST is tagged with a set of `Scope` values. Two `Scope` values can be compared for equality, and that's it, there are no other operations on `Scope` values. Here is how the scope set algorithm works in example 1:
```
doyougnu:
```suggestion
In this algorithm, each location in the AST is tagged with a set of `Scope` values, called a `ScopeSet`. Two `Scope` values can be compared for equality, and that's it, there are no other operations on `Scope` values. `ScopeSet`s are partially ordered, so we can compare two `ScopeSet`s to observe which one is larger. Here is how the scope set algorithm works in example 1:
```

commentary/scope-sets.md
```diff
@@ -0,0 +1,116 @@
...
+    (let [x 2]  -- scope 1
+      (+ _
+         x))))  -- scope 1
+```
+
+The parts which are lexically within the `my-macro` definition will be tagged with scope M:
+```
+(define-macro (my-macro binding expr)
+  (pure `(let [x 3]      -- scope M
+           (let ,binding
```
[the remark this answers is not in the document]
doyougnu:
```suggestion
           (let ,binding -- splices are already tagged. In this example binding is 
                         -- bound to `'[x 2]` which was tagged with scope 1
```

commentary/scope-sets.md
gelisam: I have an idea: 1b76422a6553aa22f2ab340a80f9cd9cc3ecf2c6

PR #279, opened 2025-12-22
interleaving commentary

gelisam:
Explaining how macro-expansion and type-checking are interleaved in the Klister implementation.


doyougnu (approved):
just minor comments. I think it is very good!

